PR #36, opened 2018-02-01
Jwilaby/bot context as revocable

justinwilaby:
This PR provides a revocable, extensible BotContext solution without the need to add properties to the `prototype` or explicitly throw when the context is disposed.

## example:
```js
// Create the plugin like any other class - it must implement BotContext
class ContextPlugin implements BotContext {
  public doStuff(): void {
    if (this.request.entities) {
     this.processEntities(this.request.entities);
    }
  }
  private processEntities(): void{
    // Process entities
  }
}
// Create a middleware that knows about the ContextPlugin type
class ContextSpecificMiddleware implements Middleware { 
  receiveActivity(context: ContextPlugin , next: () => Promise<void>): Promise<void>{
    context.doStuff(); // TypeScript should know about this without needing ambient declarations
  }
}

const bot = new Bot();
// Add as many plugins as needed to carry out the mission
bot.plugin(new ContextPlugin()
      .use(new ContextSpecificMiddleware());
```

### DIFF
--- a/libraries/botbuilder/src/bot.ts
+++ b/libraries/botbuilder/src/bot.ts
@@ -26,7 +26,7 @@ import { BotContext } from './botContext';
 export class Bot extends MiddlewareSet {
     private receivers: ((context: BotContext) => Promiseable<void>)[] = [];
     private _adapter: ActivityAdapter;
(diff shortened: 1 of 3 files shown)


Stevenic:
Closing...

PR #50, opened 2018-02-08
Jwilaby/web samples

justinwilaby:
The PR provides 2 new samples for the browser:
1. AlarmBot as a custom built chat
2. AlarmBot using the [BotFramework-WebChat](https://github.com/Microsoft/BotFramework-WebChat)


samples/alarmbot-es6-botframework-webchat/src/alarmRenderer.js
```diff
@@ -0,0 +1,11 @@
+export class AlarmRenderer {
```
Stevenic: I thought the norm in a vanilla JS project was to just put all of the code in "./lib"? The idea being that since you don't require a compile step your JS is already in output form.

samples/alarmbot-es6-botframework-webchat/src/routes.js
```diff
@@ -0,0 +1,24 @@
+// configure bots routing table
+import {alarms} from "./alarms";
+import {Prompt} from 'botbuilder-prompts';
+
+export function routes(context) {
+    if (context.ifRegExp(/(list|show) alarms/i)) {
```
Stevenic: >context.ifRegExp [](start = 8, length = 16)

This was just removed... And it looks like you're working against an older copy of the alarm bot sample.

samples/alarmbot-es6-botframework-webchat/src/webChatAdapter.js
```diff
@@ -0,0 +1,25 @@
...
+    post(activities) {
+        activitiesById[activities[0].replyToId] = activities;
+        activities.forEach(activity => Object.assign(activity, {id: activity.replyToId, from: {id: "The Bot"}}));
+        return Promise.resolve(activities);
+    }
+
+    onReceive(activity) {
+        // overwritten by bot
+    }
+
```
Stevenic: I'm assuming the reason you chose to use a function here versus a property is because in ES6 you can't show that its typed to take a function.  I wonder if it would be cleaner use a setter here. I just filled an issue with TypeScript that there's currently no way to express in an interface that something is writeonly.  This is a classic "weak reference" pattern.

samples/alarmbot-es6-botframework-webchat/src/routes.js
justinwilaby: +1 Updated to use the latest alarm bot example

PR #68, opened 2018-02-11
Botframework redux es6

johnataylor:
- sample demonstrating how BotFramework can be usefully combined with other JavaScript frameworks


howlowck:
https://haoluo12.visualstudio.com/botbuilder-redux a middleware for full Redux integration with botbuilder v4
https://haoluo12.visualstudio.com/botbuilder-redux-common has a sample that I got Redux Dev Tools working with the state

PR #76, opened 2018-02-14
New Recognizer Result Schema + LUIS Recognizer imlpementation

emadelwany:


libraries/botbuilder/src/recognizer.ts
```diff
@@ -0,0 +1,191 @@
...
+import { Intent } from './intentRecognizer';
+
+
+/**
+ * Middleware that's the base class for all intent recognizers.
+ * 
+ * __Extends BotContext:__
+ * * context.topIntent - The top recognized `Intent` for the users utterance.
+ */
+export class Recognizer implements Middleware {
```
Stevenic: >Recognizer [](start = 13, length = 10)

Why did you have to add a new Recognizer class? Is it because the result type changed?

emadelwany: @Stevenic yes, eventually all recognizers should implement this instead of IntentRecognizer and we should delete the old IntentRecognizer, I just didn't want to change the old one till we agree on the new schema and have it working for one recognizer.

libraries/botbuilder-ai/lib/luisRecognizer.d.ts
```diff
@@ -5,8 +5,7 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-import { Intent, IntentRecognizer } from 'botbuilder';
-import LuisClient = require('botframework-luis');
+import { Recognizer, RecognizerResult } from 'botbuilder';
```
chrimc62: contextId has been marked obsolete for a while since it was connected to action binding.  Do we want to support it moving forward in the V4 SDK?

libraries/botbuilder-ai/lib/luisRecognizer.js
```diff
@@ -19,36 +18,170 @@ class LuisRecognizer extends botbuilder_1.IntentRecognizer {
...
+        }
+        else {
+            return entity.entity;
+        }
+    }
+    getEntityMetadata(entity) {
+        return {
+            $startIndex: entity.startIndex,
+            $endIndex: entity.endIndex,
+            $entity: entity.entity,
```
chrimc62: Maybe we should change this to something like $text to reflect this is the original text that led to the entity?

libraries/botbuilder-ai/lib/luisRecognizer.js
```diff
@@ -19,36 +18,170 @@ class LuisRecognizer extends botbuilder_1.IntentRecognizer {
...
+            $score: entity.score,
+            $resolution: entity.resolution ?
```
chrimc62: In the general case, I don't think we need this field.  It should be in the main JSON rather than duplicated here.  The one case I can think of where you might want to put it here would be the future/past date times, but I'm not sure that is worth preserving.

libraries/botbuilder-ai/lib/luisRecognizer.js
```diff
@@ -19,36 +18,170 @@ class LuisRecognizer extends botbuilder_1.IntentRecognizer {
         else {
             this.options = Object.assign({}, appId);
         }
+        // We need to set verbose to true in order to get all intents
+        if (this.options.options) {
+            this.options.options.verbose = true;
```
chrimc62: Where do the explicit options override this?

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -43,40 +43,198 @@ export class LuisRecognizer extends IntentRecognizer {
             this.options = { appId: appId, subscriptionKey: subscriptionKey as string };
         } else {
             this.options = Object.assign({}, appId);
+        }   
+
+        // We need to set verbose to true in order to get all intents
+        if(this.options.options) {
+            this.options.options.verbose = true;
```
chrimc62: Is all the right default?  Where do you use the explicit options?

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -43,40 +43,198 @@ export class LuisRecognizer extends IntentRecognizer {
...
+            .then((result : LuisResult) => {
+                let entitiesAndMetadata = $this.getEntitiesAndMetadata(result.entities, result.compositeEntities, verbose);
+                let recognizerResult  : RecognizerResult = {
+                    text: result.query,
+                    intents: $this.getIntents(result.intents),
+                    entities: entitiesAndMetadata.entities
+                };
+                if(verbose) {
+                    recognizerResult.$instance = {
+                        entities : entitiesAndMetadata.metadata
```
chrimc62: For intents is the right Node thing to do, as not have it there or to have intents:[{}, ...] for each intent in intents?  If we did that then you can always do the path for intent meta-data--even if there is none.  If it is reasonable for people to check for the presence of intents, it would be OK to not have it.

libraries/botbuilder-ai/tests/luisRecognizerTests.js
```diff
@@ -19,43 +20,191 @@ describe('LuisRecognizer', function () {
...
+                assert(res[0].intents.Delivery);
+                assert(res[0].intents.Delivery > 0 && res[0].intents.Delivery <= 1);
+                assert(res[0].entities);
+                assert(res[0].entities['builtin.number']);
+                assert(res[0].entities['builtin.number'][0] === 2001);
+                assert(res[0].entities['builtin.datetimeV2.date']);
+                assert(res[0].entities['builtin.datetimeV2.date'][0] === '2001-02-02');
+                assert(res[0].$instance);
+                assert(res[0].$instance.entities);
+                assert(res[0].$instance.entities['builtin.number']);
```
chrimc62: Were you going to map from . to _ or :?

libraries/botbuilder-ai/lib/luisRecognizer.d.ts
emadelwany: Sure - removing.

libraries/botbuilder-ai/lib/luisRecognizer.js
emadelwany: Sounds reasonable.

libraries/botbuilder-ai/lib/luisRecognizer.js
emadelwany: They can't, I'm forcing verbose=true so that we'd always get all intents, and then the user can use recognizer::getTopIntent() to filter to the top intent if needed. If you want I can give back that control to the developer and they get top/multiple intents depending on whether they set verbose, do you prefer that?

libraries/botbuilder-ai/src/luisRecognizer.ts
emadelwany: I'm kind of taking that control from the developer as I explained in my other comment, in favor of a consistent response, but happy to give the control back to the developer.

emadelwany: I'll give the control back to the developer for the time being.

libraries/botbuilder-ai/lib/luisRecognizer.js
chrimc62: But LUIS options include a verbose flag which means you are then ignoring it.  I think we should leave control to options and the question is whether the default should be all or only the top intent.  Depends on the expected use case--we should ask Steve/CSE guys.

emadelwany: Sounds reasonable, will do.

libraries/botbuilder-ai/src/luisRecognizer.ts
emadelwany: What would be the "value" for intent in the metadata? The Score again, or something else?

$instance.intents = { 'OrderSandwich': XYZ, 'OrderDrink': KLM }, what's XYZ/KLM? Also intents isn't an array in the main JSON, did you intent to make it an array in the metadata?

libraries/botbuilder-ai/tests/luisRecognizerTests.js
emadelwany: I still need to give this more thought. : is a no go since it's not valid JS. _ is reasonable but it looks ugly. I'll use _ for now but I'm not very satisfied with it, might change it later.

emadelwany:
@chrimc62 I addressed (and accepted) all your feedback, only pending issue is how to show intents in metadata, can you see the follow questions I posted about it? thanks!

libraries/botbuilder-ai/src/luisRecognizer.ts
emadelwany: I opted in for having an empty object placeholder.

emadelwany:
Closing this for till we agree on the final schema. Will create a new one later.
cc @chrimc62 @Stevenic @johnataylor @willportnoy @justinwilaby

PR #54, opened 2018-02-09
Auth Stack updates

pcostantini:
This PR contains the new auth stack ported from C#.

**The branch is based on the Connector/Models split, which contains the merge of Connector and Auth into the same package.**

_Changes_
- New JwtTokenValidation + JwtTokenExtractor
- Emulator and Channel validation are separated from each other
- BotFrameworkAdapter adepter to use new stack
- Jwt Validation tests

_Usage_
```TypeScript
let credentialsProvider = new SimpleCredentialProvider('AppID', 'AppPassword');

// using await/async support
await JwtTokenValidation.assertValidActivity(activity, authHeader, credentialsProvider);

// using promise
JwtTokenValidation.assertValidActivity(activity, authHeader, credentialsProvider).then(() => {
    // request is validated
    // continue
}).catch(err => {
   // request is not authorized
   // return 401
});
```

_Tests_
image

PR #81, opened 2018-02-15
Auth Stack updates

amitstein:
This PR  ([this ](https://github.com/Microsoft/botbuilder-js/pull/54) is the original) contains the new auth stack ported from C#.

**The branch is based on the Connector/Models split, which contains the merge of Connector and Auth into the same package.**

_Changes_
- New JwtTokenValidation + JwtTokenExtractor
- Emulator and Channel validation are separated from each other
- BotFrameworkAdapter adepter to use new stack
- Jwt Validation tests

_Usage_
```TypeScript
let credentialsProvider = new SimpleCredentialProvider('AppID', 'AppPassword');

// using await/async support
await JwtTokenValidation.assertValidActivity(activity, authHeader, credentialsProvider);

// using promise
JwtTokenValidation.assertValidActivity(activity, authHeader, credentialsProvider).then(() => {
    // request is validated
    // continue
}).catch(err => {
   // request is not authorized
   // return 401
});
```

_Tests_
image


libraries/botframework-connector/src/auth/microsoftAppCredentials.ts
```diff
@@ -1,83 +1,156 @@
...
                 if (!err) {
                     if (body && response.statusCode && response.statusCode < 300) {
                         // Subtract 5 minutes from expires_in so they'll we'll get a
                         // new token before it expires.
-                        var oauthResponse = JSON.parse(body);
-                        this.accessToken = oauthResponse.access_token;
-                        this.accessTokenExpires = new Date().getTime() + ((oauthResponse.expires_in - 300) * 1000);
-                        cb(null!, this.accessToken);
+                        var oauthResponse = <OAuthResponse>JSON.parse(body);
+                        oauthResponse.expiration_time = new Date(Date.now() + (oauthResponse.expires_in * 1000) - 60000);
```
Stevenic: >60000 [](start = 114, length = 5)

This is 1 minute not 5. I double checked the v3 code and we definitely subtract 5 minutes from this.

libraries/botframework-connector/src/auth/microsoftAppCredentials.ts
```diff
@@ -1,83 +1,156 @@
...
+                }
             }
-        });
+        }
+
+        // We need to refresh the token, because:
+        // 1. The user requested it via the forceRefresh parameter
+        // 2. We have it, but it's expired
+        // 3. We don't have it in the cache.
+        let oAuthToken = await this.refreshToken();
```
Stevenic: >let oAuthToken = await this.refreshToken(); [](start = 8, length = 43)

We recently made a change to the v3 code to ensure that we only attempt 1 refresh of the token at a time to prevent the login service from timing out.  The way this is currently coded if we were to get hammered with 100 simultaneous messages on startup we would attempt to retrieve 100 simultaneous tokens which will cause timeouts.  We need to wrap the refresh() call with a promise that everyone joins on to wait for the refresh to complete. You can review my v3 code to see how this works.

libraries/botframework-connector/src/auth/microsoftAppCredentials.ts
amitstein: fixed

libraries/botframework-connector/src/auth/microsoftAppCredentials.ts
amitstein: fixed

PR #83, opened 2018-02-16
fix WebChatAdapter

billba:


samples/alarmbot-es6-botframework-webchat/src/webChatAdapter.js
```diff
@@ -1,25 +1,46 @@
...
+                )
+            }
+        }
+    }
+
     post(activities) {
-        activitiesById[activities[0].replyToId] = activities;
-        activities.forEach(activity => Object.assign(activity, {id: activity.replyToId, from: {id: "The Bot"}}));
-        return Promise.resolve(activities);
+        let sentActivities = activities.map(activity => Object.assign({}, activity, {
```
Stevenic: >let [](start = 8, length = 3)

very minor nit but can be 'const'

samples/alarmbot-es6-botframework-webchat/src/webChatAdapter.js
```diff
@@ -1,25 +1,46 @@
...
+            id: Date.now().toString(),
+            channelId: 'WebChat',
+            conversation: { id: 'bot' },
+            from: { id: 'bot' }
+        }));
+
+        sentActivities.forEach(activity => this.activity$.next(activity));
```
Stevenic: I should have caught this in the previous code review but you're not handling the 'delay' activity we made up.  This isn't an official activity type YET but Dan is considering adding it given that it's natively supported in channels like Kik and Line. We currently model 'delay' in the adapter by just doing:  setTimeout(() => { next-activity }, activity.value)

Stevenic:
Nice consolidation... I should have noticed a few things in my original review. My bad

samples/alarmbot-es6-botframework-webchat/src/webChatAdapter.js
billba: My preference would be to flow this into Web Chat and letting it handle it natively.

Stevenic: Sounds good...

Issue #86, opened 2018-02-19
BaseUri for LuisRecognizer is missing /luis


mattmazzola:
https://github.com/Microsoft/botbuilder-js/blob/a10e27053e166800faa54f3cd673a8dfb0bfdc2d/libraries/botbuilder-ai/src/luisRecognizer.ts#L49

Expected:
The third parameter to LuisRecognizer is optional and should default to the correct endpoint.
This allows user to program such as:
```typescript
const luisRecognizer = new LuisRecognizer("<removed-app-id>", "<removed-subscription-key>")
...

bot.use(luisRecognizer)
```

Actual:

Because the base url is wrong, the requestUrl generated will be:
`https://westus.api.cognitive.microsoft.com/v2.0/apps/<appId>?subscription-key=<subscription-key>`
Which will return 404 and leave developer confused thinking their app id is wrong.

However, it should be:
`https://westus.api.cognitive.microsoft.com/luis/v2.0/apps/<appId>?subscription-key=<subscription-key>`

The work around is that the user must manually enter the base url as:
```
const luisRecognizer = new LuisRecognizer("<app-id>", "<subscription-key>", "https://westus.api.cognitive.microsoft.com/luis")
```

Stevenic:
I commented on your pull request #87

mattmazzola:
Not sure how I reproduced this after looking at PR comment so closing for now.

Stevenic:
Look for new bits to get published this week and then please re-open if the issue persist.

PR #87, opened 2018-02-19
Add /luis to default cognitive services url

mattmazzola:
The base url was pointing to cognitive services but didn't specify the service LUIS so the requests would fail with 404 Not Found

Closes: #86

### DIFF
--- a/libraries/botbuilder-ai/src/luisRecognizer.ts
+++ b/libraries/botbuilder-ai/src/luisRecognizer.ts
@@ -46,7 +46,7 @@ export class LuisRecognizer extends IntentRecognizer {
         }
 
         // Create client and override callbacks
-        const baseUri = (this.options.serviceEndpoint || 'https://westus.api.cognitive.microsoft.com');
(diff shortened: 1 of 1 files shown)


Stevenic:
Matt if you look at the next line of code after the one you changed it creates the client with `/luis/` appended to the url.

```JavaScript
const baseUri = (this.options.serviceEndpoint || 'https://westus.api.cognitive.microsoft.com');
this.luisClient = new LuisClient(baseUri + '/luis/');
```

This code was recently updated (there were other issues) and has not be deployed to NPM yet. Was the issue you were seeing based off the bit currently published to NPM? Because the code wouldn't line up which whats in the "master"

mattmazzola:
Hmm, yea.  Given that line I don't see how the base uri was ever generated without `/luis`; however, I was debugging through the middleware and took the value requestUrl directly from the LuisClient._getIntentsAndEntitiesV2 at runtime and `/luis` was missing at one point and all requests returned 404.

I re-ran it now without explicitly setting baseUri and it works fine...  I'm very confused.  I guess this is closed as non-issue then.

Issue #97, opened 2018-02-24
[bug] Bots using Console Adapter throw an unhandled promise exception for all incoming messages


microsoftly:
## Environment details
* node version: ```8.2.1```
* relevant tsconfig: 
``` javascript
{
    "target": "es6",
    "module": "commonjs",   
    "moduleResolution": "node"
}
```
relevant dependency versions: 
``` javascript
{
    "botbuilder": "^4.0.0-m1.0",
    "botbuilder-node": "^4.0.0-alpha2",
    "botbuilder-services": "^4.0.0-m1.0",
}
```
## Repro steps
``` typescript
import { Bot, MemoryStorage, BotStateManager } from 'botbuilder';
import { ConsoleAdapter } from 'botbuilder-node';

const adapter = new ConsoleAdapter().listen();

const bot = new Bot(adapter)
    .use(new MemoryStorage())
    .use(new BotStateManager())
    .onReceive((context) => {
        context.reply('HELLO!');
    });
```

send any message to the bot from the console

### Expected behavior
bot responds ```hey```
### Actual behavior
no bot response, but logged error of:
```
(node:63716) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: Cannot read property 'id' of undefined
(node:63716) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

## Research 
I've found the source of the error, which happens [on this line in the bot state manager](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder/src/botStateManager.ts#L139). There is no user defined for the ref.

The value of ref at the time of the error is:
```javascript
{
    channelId:"console",
    conversation:Object {id: "Convo1"},
    from:Object {id: "user", name: "User1"},
    id:"0",
    recipient:Object {id: "bot", name: "Bot"},
    text:"hey",
    timestamp:"2018-02-24T02:28:37.669Z",
    type:undefined
}
```

PR #98, opened 2018-02-24
Tests for Console Adapter bug that throws unhandled exceptions when used with a bot

microsoftly:
# Issue reference #97 

Because ```ConsoleAdapter``` is part of the botbuilder-node package, a new test file for testing the integration between core and other libraries (in this case botbuilder-node) seemed to be fitting. While this currently only has the test for the ```ConsoleAdapter``` I would expect it to be an appropriate file to add additional integration tests.

## Notes
 * this PR does not provide a solution, it merely gives a test to verify the bad state and the fix.
 * use of the console Adapter within a test scope is harmless, though the particular test will run slightly slower due to IO constraints.

### DIFF
--- a/libraries/botbuilder/tests/coreIntegrationTests.js
+++ b/libraries/botbuilder/tests/coreIntegrationTests.js
@@ -1,0 +1,42 @@
+const assert = require('assert');
+const builder = require('../');
+const { ConsoleAdapter } = require('../../botbuilder-node');
+
+describe('Core integration tests', () => {
+    describe('console adapter', () => {
+        let adapter;
+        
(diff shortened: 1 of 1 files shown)


Stevenic:
I'm sorry for the delay on this.  I somehow saw this in the PR list and thought it was something I had submitted but parked. My bad...

So we just did a massive restructuring of all the packages and this test would no longer pass. Could you look over the new unit tests (work in progress) and provide feedback on what you think is lacking.  I haven't gotten to doing any broad integration tests yet.

microsoftly:
No worries.

It looks like this may have solved itself since the last bit of work has been made

PR #102, opened 2018-02-27
New Luis Recognizer Schema

emadelwany:


libraries/botbuilder/lib/recognizerResult.d.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+/** second comment block */
+/**
+ * A result from a recognizer
+ */
+export interface RecognizerResult {
+    text: string;
+    intents: any;
+    entities: any;
+    $instance?: any;
```
chrimc62: Hmm.  We could use this for meta-data about the whole result, but it would not be the most likely place right now.

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -43,40 +45,188 @@ export class LuisRecognizer extends IntentRecognizer {
...
+            compositeEntityTypes = compositeEntities.map(compositeEntity => compositeEntity.parentType);
+            compositeEntities.forEach(compositeEntity => {
+                entities = $this.populateCompositeEntity(compositeEntity, entities, entitiesAndMetadata, verbose);
             });
+        }
+
+        entities.forEach(entity => {
+            // we'll address composite entities separately
+            if(compositeEntityTypes.indexOf(entity.type) > -1)
+                return;
```
chrimc62: Probably a philosophical issue and does not need to change, but I prefer:
1) Always using {} rather than relying on indenting for if clauses.  
2) Don't do return in the middle of functions--it is in effect a goto where it looks like you can add code later, but don't realize it won't be executed.
These both come from writing solid code

libraries/botbuilder/lib/recognizerResult.d.ts
emadelwany: Yep, it's a residual from the old approach, removing.

libraries/botbuilder-ai/src/luisRecognizer.ts
emadelwany: 1. Added the {}
2. I agree in general but in this instance the return just acts as a "continue" in order to reduce indentation in the function, I think reduced indentation in short blocks makes code more readable at a low expensive of an easy to follow goto.

PR #113, opened 2018-03-13
M2 Developer Workflow Tools Improvements

sayar:
Added npm postinstall scripts to automate lerna bootstrapping and make lerna transparent to the developer.
Added various npm run scripts to enable a single command to link botbuilder to the global node_modules folder to make it easier for developers to add their local cloned repo to a project.

The new developer workflow would look like this:
```bash
git clone git@github.com:Microsoft/botbuilder-js.git
cd botbuilder-js
git checkout m2
npm install 
npm run dev #ignore error messages
```

Any project that you would like to use your local clone botbuilder sdk, just go into that project and run `npm link botbuilder` to have it use the globally symlinked botbuilder module pointing to your local clone.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,21 @@
 {
+	"private": true,
+	"name": "botbuilder-packages",
+	"scripts": {
+		"postinstall": "lerna bootstrap --hoist",
+		"build": "lerna run build",
+		"clean": "lerna run clean",
+		"test": "lerna run test --stream",
+		"build-docs": "lerna run build-docs"
+	},
 	"dependencies": {
 		"replace-in-file": "^3.1.1",
 		"typedoc": "^0.9.0",
 		"typedoc-plugin-external-module-name": "^1.0.10",
 		"typedoc-plugin-markdown": "^1.0.12"
+	},
+	"devDependencies": {
+		"@types/jsonwebtoken": "7.2.5",
+		"lerna": "^2.9.0"
 	}
 }


package.json
```diff
@@ -1,8 +1,24 @@
 {
+	"private": true,
+	"name": "botbuilder-packages",
+	"scripts": {
+		"postinstall": "lerna bootstrap --hoist",
+		"build": "lerna run build",
+		"clean": "lerna run clean",
+		"test": "lerna run test --stream",
+		"build-docs": "lerna run build-docs",
+		"link-global": "npm link ./libraries/botbuilder",
```
Stevenic: Whats the need for this? Lerna should be managing all of the symlinks.

codefoster:
FWIW, this method alone did not work. I got a fail on the `npm run dev`. I would have chocked it up to an anomaly on my system, but the exact same thing happened to @ryanvolum.
Each of use tried @billba's method (from Team) and then @sayar's method here and then switched back to the `m2` branch to get it to work.
Using @sayar's method, the error we got on `npm run dev` had something to do with NOVERSIONS (don't remember the exact message now).

codefoster:
Found that error I was getting on @sayar's branch...
```bash
npm ERR! code ETARGET
npm ERR! notarget No matching version found for botframework-schema@4.0.0-m2.0
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.
npm ERR! notarget
npm ERR! notarget It was specified as a dependency of 'botbuilder-core'
npm ERR! notarget
```

package.json
sayar: It was to make it easier for people to link to a clone of the botbuilder-js project from outside the repo but I can remove it.

sayar:
@codefoster as we chatted about today in person, you can ignore all those npm errors. When you run `npm link`, npm likes to run a `npm install` which fails as it's a lerna project. I'm going to remove `npm run dev`. It's too specific for how I like to code.

sayar:
@stevenic this PR is ready to go now

justinwilaby:
I've since changed the test script to include nyc and coveralls hooks for all packages. This branch should probably be rebased from master before pushing to be sure these changes do not get stomped. 

Also, the [test scripts](https://github.com/Microsoft/botbuilder-js/commit/13506b4afa7267d82be0adb0f420ff94450c378f#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) for the web got bumped somehow and should probably be restored.

PR #116, opened 2018-03-14
Updated the tsconfig.json to prevent unneeded types from being pulled in

justinwilaby:
This fixes the issue with running tests but **does not** address the failing unit tests.


johnataylor:
looks good

PR #122, opened 2018-03-16
Stevenic/invoke

Stevenic:
With this change... A bot would reply to an incoming `invoke` activity by sending an outgoing `invokeResponse` activity containing the status code and body to return.  The `BotFrameworkAdapter` will write this activity to a map and unpack it once the promise for the overall request unwinds. Example:  

```JS
adapter.processRequest(req, res, (context) => {
     if (context.request.type === 'invoke') {
          return context.sendActivity({ type: 'invokeResponse', value: { status: 200, body: 'body' }});
     }
});
```


justinwilaby:
@Stevenic - looks like you'll need to rebase from master to get the build to pass. I implemented travis ci and coveralls yesterday.

Stevenic:
@justinwilaby I was planning to do that anyway

PR #143, opened 2018-03-24
Added simple prompt sample

ryanvolum:
Created simple sample that demonstrates how to prompt using just the state primitives (no abstraction).

Sample prompts for name and age - once it has these pieces of information it greets the user. This sample keeps it simple by leaving out any validation.


samples/primitive-prompts-ts/src/app.ts
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            // Prompt for age if we don't have it
+            if (!state.age) {
+                state.activePrompt = `agePrompt`;
+                return context.sendActivity(`How old are you?`);
+            }
+
+            // Now that we have age and name, greet the user!
+            return context.sendActivity(`Hello ${state.name}! You are ${state.age} years old.`);
```
Stevenic: >return [](start = 12, length = 6)

One minor comment... We've started using async/await in out TypeScript samples

PR #149, opened 2018-03-27
Unit Tests coverage increased

mostafaelaraby:
Updated unit tests to have a better code coverage in language translator and locale converter


Stevenic:
> CodeFlow for GitHub does not download executable files. If you trust this file and would like to run it, please download it yourself through GitHub.

What are all these additional .map files? 

---
Refers to: libraries/botbuilder-ai/lib/.languageTranslator.js_LOCAL_19236.map.swp:1 in 10bec01. [](commit_id = 10bec01f524cfc32fd65c061e91332688cfd6314, deletion_comment = False)

libraries/botbuilder-ai/src/languageTranslator.ts
```diff
@@ -92,7 +92,6 @@ export class LanguageTranslator implements Middleware {
                     message.text = text;
                     return Promise.resolve();
                 })
-                .catch(error => Promise.reject(error));
```
Stevenic: Oh yeah... Don't do that. If anything you'd want to re-throw the error.

Stevenic:
Not exactly sure why it thinks our coverage went down but the main issue I see with this PR is all of the added .map files.  I think you also may be running an older version of the TypeScript compiler.

PR #150, opened 2018-03-29
Added LUIS es6 sample

alyssaong1:
Largely based on the alarm bot sample, except using LUIS to identify add/delete/show. 

Used the Calendar prebuilt domain instead of alarm, because I can't seem to find the builtin alarm intents & entities in LUIS anymore. 

Left subId + key in for easy testing.


Stevenic:
Awesome :)

PR #156, opened 2018-04-01
Stevenic/m2 tweaks

Stevenic:
This PR contains the various tweaks we've agreed to. Major changes:

- Renamed `BotContext` to `TurnContext`.
- Renamed `Middleware.onProcessRequest` to `Middleware.onTurn`.
- Renamed `BotFrameworkAdapter.processRequest` to `BotFrameworkAdapter.processActivity`.
- Renamed `BotFrameworkAdapter.startConversation` to `BotFrameworkAdapter.createConversation`.
- Renamed `context.request` to `context.activity`.
- Exposed context cache as `context.services`.
- Added ability to pass an optional namespace to `ConversationState` and `UserState`.
- Removed `BatchOutput` class.


billba:
Love all of these except `processActivity` -- my notes showed that we hadn't settled on that one.

`processRequest` was a at least accurate - `processActivity` is weird because it neither starts nor ends with an activity. I supposed `processContext` is even more accurate, but what this method is really about is creating a Turn and it would be nice to encapsulate that somehow.

Stevenic:
@billba `processActivity()` is what it's called on the C# side of things and we said the goal was to align the names which is why we renamed `startConversation()` to `createConversation()` as well.  I'd propose we just try to finalize all of that. I personally like the idea of moving back to some form of `listen()` method and have it just take the callback in.

Stevenic:
I'm going to go ahead and merge in this change set because we have other PR's and I'm already going to have to resolve a few merge conflicts. I'll hold off on publishing the M3 bits yet.

PR #166, opened 2018-04-06
CosmosDB Storage Provider - Feedback

pcostantini:
This PR includes the feedback ported from the [C# PR](github.com/Microsoft/botbuilder-dotnet/pull/384).

The same provider now works with both DocumentDB and Graph APIs.
Thus, renamed to just `CosmosDbStorage`.

- Read: SELECT query includes only required attributes
- Move DB/Collection initialization check to the provider instance (instead of global cache)
- Input parameters validation


pcostantini:
The same provider now works with both DocumentDB and Graph APIs.
Thus, renamed to just `CosmosDbStorage`.

PR #178, opened 2018-04-11
AzureTableStorage improvements

pcostantini:
- Support for storing objects bigger than 32kb serialized
  When storing unicode text, this is 32768 characters.
  The AzureTables Storage provider stores the state object item as JSON.
  The workaround is to store the object in chunks of 32768 characters, using the PartitionKey as the Key and RowKey as the chunk position.
   - On Write: serialize to JSON and split into chunks, then create TableEntity with an incrementing RowKey. Delete any RowKey > than last generated RowKey (e.g.: cleanup chunks from a previous bigger object)
   - On Read: Query the table for the specific PartitionKey, retrieving all Rows and ordering by RowKey
   - On Delete: Query the table for the specific PartitionKey to delete, iterate over rows and delete them.
- Use TableBatch when updating & deleting items
- Validate parameters
- Docs update
- Tests: Add Emulator check


Stevenic:
Why wouldn't you just split the object across multiple columns? That's the typical approach to storing larger objects and lets you push the max size of an object to almost 1mb which seems more than adequate. If you want to store even larger objects simply add support for GZip compression as well.

pcostantini:
Hi @Stevenic ,
It makes much more sense to flatten the object and store each flattened field as columns.
I'll re-create both PRs with the new approach.
Regards,
-PabloC

PR #179, opened 2018-04-11
Added 2 ways translation  to translate messages sent to the user

mostafaelaraby:


libraries/botbuilder-ai/src/languageTranslator.ts
```diff
@@ -14,7 +14,8 @@ export interface TranslatorSettings {
     nativeLanguages: string[],
     noTranslatePatterns: Set<string>,
     getUserLanguage?: ((c: TurnContext) => string) | undefined,
```
Stevenic: >c: TurnContext) => string [](start = 24, length = 25)

These signatures should be more consistant.  use `context` instead of `c` for instance.

Stevenic:
>     translatorKey: string,

Minor nit: The delimiter here I think should technically be a semicolon. 

---
Refers to: libraries/botbuilder-ai/src/languageTranslator.ts:13 in cfbbe78. [](commit_id = cfbbe78680698b651b90ff1ffcbd716df30e23ea, deletion_comment = False)

libraries/botbuilder-ai/src/languageTranslator.ts
Stevenic: >undefined [](start = 53, length = 9)

You've marked this as optional so you don't need the 'undefined'

libraries/botbuilder-ai/src/languageTranslator.ts
```diff
@@ -46,37 +49,42 @@ export class LanguageTranslator implements Middleware {
...
         
         let targetLanguage = (this.nativeLanguages.indexOf(sourceLanguage) >= 0) ? sourceLanguage : this.nativeLanguages[0];
+        
+        await this.translateMessageAsync(context, context.activity, sourceLanguage, targetLanguage);
+
+        if (this.translateBackToUserLanguage) {
+            context.onSendActivities(async (newContext, activities, newNext) => {
+                let currentMessageActivity = activities[0];
+                await this.translateMessageAsync(newContext, currentMessageActivity, targetLanguage, sourceLanguage);
+                activities[0].text = currentMessageActivity.text;
```
Stevenic: >activities[0].text = currentMessageActivity.text; [](start = 16, length = 49)

This assignment happens translateMessageAsync(). You're not cloning the activity so why the additional assignment?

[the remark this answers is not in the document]
MahmoudKhaledAli: Modified in latest commit.

PR #186, opened 2018-04-16
Emit Luis Trace Info and fix Luis Bot

emadelwany:


libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -89,21 +108,25 @@ export class LuisRecognizer implements Middleware {
...
+                    context.services.set(this.cacheKey, {recognizerResult, luisResult});
+                    
+                    return this.emitTraceInfo(context, luisResult, recognizerResult).then(() => {
+                        return recognizerResult;
+                    });
                 });
     
         }
-        return Promise.resolve(cached);
+        return this.emitTraceInfo(context, cached.recognizerResult, cached.luisResult).then(() => {
```
Stevenic: >this.emitTraceInfo(context, cached.recognizerResult, cached.luisResult) [](start = 15, length = 71)

Do you really want to send multiple of these per/turn?  I would have thought just once when the initial call to recognize occurs...

Stevenic:
My only comment was the fact you're potentially sending the trace activity multiple times per/turn.

The other thing is I'm not sure I can take this change just yet as it would prevent me from being able to update the published bits. The reason being that there's not logic in the BotFrameworkAdapter to filter out the `trace` activity for non-emulator channels.  I can try to add that and submit a PR against your branch or you could add that to your PR...

Stevenic:
Sorry... meant to create a separate branch but ended up checking the `trace` blocking logic into your repo. Might look over my commit.

libraries/botbuilder-ai/src/luisRecognizer.ts
emadelwany: @Stevenic yes, we want all responses to include the trace activity so that the developer can inspect each one of them in the emulator. Otherwise we'll need to cache them in the emulator and figure out how to correlate the first call to future identical calls (i.e. your cache key logic).

emadelwany:
@Stevenic  - thanks for doing the emulator filtering. Looks like the change needs a test for the build check to succeed though?

libraries/botbuilder-ai/src/luisRecognizer.ts
Stevenic: But that's not what that's doing... That's going to send a trace message every time the developer looks up the cached results for the received request. That means I could send 1 message to the bot and get 3 or 4 trace activities back depending on how my bot is coded. You're assuming the developer is only going to call `recognize()` once per/turn and the results are being cached so they can call it as many times as they want.

emadelwany: Got it - thanks for explaining, I misunderstood the purpose of the cache and though it was to reduce LUIS service calls across turns, I re-read the code and now your original feedback makes more sense. 

I changed it so that it wouldn't emit the trace activity on cache-hits and I updated the unit tests to ensure that the activity is emitted only once on multiple calls to recognize() on the same context.

PR #209, opened 2018-04-24
Fixed timezone bug in LocaleConverter

MahmoudKhaledAli:


Stevenic:
You broke CodeFlow with too many commits :(

PR #214, opened 2018-04-27
Stevenic/dialog tweaks

Stevenic:


libraries/botbuilder-dialogs/lib/dialog.d.ts
```diff
@@ -7,50 +7,233 @@
...
+ * Dialog optimized for prompting a user with a series of questions. Waterfalls accept a stack of
+ * functions which will be executed in sequence. Each waterfall step can ask a question of the user
+ * and the users response will be passed as an argument to the next waterfall step.
+ *
+ * For simple text questions you can send the user a message and then process their answer in the
+ * next step:
+ *
+ * ```JS
+ *  dialogs.add('namePrompt', [
+ *      async function (dc) {
```
johnataylor: suggestion: consider making this an arrow function?
   async (dc) => { }

PR #215, opened 2018-04-30
Initial version of typescript support for LUISGen.

chrimc62:


libraries/botbuilder-ai/src/numberWithUnits.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+*/
+
+/** A number together with units returned by a recognizer. */
+export interface NumberWithUnits {
+    /** A recognized number if any. */
+    readonly number?: number;
```
Stevenic: >number [](start = 13, length = 6)

Should this be value or something? Number isn't a reserved word but it's a type in TypeScript

libraries/botbuilder-core-extensions/lib/recognizerResult.js.map
```diff
@@ -0,0 +1 @@
+{"version":3,"file":"recognizerResult.js","sourceRoot":"","sources":["../src/recognizerResult.ts"],"names":[],"mappings":";AAAA;;GAEG;AACH;;;EAGE"}
```
Stevenic: There's no src file... If you didn't make an changes to core-extensions could you revert all of these lib files?

PR #271, opened 2018-06-06
Johtaylo/choice lang defaults

johnataylor:
sync with c# PR


libraries/botbuilder-prompts/src/choicePrompt.ts
```diff
@@ -105,6 +105,27 @@ export interface ChoicePrompt<O = FoundChoice> {
     recognize(context: TurnContext, choices: (string|Choice)[]): Promise<O|undefined>;
 }
 
+function inlineChoiceOptions(culture?: string): ChoiceFactoryOptions
```
Stevenic: This needs a comment of /** @private */ otherwise it will end up in the docs.

libraries/botbuilder-prompts/src/choicePrompt.ts
```diff
@@ -129,7 +150,7 @@ export interface ChoicePrompt<O = FoundChoice> {
 export function createChoicePrompt<O = FoundChoice>(validator?: PromptValidator<FoundChoice, O>, defaultLocale?: string): ChoicePrompt<O> {
     return {
         style: ListStyle.auto,
-        choiceOptions: {},
+        choiceOptions: inlineChoiceOptions(defaultLocale),
```
Stevenic: >inlineChoiceOptions(defaultLocale), [](start = 23, length = 35)

Seems like you want this to be a map that's consulted at prompt execution time.  The current approach would always use the same options regardless of the users preferred language.

johnataylor: 1) This PR makes the JavaScript the same as the C#. However, we actually want the dynamic behavior in both implementations.
2) The extra yes/no localization in C# for the confirm prompt should also be in here for JavaScript - and again we would want that to be dynamic (i.e. set in the method not the constructor.)

PR #282, opened 2018-06-18
Port Token Endorsements

enzocanoo:
Based on C# work (https://github.com/Microsoft/botbuilder-dotnet/pull/655)


enzocanoo:
We noticed that Travis does not run any tests from `botframework-connector`.

```
"test:travis": "lerna run build && nyc mocha \"libraries/botbuilder*/tests/*.test.js\" && nyc report --reporter=text-lcov | coveralls",
```

from main [package.json](https://github.com/Microsoft/botbuilder-js/blob/master/package.json#L8)

PR #291, opened 2018-07-11
Update the luis-bot-es6 example for botbuilder 4.0.0-preview1.2

gwilymhumphreys:
I believe this fixes up some problems from changes to the botbuilder api. 

Also removed .then / .catch statements in favour of await.

### DIFF
--- a/samples/luis-bot-es6/app.js
+++ b/samples/luis-bot-es6/app.js
@@ -49,90 +49,91 @@ adapter.use(state);
 
 // Listen for incoming requests
 server.post('/api/messages', (req, res) => {
+
     // Route received request to adapter for processing
     adapter.processActivity(req, res, async context => {
         if (context.activity.type === 'conversationUpdate' && context.activity.membersAdded[0].name === 'Bot') {
             // Welcome message here, bot will message you first
-            await context.sendActivity(`Hi! I'm a simple reminder bot. I can help add reminders, delete and show them.`);
+            return await context.sendActivity(`Hi! I'm a simple reminder bot. I can help add reminders, delete and show them.`);
         } else if (context.activity.type === 'message') {
+
+            // Handle a message from the user
             const utterance = (context.activity.text || '').trim().toLowerCase();
 
             // Create dialog context
             const dc = dialogs.createContext(context, state);
 
             // Call LUIS model
-            await model
-                .recognize(context)
-                .then(res => {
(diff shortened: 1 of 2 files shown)


samples/luis-bot-es6/app.js
```diff
@@ -49,90 +49,91 @@ adapter.use(state);
...
-                                    `Hi! I'm a simple reminder bot. I can help add reminders, delete and show them.`
-                                );
-                            }
-                        });
+                        await context.sendActivity(`Nothing to cancel.`);
+                    }
+
+                    // Continue current dialog
+                } else {
+                    const res = await dc.continue()
```
Stevenic: >const res = [](start = 20, length = 11)

Why are you capturing this? You don't do anything with it and honestly its not needed.

samples/luis-bot-es6/app.js
```diff
@@ -49,90 +49,91 @@ adapter.use(state);
 
 // Listen for incoming requests
 server.post('/api/messages', (req, res) => {
+
     // Route received request to adapter for processing
     adapter.processActivity(req, res, async context => {
         if (context.activity.type === 'conversationUpdate' && context.activity.membersAdded[0].name === 'Bot') {
             // Welcome message here, bot will message you first
-            await context.sendActivity(`Hi! I'm a simple reminder bot. I can help add reminders, delete and show them.`);
+            return await context.sendActivity(`Hi! I'm a simple reminder bot. I can help add reminders, delete and show them.`);
```
Stevenic: >return [](start = 12, length = 6)

You don't technically need the return here given your code.

samples/luis-bot-es6/app.js
```diff
@@ -49,90 +49,91 @@ adapter.use(state);
...
         } else if (context.activity.type === 'message') {
+
+            // Handle a message from the user
             const utterance = (context.activity.text || '').trim().toLowerCase();
 
             // Create dialog context
             const dc = dialogs.createContext(context, state);
```
Stevenic: >const dc = dialogs.createContext(context, state); [](start = 12, length = 49)

This isn't being setup correctly.  Your not actually getting the state object and its kind of working because you're using JS. TypeScript would have given an error.  Look at the dialog sample for how to properly get the state.

Stevenic:
Closing this PR for now as we await additional fixes.

PR #293, opened 2018-07-13
Stevenic/properties: DO NOT COMMIT

Stevenic:
No unit tests or docs yet but agreed upon state additions.


libraries/botbuilder-core-extensions/src/botState.ts
```diff
@@ -57,7 +57,19 @@ export class BotState<T extends StoreItem = StoreItem> implements Middleware {
      * @param storageKey Function called anytime the storage key for a given turn needs to be calculated.
      */
     constructor(protected storage: Storage, protected storageKey: StorageKeyFactory) { }
+
+    /** NEW */
+    public readonly properties = new Map();
```
tomlm: >ap() [](start = 38, length = 4)

Make this a Dictionary<string, PropertyAccessor<T>>

stevengum:
I've begun adding tests and have also removed `PropertyAccessor.has()`.

stevengum:
> - BotState implements IPropertyContainer, with SimplePropertyAccessor which pulls data from an IPropertyContainer. This is from discussion with Steve and essentially isolates further the implementation of where a propertyaccessor is getting data from.

Tom dropped the IPropertyContainer, so I didn't go ahead with it here. ([commit 1fd505d](https://github.com/Microsoft/botbuilder-dotnet/pull/760/commits/1fd505d7f5535bbecc3c0aa60906335113235929))

> - BotState.Clear in JS code does not return a promise, even though it seems like correct implementation is to call into storage to clear the state. I made it ClearAsync() in the C# code, but we should discus.

This goes into the whole async vs non-async discussion that we've been having, in the end I decided to leave the current JS implementation synchronous as it just clears the current state (on the TurnContext) and doesn't do any IO.

> - BotState.CreateProperty() default value is a little bit trickier in C#, because it isn't as easy to simply clone an object as it is in JS. If you notice I have a bit of heuristics to make a copy, but it might be a lot clearer to have a delegate to ask for a new default value. Thoughts?

Since it's possible in JS to just clone the object I left that current implementation. However, I did create a DCR Microsoft/botbuilder-dotnet#791, that would add functionality to the current delegate used in C#. If we take that DCR Steve or I can change the implementation in JS.

Finally I did some additional tslinting on the code already being touched by this PR.

PR #299, opened 2018-07-17
Manual fix for botframework-connector HttpOperationResponse.bodyAsJson references

stevengum:
Manual update of the references to `HttpOperationResponse.bodyAsJson`. Change to `HttpOperationResponse.parsedBody`.

I tried regenerating the code with the latest version of autorest and its typescript extension, but this led to different errors when trying to build the packages.

This doesn't solve the current breaks in the [build](https://travis-ci.org/Microsoft/botbuilder-js/builds/405019255), but Travis will at least get past the build step and run the tests.

___

**Note:** Commit that renamed property in [ms-rest-js](https://github.com/Azure/ms-rest-js/commit/4df2d140090b81718e9b3b334963aa36bc9dc5fb#diff-4009293406012034994cbddfed54b63a).


stevengum:
Closing as we are likely going to take #302.

Issue #92, opened 2018-02-21
bug: @types/react package breaks build


Stevenic:
I can't figure out where this is coming from but if you sync and do `lerna bootstrap --hoist` you will get an `@types/react` package added to the root `node_modules` directory which will prevent any of the packages from building.  Manually deleting this package fixes the build break.

I can't seem to figure out where it's getting pulled in from and furthermore have no clue as to why it causes every build to break given that none of the packages being built depend on it.  If anyone has cycles too dig into this further that would be awesome.

billba:
It's coming from [the Web Chat sample](https://github.com/Microsoft/botbuilder-js/tree/master/samples/alarmbot-es6-botframework-webchat) which includes `botframework-webchat` which currently includes `@types/react` as a dev dependency. This is a known issue with Web Chat because it frequently causes versioning issues when included as a component in React apps. I haven't tried `lerna` yet so I haven't seen this problem.

Stevenic:
I assumed it was something from WebChat. Just sucks the way TyeScript pulls everything in thats under `node_modules/@types` whether you're using it or not.  It's not so much a lerna issue as it is a TS issue... We have a single root `node_modules` directory that everything gets cached into (that's the lerna piece) for performance and to better manage cross package dependencies.  I've just been deleting it before building the packages.

Stevenic:
Things are better now but this still crops up occasionally. closing for now.

PR #308, opened 2018-07-26
move core and core-extensions into botbuilder

stevengum:
- Moved `botbuilder-core`, `botbuilder-core-extensions` into `botbuilder`.
- Updated references that were pointing at core/core-extensions to now point at just botbuilder.
- Added additional tests
- Added typescript to the `before_install` portion of the `.travis.yml`
- Changed botbuilder-prompts to use two specific types, `"node"` and `"mocha"`. This should prevent running into #92.


stevengum:
@Kaiqb please take a look.

PR #321, opened 2018-08-10
Update LUIS interface to match C# including adding type/subtype and c…

chrimc62:
…alling

pattern in preperation for moving to official swagger generated SDK.

### DIFF
--- a/libraries/botbuilder-ai/src/luisRecognizer.ts
+++ b/libraries/botbuilder-ai/src/luisRecognizer.ts
@@ -99,8 +99,14 @@ export class LuisRecognizer {
      */
     constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean) {
         this.application = application;
-        this.options = Object.assign({}, options);
-        this.includeApiResults = includeApiResults;
+        this.options = Object.assign({
+            includeAllIntents: false,
+            includeInstanceData: true,
+            log: true,
+            spellCheck: false,
+            staging: false
+        }, options);
+        this.includeApiResults = !!includeApiResults;
 
         // Create client and override callbacks
         // TODO: Update this to the official SDK once available
@@ -125,7 +131,7 @@ export class LuisRecognizer {
                 {
                     timezoneOffset: this.options.timezoneOffset,
                     verbose: this.options.includeAllIntents,
-                    allowSampling: this.options.log && this.options.log.toString()
+                    allowSampling: this.options.log ? "true" : "false"
                 }
             )
                 .then((luisResult: LuisResult) => {
@@ -186,7 +192,7 @@ export class LuisRecognizer {
             recognizerResult: recognizerResult,
             luisResult: luisResult,
             luisOptions: {
-                Staging: this.options && this.options.staging
+                Staging: this.options.staging
             },
             luisModel: {
                 ModelID: this.application.applicationId


libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -71,19 +85,26 @@ export interface LuisRecognizerSettings {
...
-     * @param settings Settings used to configure the instance.
+     * @param application LUIS application to use.
+     * @param options Options used to control predictions.
      */
-    constructor(settings: LuisRecognizerSettings) {
-        this.settings = Object.assign({}, settings);
+    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean) {
+        this.application = application;
+        this.options = Object.assign({}, options);
+        this.includeApiResults = includeApiResults;
```
Stevenic: >includeApiResults [](start = 33, length = 17)

minor nit but given this can be undefined you can use "!!includeApiResults" to ensure that its assigned as a boolean.

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -99,15 +120,22 @@ export class LuisRecognizer {
         const cached = context.services.get(this.cacheKey);
         if (!cached) {
             const utterance = context.activity.text || '';
-            return this.luisClient.getIntentsAndEntitiesV2(this.settings.appId, this.settings.subscriptionKey, utterance, this.settings.options)
+            return this.luisClient.getIntentsAndEntitiesV2(
+                this.application.applicationId, this.application.endpointKey, utterance,
+                {
+                    timezoneOffset: this.options.timezoneOffset,
+                    verbose: this.options.includeAllIntents,
+                    allowSampling: this.options.log && this.options.log.toString()
```
Stevenic: >this.options.log && this.options.log.toString() [](start = 35, length = 47)

what is this check for? Are you trying to validate that it was set?

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -158,10 +186,10 @@ export class LuisRecognizer {
             recognizerResult: recognizerResult,
             luisResult: luisResult,
             luisOptions: {
-                Staging: this.settings.options && this.settings.options.staging
+                Staging: this.options && this.options.staging
```
Stevenic: >this.options [](start = 25, length = 12)

you should always have a this.options given you set it in your constructor

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -71,19 +85,26 @@ export interface LuisRecognizerSettings {
...
      * Creates a new LuisRecognizer instance.
-     * @param settings Settings used to configure the instance.
+     * @param application LUIS application to use.
+     * @param options Options used to control predictions.
      */
-    constructor(settings: LuisRecognizerSettings) {
-        this.settings = Object.assign({}, settings);
+    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean) {
+        this.application = application;
+        this.options = Object.assign({}, options);
```
Stevenic: >{} [](start = 37, length = 2)

If it simplifies things you can pre-set your option default values here. That way they'll only get changed if the caller assigns them.

PR #324, opened 2018-08-13
[DO NOT MERGE] Fix botbuilder-dialog tests

enzocanoo:
We've found that the [Travis-CI](https://travis-ci.org/Microsoft/botbuilder-js/branches) build is failing in the `master` branch after some changes to the dialogs package were merged (PR #320)

Unit tests impacted by the changes to dialogs started failing because they need to be adapted to the refactored version of `Dialog`, `DialogSet` and `DialogState` .

Only unit tests code is needed to fix the build, but what seemed like a trivial fix with minor unti tests code changes proved more complex.

@Stevenic: We created this PR to validate the approach and the fixes with you before moving forward with the rest of the failing unit tests. We are aware of upcoming changes to the dialogs dependency that might involve some rework afterwards.


stevengum:
The results of these specific tests in the Travis build can be found here: https://travis-ci.org/Microsoft/botbuilder-js/builds/415662207#L912

enzocanoo:
Tests already fixed in master.

PR #325, opened 2018-08-14
Stevenic/remove libs

Stevenic:
Removed checked in lib files to help reduce merge conflicts.


libraries/botframework-luis/.gitignore
```diff
@@ -1,5 +1,6 @@
 /**/node_modules
 /**/.vscode
+/**/lib
```
stevengum: The `lib` folder for botrframework-luis still [exists](https://github.com/Microsoft/botbuilder-js/tree/stevenic/remove-libs/libraries/botframework-luis/lib), should we be adding it to the `.gitignore`?

PR #329, opened 2018-08-15
add botframework-config package and unit tests

tomlm:
pull classes out of botbuilder-tools and into botbuilder-config package
Add unit tests


justinwilaby:
Can we remove the libs  directory? It makes the PR's more cluttered.

libraries/botframework-config/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:Microsoft/botbuilder-js.git"
+  },
+  "devDependencies": {
+    "@types/uuid": "^3.4.3",
+    "mocha": "^5.0.5"
+  },
+  "dependencies": {
+    "linq-collections": "^1.0.254",
```
justinwilaby: This dependency is heavy in comparison to other more lightweight options. [node-klaw](https://github.com/jprichardson/node-klaw) is trusted, extremely lightweight and more widely know in the JS community.  We might consider this when taking on dependencies.

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -0,0 +1,230 @@
...
+
+        for (let service of this.services) {
+            (<ConnectedService>service).decrypt(secret);
+        }
+    }
+
+
+    // remove service by name or id
+    public disconnectServiceByNameOrId(nameOrId: string): IConnectedService {
+        let svs = new List<IConnectedService>(this.services);
```
justinwilaby: This can be simplified by the following:
```js
const {services = []} = this;
let i = services.length;
while (i--) {
  const service = services[i];
 if (service.id == nameOrId || service.name == nameOrId) {
    return services.splice(i, 1)[0];
  }
}
```
It also does not require taking on a dependency.

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -0,0 +1,230 @@
...
+            }
+        }
+        throw new Error(`a service with id or name of [${nameOrId}] was not found`);
+    }
+
+    // remove a service
+    public disconnectService(type: string, id: string): void {
+        let svs = new List<IConnectedService>(this.services);
+
+        for (let i = 0; i < svs.count(); i++) {
```
justinwilaby: Same here. Consider using `Array.prototype.splice` in a `while` loop instead.

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    private legacyDecrypt(encryptedValue: string, secret: string): string {
+        // LEGACY for pre standardized SHA256 encryption, this uses some undocumented nodejs MD5 hash internally and is deprecated
+        const decipher = crypto.createDecipher('aes192', secret);
+        let value = decipher.update(encryptedValue, 'hex', 'utf8');
+        value += decipher.final('utf8');
+        return value;
+    }
+
+    public static serviceFromJSON(service: IConnectedService): ConnectedService {
```
justinwilaby: Since this static is a member of a node-only class, consider moving this to a utility class that can be imported and used exclusively.

justinwilaby: Scratch that. I see now that the data models are also node-exclusive. Is this what we want? The emulator uses these data models on both the client and node side. There are also scenarios I can come up with where these would be useful in a web-based bot. Would it make sense to move encryption/decryption out of the data models so they can be agnostic to the platform?

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -0,0 +1,230 @@
...
+            this.decrypt(secret);
+    }
+
+    public clearSecret() {
+        this.secretKey = '';
+    }
+
+    // connect to a service
+    public connectService(newService: IConnectedService): void {
+        if (Enumerable.fromSource(this.services)
```
justinwilaby: This can be simplified using the following: 
```js
const alreadyConnected = this.services.some(service => service.type === newService.type && service.id === newService.id
```
with the added benefit of not needing a dependency.

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -0,0 +1,230 @@
...
+            // give unique name
+            let nameCount = 1;
+            let name = newService.name;
+
+            while (true) {
+                if (nameCount > 1) {
+                    name = `${newService.name} (${nameCount})`;
+                }
+
+                if (!Enumerable.fromSource(this.services).where(s => s.name == name).any())
```
justinwilaby: This loop is nested. Consider using a map and reference it's keys:
```js
const namesMap = this.services.reduce((map, service) => (map[name] = true, map), {});

while (true) {
  ...
  if (namesMap[name]) {
    break;
  }
}
...
```

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -0,0 +1,230 @@
...
+        return botConfig;
+    }
+
+    public toJSON(): Partial<IBotConfiguration> {
+        const { name, description, services, secretKey } = this;
+        return { name, description, services, secretKey };
+    }
+
+    public static async loadBotFromFolder(folder?: string, secret?: string): Promise<BotConfiguration> {
+        let files = Enumerable.fromSource(await fsx.readdir(folder || process.cwd()))
```
justinwilaby: Consider using [node-klaw](https://github.com/jprichardson/node-klaw) for directory contents traversal. It's very lightweight, trusted and widely known in the js community.

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -0,0 +1,230 @@
...
+            this.validateSecretKey(secret);
+        }
+
+        let hasSecret = (this.secretKey && this.secretKey.length > 0);
+
+        // make sure that all dispatch serviceIds still match services that are in the bot
+        for (let service of this.services) {
+            if (service.type == ServiceTypes.Dispatch) {
+                let dispatchService = <IDispatchService>service;
+                dispatchService.serviceIds = Enumerable.fromSource(dispatchService.serviceIds)
```
justinwilaby: This can be simplified by the following:
```js
const serviceIdMap = this.services.reduce((map, service) => (map[service.id] = true, map), {});
dispatchService.serviceIds = dispatchService.serviceIds.filter(id => serviceIfMap[id]);
```
*Does not require nesting a loop or a dependency*

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -0,0 +1,230 @@
...
+                this.services = svs.toArray();
+                return;
+            }
+        }
+    }
+
+
+    // make sure secret is correct by decrypting the secretKey with it
+    public validateSecretKey(secret: string): void {
+        if (!secret || secret.length == 0) {
```
justinwilaby: a string with a length of zero resolves to false. It's safe to remove `|| string.length == 0`

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -0,0 +1,230 @@
...
+            throw new Error('You are attempting to perform an operation which needs access to the secret and --secret is missing');
+        }
+
+        try {
+            if (!this.secretKey || this.secretKey.length == 0) {
```
justinwilaby: Same here. Empty strings resolve to false making a length check superfluous.

PR #334, opened 2018-08-16
Add new Attachment and Confirm Prompt tests

stevengum:


libraries/botbuilder-dialogs/tests/attachmentPrompt.test.js
```diff
@@ -0,0 +1,180 @@
...
+        });
+
+        const convoState = new ConversationState(new MemoryStorage());
+        adapter.use(convoState);
+
+        const dialogState = convoState.createProperty('dialogState');
+        const dialogs = new DialogSet(dialogState);
+        dialogs.add(new AttachmentPrompt('prompt'));
+        
+        adapter.send('Hello')
```
stevengum: There's a difference in how this test operates versus the ConfirmPrompt tests because of how the ConfirmPrompt calls the ChoiceFactory through the base Prompt class (via [Prompt.appendChoices()][appendChoices]). @stevenic do you have a recommendation for how to get the two to act similarly?


  [appendChoices]: https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-dialogs/src/prompts/prompt.ts#L164

PR #337, opened 2018-08-16
Change to deploy to MyGet only on cron jobs

BruceHaley:


Stevenic:
If you look @BruceHaley you'll see that the build still fails

PR #343, opened 2018-08-17
Added missing event names to continueConversation() and createConversation()

Stevenic:
Noticed that the event names were missing.


stevengum:
There was a merge conflict due to my merging in #345 before this one... The conflict has been resolved.

PR #344, opened 2018-08-17
Fix dropped choices bug in DialogContext.prompt(), reenable ChoicePrompt and DateTimePrompt tests

stevengum:


libraries/botbuilder-dialogs/src/dialogContext.ts
```diff
@@ -135,7 +135,8 @@ export class DialogContext {
     public async prompt(dialogId: string, promptOrOptions: string|Partial<Activity>|PromptOptions, choices?: (string|Choice)[]): Promise<DialogTurnResult> {
         let options: PromptOptions;
         if (typeof promptOrOptions === 'object' && (promptOrOptions as PromptOptions).prompt !== undefined) {
-            options = Object.assign({}, promptOrOptions as PromptOptions);
```
Stevenic: This was actually kind of by design. Now you can either pass in a prompt & choices or just options but not both.  Looking at this though we probably should change this test up some because I did make "PromptOptions.prompt" an optional member.

It might be better to reverse this and look for promptOrOptions to either be a string or an object with a "type" property.

stevengum: Going down the Activity-checking route, we could change the method to look like this:

```js
public async prompt(dialogId: string, promptOrOptions: PromptOptions): Promise<DialogTurnResult>
public async prompt(dialogId: string, promptOrOptions: string|Partial<Activity>,
    choices?: (string|Choice)[]): Promise<DialogTurnResult> {
  let options: PromptOptions;
  if ((typeof promptOrOptions === 'object' && (promptOrOptions as Activity).type !== undefined) || 
    typeof promptOrOptions === 'string') {
    options = { prompt: promptOrOptions as string|Partial<Activity>, choices: choices };
  } else {
    options = Object.assign({}, promptOrOptions as PromptOptions);
  }
  return this.begin(dialogId, options);
}
```

PR #350, opened 2018-08-17
Remove travis caching of node_modules

JuanAr:
- Travis had been caching the node_modules directory entirely, which means it didn't detect when a dependency change broke stuff from under us
- One of those dependency changes was an update to `ms-rest` which should be making the build fail.
- We might want to revisit the rationale to cache node_modules and try another tooling like [`yarn`](https://yarnpkg.com/en/)

We created this PR to start the discussion on how to better deal with caching and perf.


stevengum:
Speaking from personal observations on this project, caching the `node_modules/` folders has blocked some PRs. As someone with access to Travis for this repo, I can clear caches as needed, so it's only a minor pain as opposed to a major pain.

I'm not entirely sure how the caching works for PRs coming in from forks, but I think a slightly longer build would be better than potentially blocking community contributions for trivial reasons.

@justinwilaby, can you chime in on your experiences with Travis and caching `node_modules/`? @stevenic what do you think?

PR #351, opened 2018-08-18
Switch to the official LUIS node SDK.

chrimc62:


stevengum:
Cross linking for tracking information [**azure-sdk-for-node/lib/services/luis**](https://github.com/Azure/azure-sdk-for-node/tree/master/lib/services/luis)

PR #364, opened 2018-08-23
Update to 2.echobot-with-counter

yochay:
Added bunch of comments and made minor code update.


vishwacsena:
We should not take this change into master. The samples need to be in botbuilder-samples instead of botbuilder-js.

Issue #314, opened 2018-08-01
Blob Storage conditional header(s) is not met


luspac:
I am having an issue with Blob Storage. 

I am able to read and write information to Blob Storage but receive a warning:
Unhandled promise rejection (rejection id: 1): StorageError: The condition specified using HTTP conditional header(s) is not met.

To access the Blob Storage I pass in my settings. 
```
const {BlobStorage} = require('botbuilder-azure');

var storage = new BlobStorage({
    containerName: '<container name>',
    storageAccountOrConnectionString: '<account or connection string>',
    host: '<host>'
});
```

The definition of BlobStorage requires a storageAccessKey but I immediately get thrown an error if I add that parameter to the Blob Storage connection. 

The way in which I am accessing the storage is handled in this function below. Any insight into this would be much appreciated. 

```
async function logMessageText(storage, context) {
    let utterance = context.activity.text;
    let storeItems = await storage.read(["UtteranceLog"])
    console.log(storeItems);
    try {
        // check result
        var utteranceLog = storeItems["UtteranceLog"];

        if (typeof (utteranceLog) != 'undefined') {
            // log exists so we can write to it
            storeItems["UtteranceLog"].UtteranceList.push(utterance);

            try {
                await storage.write(storeItems)
                context.sendActivity('Successful write to utterance log.');
            } catch (err) {
                context.sendActivity(`Write failed of UtteranceLog: ${err}`);
            }

        } else {
            context.sendActivity(`need to create new utterance log`);
            storeItems["UtteranceLog"] = { UtteranceList: [`${utterance}`]}

            try {
                await storage.write(storeItems)
                context.sendActivity('Successful write to log.');
            } catch (err) {
                context.sendActivity(`Write failed: ${err}`);
            }
        }
    } catch (err) {
        context.sendActivity(`Read rejected. ${err}`);
    };
}
```

xtellurian:
I'm seeing the same issue.

```js
const adapter = new BotFrameworkAdapter({
    appId: process.env.MICROSOFT_APP_ID,
    appPassword: process.env.MICROSOFT_APP_PASSWORD
});

var container_name = process.env.BLOB_CONTAINER; // container name
var storageAccountName = process.env.STORAGE_ACCOUNT_NAME; // single string, eg, 'test'
var storageAccountKey = process.env.STORAGE_ACCOUNT_KEY; //base64 encoded key
[...]

Issue #297, opened 2018-07-13
BotState should be POCO


tomlm:
BotState in C# was changed to not require .ETag, and in fact just store POCO objects with LastWrite semantics.

JS code still uses StoreItem definition which defines ETAG semantics. 

Change:
* lib should be POCO with no ETag (dropping StoreItem)

PR #369, opened 2018-08-23
fix BlobStorage.read to initialize new blob if blob doesn't exist

stevengum:
Should address #314...


libraries/botbuilder-azure/src/blobStorage.ts
```diff
@@ -117,14 +117,27 @@ export class BlobStorage implements Storage {
...
                 Promise.all<DocumentStoreItem>(sanitizedKeys.map((key) => {
-                    return new Promise((resolve, reject) => {
-                        this.client.getBlobMetadataAsync(container.name, key).then((blobMetadata) => {
-                            this.client.getBlobToTextAsync(blobMetadata.container, blobMetadata.name).then((result) => {
-                                let document: DocumentStoreItem = JSON.parse(result as any);
-                                document.document.eTag = blobMetadata.etag;
-                                resolve(document);
-                            }, err => resolve(null));
-                        }, (err) => resolve(null));
+                    return this.checkForBlob(container.name, key).then((blobResult) => {
```
johnataylor: no need to define a one line function here - just put the check inline (all the other code uses client)

libraries/botbuilder-azure/src/blobStorage.ts
```diff
@@ -117,14 +117,27 @@ export class BlobStorage implements Storage {
...
+                        if (blobResult.exists) {
+                            return this.client.getBlobMetadataAsync(container.name, key).then((blobMetadata) => {
+                                return this.client.getBlobToTextAsync(blobMetadata.container, blobMetadata.name).then((result) => {
+                                    const document: DocumentStoreItem = JSON.parse(result as any);
+                                    return document;
```
johnataylor: don't we still need the eTag?

libraries/botbuilder-azure/src/blobStorage.ts
```diff
@@ -222,6 +234,10 @@ export class BlobStorage implements Storage {
         return checkedCollections[key];
     }
 
+    private checkForBlob(containerName: string, blob: string): Promise<azure.BlobService.BlobResult> {
```
johnataylor: see previous comment - pulling this code into this function adds little value

libraries/botbuilder-azure/src/blobStorage.ts
stevengum: This is the document from blobStorage, so it already has an eTag in the current implementation.

There is a related issue saying that BotState should just be POCO (#297); so in the code below should we be bothering with eTags at all?

libraries/botbuilder-azure/src/blobStorage.ts
```diff
@@ -117,14 +117,27 @@ export class BlobStorage implements Storage {
...
+                                accessConditions: azure.AccessCondition.generateEmptyCondition(),
+                                parallelOperationThreadCount: 4
+                            } as azure.BlobService.CreateBlobRequestOptions;
+                            return this.client.createBlockBlobFromTextAsync(container.name, blobResult.name, '{}', createBlobOptions).then((creationResult) => {
+                                return this.client.getBlobToTextAsync(creationResult.container, creationResult.name).then((result) => {
+                                    const document: DocumentStoreItem = JSON.parse(result as any);
+                                    document.document = { eTag: '*' };
+                                    return document;
+                                });
+                            });
```
tomlm: Why are we doing this delta at all?  God sheds a tear when a blob is wasted.

libraries/botbuilder-azure/src/blobStorage.ts
stevengum: The two (#297 and this) aren't mutually exclusive so I added the eTag semantics back to the PR. The PR is much simpler now than it was originally.

libraries/botbuilder-azure/src/blobStorage.ts
stevengum: Done!

stevengum:
The PR is much simpler now, and `.read()` will return an empty DocumentStoreItem if the blob does not exist. `BlobStorage.read()` no longer swallows Errors and the tests now show the actual Error.message instead of just displaying an Error object.

@johnataylor, @tomlm

PR #370, opened 2018-08-24
removed translation and locale middleware

johnataylor:
removed translation and locale middleware


johnataylor:
Added branch https://github.com/Microsoft/botbuilder-js/tree/johtaylo/translation-sample with these file on it.

Issue #349, opened 2018-08-17
Address linter issues  


benbrown:
tslint is now highlighting a bunch of code hygiene issues.

we need to address as many as possible, and adjust the rules to reflect our real goals for code style.

stevengum:
Not sure if this is the right place for it; but @benbrown and @johnataylor should we disable the TSLint if the mocha tests get a `'1'` status code?  I suppose the first question is actually "can we stop the travis build if a unit test fails?"

stevengum:
Disregard previous comment; we should run both regardless so that diligent devs clean up their code before it gets merged, even if the build breaks on a unit test... We should in the future evaluate whether or not tslint should throw errors or just warnings.

benbrown:
Work in progress in the `benbrown/linter` branch fyi

benbrown:
I've gone through all the files in `botbuilder` and `botbuilder-ai` now and updated them to the best of my ability. There are a few warnings still coming out of each project, which I'm still working on.

benbrown:
`botbuilder-dialogs` is now updated, and I've also merged in the recent changes from master.

benbrown:
The vast majority of linter errors have been corrected, and the linter rules have been updated. This has been merged!

PR #376, opened 2018-08-27
Update all of the typescript files to be (virtually) linter error free

benbrown:
I updated almost every single file in this repo with minor typescript style fixes. This included:

* Updating files to match linter syntax expectations
* Minor changes to some syntax (things like using object spread notation instead of object.assign)
* Added some missing dependencies to packages
* Updated the tslint.json file to remove some rules from the linting process

I turned off a rules in the tslint.json file that seemed like they would be very hard to enforce, or if they would potentially compromise some other part of the build process. For example, `no-redundant-jsdoc` complains about a lot of documentation comments that are used in the build process.   The rules I disabled definitely need to be reviewed, and if we need to turn some back on, I'm game to take a second (or third) pass!

Overall, this brought the number of linter errors down from somewhere near 4500 to about 20.  Those 20 errors probably deserve to  be looked at again, though are mostly things like console.logs and a few TODO: comments.


stevengum:
Crosslinking #349.

@benbrown in the future could you keep your PRs  to a smaller amount of commits? (One of the tools that I know @stevenic likes to use supports only 64 commits when reviewing PRs)

I learned about the limit today when I was going to use it to review this massive and very much needed PR. 😄

benbrown:
@stevengum - yes absolutely. I was trying to be fairly granular with commits because I wanted to be able to backtrack if tests started to fail.

libraries/botbuilder-dialogs/src/waterfallDialog.ts
```diff
@@ -108,20 +109,23 @@ export class WaterfallDialog<O extends object = {}> extends Dialog<O> {
...
             const step: WaterfallStepContext<O> = {
                 index: index,
                 options: <O>state.options,
                 reason: reason,
                 result: result,
                 values: state.values,
-                next: async (result?: any) => {
-                    if (nextCalled) { throw new Error(`WaterfallStepContext.next(): method already called for dialog and step '${this.id}[${index}]'.`) }
-                    return await this.dialogResume(dc, DialogReason.nextCalled, result);
+                next: async (result2?: DialogTurnResult<any>): Promise<DialogTurnResult<any>> => {
```
stevengum: I don't know if this is correct, and as someone who prefers more descriptive variable and argument names, `result2` isn't terribly helpful.

I think the more correct code here is this:

```js
next: async (stepResult?: any): Promise<DialogTurnResult<any>> => {
```
Otherwise TypeScript devs will have to return a pass in a `DialogTurnResult` which is in turn wrapped in another `DialogTurnResult`.

libraries/botbuilder/src/consoleAdapter.ts
```diff
@@ -196,15 +206,17 @@ export class ConsoleAdapter extends BotAdapter {
      * Logs text to the console.
      * @param line Text to print.
      */
-    protected print(line: string) {
+    protected print(line: string): void {
+        // tslint:disable-next-line:no-console
```
stevengum: If you use `// tslint:disable:no-console`, it will actually disable the rule throughout the file so you don't need to add it for [every line](https://palantir.github.io/tslint/usage/rule-flags/).

libraries/botbuilder/src/consoleAdapter.ts
```diff
@@ -196,15 +206,17 @@ export class ConsoleAdapter extends BotAdapter {
...
         console.log(line);
     }
 
     /**
      * Logs an error to the console.
      * @param line Error text to print.
      */
-    protected printError(line: string) {
+    protected printError(line: string): void {
+        // tslint:disable-next-line:no-console
```
stevengum: See earlier comment regarding this rule.

libraries/botbuilder/src/fileTranscriptStore.ts
```diff
@@ -54,22 +55,28 @@ export class FileTranscriptStore implements TranscriptStore {
...
+        const transcriptFolder: string = this.getTranscriptFolder(channelId, conversationId);
+
+        return fs.exists(transcriptFolder).then((exists: boolean) => {
             if (!exists) { return pagedResult; }
 
             return fs.readdir(transcriptFolder)
-                .then(files => files
-                    .filter(f => f.endsWith('.json'))               // .json only
+                .then((files: string[]) => files
+                    .filter((f: string) => f.endsWith('.json'))               // .json only
```
stevengum: Spacing on the comment is no longer in line...

libraries/botframework-config/src/botConfiguration.ts
```diff
@@ -263,48 +314,50 @@ export class BotConfiguration implements Partial<IBotConfiguration> {
                 throw err;
             }
 
-            return service;
+            // return service;
```
stevengum: Why is this commented out?

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -98,21 +98,44 @@ export class LuisRecognizer {
...
         this.luisClient = new LuisClient(creds, baseUri);
     }
 
+    /**
+     * Returns the name of the top scoring intent from a set of LUIS results.
+     * @param results Result set to be searched.
+     * @param defaultIntent (Optional) intent name to return should a top intent be found. Defaults to a value of `None`.
+     * @param minScore (Optional) minimum score needed for an intent to be considered as a top intent. If all intents in the set are below this threshold then the `defaultIntent` will be returned.  Defaults to a value of `0.0`.
+     */
+    // tslint:disable-next-line:function-name
```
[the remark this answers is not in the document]
benbrown: concur!

libraries/botbuilder-azure/src/azureBlobTranscriptStore.ts
```diff
@@ -307,14 +386,15 @@ export class AzureBlobTranscriptStore implements TranscriptStore {
             getBlobToTextAsync: this.denodeify(blobService, blobService.getBlobToText),
             deleteBlobIfExistsAsync: this.denodeify(blobService, blobService.deleteBlobIfExists),
             listBlobsSegmentedWithPrefixAsync: this.denodeify(blobService, blobService.listBlobsSegmentedWithPrefix),
+            // tslint:disable-next-line:no-string-literal
```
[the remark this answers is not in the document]
benbrown: This one I included because `listBlobDirectoriesSegmentedWithPrefix` doesn't seem to be a declared property on blobService.  I wasn't sure how to remedy that -- there is a subclass defined at the bottom of that file, but it looks to be defining the async versions of existing functions. Any ideas?

libraries/botbuilder-core/src/internal.ts
```diff
@@ -6,22 +6,26 @@
 /**
  * @private
  */
- export function shallowCopy<T>(value: T): T {
+export function shallowCopy<T>(value: T): T {
     if (Array.isArray(value)) { return value.slice(0) as any; }
+    // tslint:disable-next-line:prefer-object-spread
```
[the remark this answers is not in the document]
benbrown: Concur!

libraries/botbuilder-dialogs/src/choices/recognizeChoices.ts
```diff
@@ -55,25 +55,32 @@ export function recognizeChoices(utterance: string, choices: (string|Choice)[],
...
+         }
     }
 
     // Normalize choices
-    const list: Choice[] = (choices || []).map((choice, index) => typeof choice === 'string' ? { value: choice } : choice).filter((choice) => choice);
-    
+    const list: Choice[] = (choices || []).map(
+        (choice: Choice) => typeof choice === 'string' ? { value: choice } : choice
+    ).filter(
+        (choice: Choice) => choice // TODO: does this do anything?
```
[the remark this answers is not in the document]
benbrown: I actually found probably 10 places in the code where this strange filter pattern is in place.  I will open an issue!

libraries/botbuilder-dialogs/src/dialogSet.ts
```diff
@@ -151,69 +151,76 @@ import { DialogContext, DialogState } from './dialogContext';
...
      * ```
      * @param context Context for the current turn of conversation with the user.
      */
     public async createContext(context: TurnContext): Promise<DialogContext> {
-        if (!this.dialogState) { throw new Error(`DialogSet.createContextAsync(): the dialog set was not bound to a stateProperty when constructed.`) }
-        let state = await this.dialogState.get(context, { dialogStack: [] });
+        if (!this.dialogState) {
+            throw new Error(`DialogSet.createContextAsync(): the dialog set was not bound to a stateProperty when constructed.`);
+        }
+        const state: DialogState = await this.dialogState.get(context, { dialogStack: [] });
```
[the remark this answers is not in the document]
benbrown: Makes sense. I've added this.

libraries/botbuilder-dialogs/src/waterfallDialog.ts
benbrown: Thanks for busting me on that lazy variable name!

libraries/botbuilder-core/src/internal.ts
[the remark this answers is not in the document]
benbrown: Ah, the reason this was left as an Object.assign is that TS throws an error about value being declared as a `T`, not an object.

Do you think this works?

`if (typeof value === 'object') { return {...value as any}; }`

libraries/botframework-config/src/botConfiguration.ts
benbrown: It didn't look like it was doing anything there, but I wasn't 100% sure.

It is returning inside a catch block, inside a function with a void return type.

benbrown:
I just merged in recent changes to master and re-fixed a few files.

libraries/botbuilder-azure/src/azureBlobTranscriptStore.ts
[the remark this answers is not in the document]
stevengum: I'll file a cleanup issue for this; searching for the method name, `listBlobDirectoriesSegmentedWithPrefix` in Azure/azure-storage-node revealed a couple of issues and PRs related to adding the method to the d.ts files. 

Issue: Azure/azure-storage-node#504
PR: Azure/azure-storage-node#506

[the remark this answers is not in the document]
stevengum: :shipit:

libraries/botframework-config/src/botConfiguration.ts
stevengum: @tomlm, can you comment on this? The `return service` doesn't look like it's supposed to be used, based on successful completion of the first `try` block and the lack of a return value there.

stevengum:
@benbrown, looks great! Once @tomlm chimes in on `botConfiguration.ts` we can get this merged in.

libraries/botbuilder-dialogs/src/choices/recognizeChoices.ts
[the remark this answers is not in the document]
stevengum: Corresponding issue: #380

cleemullins:
@benbrown This PR is full of amazing goodness! The sooner this gets merged in the better, as it's only going to drift as changes continue to happen. What are we waiting on that's blocking?

libraries/botframework-config/src/botConfiguration.ts
cleemullins: This looks like it should be deleted. If we can't get confirmation from @tomlm, consider leaving it as-is, so we can get the rest of this PR-Of-Happy-Linting commited!!

stevengum: @benbrown, I concur with Chris, let's just revert this line for now to leave the current behavior and get this in.

benbrown:
I just reverted that orphan `return service` line.  Should be GTG!

benbrown:
@stevengum @cleemullins -- being extra careful here, but:

when I put that `return service;` back in, it fails to build since now that variable is properly declared and goes out of scope where it was called previously.

so I pushed another change here where I've removed that line.

cleemullins:
@benbrown Merge it in! 👍

Issue #373, opened 2018-08-24
`botframework-config` Models Depends on Node built-in


justinwilaby:
`botframework-config/src/models` depends on the `crypto` library which is a node built-in. 

This prevents these models from being used on a web client which breaks backwards compatibility when migrating from [msbot](https://github.com/Microsoft/botbuilder-tools/tree/master/packages/MSBot).

My suggestion is to move encryption into a separate file and maintain dependency free data models.

(this currently blocks the [Emulator](https://github.com/microsoft/botframework-emulator) from migrating to this library. )

PR #377, opened 2018-08-27
isolate nodejs functions 

tomlm:
Created BotConfigurationBase which has no node dependencies
Moved all file and crypto functionality to BotConfiguration which derives from BotConfigurationBase.
changed service.encrypt() to take in encrypt function as parameters to isolate them from nodejs crypto requirements

_Edit: Addresses #373_


stevengum:
Crosslinking with #373.

PR #385, opened 2018-08-29
Use lerna 3.2.1 as default

enzocanoo:
Force `lerna@3.2.1` due to an issue with the license of future versions.


enzocanoo:
@cleemullins, we had an issue related to the build cache so we deleted the cache and the build started to work.

iMicknl:
FYI. https://github.com/lerna/lerna/pull/1633

PR #386, opened 2018-08-29
reenable DialogContext tests

stevengum:
and remove inapplicable tests.


libraries/botbuilder-dialogs/tests/dialogContext.test.js
```diff
@@ -1,365 +1,431 @@
...
+            }
+        ]));
 
-        dc.begin('b').catch((err) => {
-            assert(err);
-            done();
+        adapter.send(beginMessage);
+    });
+
+    it('should return error if begin() called with invalid dialog ID.', function (done) {
```
cleemullins: "ID" should be "Id" or "dialogId" in javascript land.

Issue #396, opened 2018-09-04
Some inline code examples in botbuilder-dialogs is out of date


benbrown:
Examples in the typescript files need to be updated with the latest syntax for creating Waterfall dialogs.

benbrown:
Fixed in #406

PR #404, opened 2018-09-06
Fix for #394: Single line comments reverted to comment blocks

### LINKED ISSUE #394: Changing /* comments */ to // comments in some places broke tsdoc output
When I updated the comment format to pass the tslint rules, it broke some docs output, for example on lines line this:

https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/instanceData.ts#L13

benbrown:
This change reverts changes I made to address a linter rule. @stevenic informed me that changing /** comments */ to // comments prevented them from being used in the autogenerated docs.  Oops!

This PR changes those comments back to block comments, and should fix #394


libraries/botbuilder-dialogs/src/dialog.ts
```diff
@@ -61,21 +85,21 @@ export enum DialogTurnStatus {
  * @param T (Optional) type of result returned by the dialog when it calls `dc.end()`.
  */
 export interface DialogTurnResult<T = any> {
-    // Gets or sets the current status of the stack.
+    //** Gets or sets the current status of the stack.
```
stevengum: Shouldn't this be :

```js
/**
 * Gets or sets the current status of the stack.
 */
```

libraries/botbuilder-dialogs/src/dialog.ts
```diff
@@ -61,21 +85,21 @@ export enum DialogTurnStatus {
...
     status: DialogTurnStatus;
 
-    // Final result returned by a dialog that just completed. Can be `undefined` even when [hasResult](#hasResult) is true.
+    //** Final result returned by a dialog that just completed. Can be `undefined` even when [hasResult](#hasResult) is true.
```
stevengum: See earlier comment.

libraries/botbuilder-dialogs/src/dialogContext.ts
```diff
@@ -1,90 +1,92 @@
...
  * @remarks
- * This is typically created through a call to `DialogSet.createContext()` and is then passed 
+ * This is typically created through a call to `DialogSet.createContext()` and is then passed
  * through to all of the bots dialogs and waterfall steps.
- * 
+ *
  * ```JavaScript
  * const conversation = conversationState.get(context);
- * const dc = dialogs.createContext(context, conversation);  
+ * const dc = dialogs.createContext(context, conversation);
```
stevengum: I really want you to change this comment as it's out of date; but that's out of scope for this PR... Do we have an issue for out of date doc strings?

libraries/botbuilder-dialogs/src/dialog.ts
```diff
@@ -61,21 +85,21 @@ export enum DialogTurnStatus {
...
 /**
  * Base class for all dialogs.
  */
 export abstract class Dialog<O extends object = {}> {
-    // Signals the end of a turn by a dialog method or waterfall/sequence step.
+    //** Signals the end of a turn by a dialog method or waterfall/sequence step.
     public static EndOfTurn: DialogTurnResult = { status: DialogTurnStatus.waiting };
 
-    // Unique ID of the dialog.
+    //** Unique ID of the dialog.
```
stevengum: See earlier comments for this line and the docstring for `Dialog.EndOfTurn`.

libraries/botbuilder-dialogs/src/dialog.ts
benbrown: yes!

libraries/botbuilder-dialogs/src/dialogContext.ts
benbrown: yes, I've started that in #396.  I can update this one though!

benbrown: I updated this one and all the other references to createContext I could find.

Stevenic:
There's a build break in LuisRecognizer. I can look into it this afternoon

benbrown:
Ugh! I see it in the diff - dropped the endpoint property

benbrown:
I just pushed a change that re-adds the endpoint field which had inadvertently been removed.

In its place I noticed an azureRegion field which is NOT currently in master, so can someone sanity check that? @Stevenic

benbrown:
Aha! This explains it: 
https://github.com/Microsoft/botbuilder-js/commit/7b022d1d8d141fb61138ff88e08331a26c607e7c#diff-d746df635d42aa7e8228fc3af83d6852

benbrown:
@cleemullins fixed!

PR #406, opened 2018-09-07
Update inline code samples to reflect new Dialog syntax

benbrown:
This PR updates the code samples to use the new WaterfallDialog syntax.  In addition, I've added required `await`s in a few places.

No running code has changed - only inline code samples in comment blocks.


benbrown:
This should resolve #396

libraries/botbuilder-dialogs/src/dialogSet.ts
```diff
@@ -23,33 +23,34 @@ import { DialogContext, DialogState } from './dialogContext';
...
+ *     async (dc, step) => {
  *         dc.activeDialog.state.profile = {};
- *         await dc.prompt('textPrompt', `What's your name?`);
+ *         return await dc.prompt('textPrompt', `What's your name?`);
  *     },
- *     async function (dc, name) {
- *         dc.activeDialog.state.profile.name = name;
- *         await dc.prompt('textPrompt', `What's your phone number?`);
+ *     async (dc, step) => {
+ *         dc.activeDialog.state.profile.name = step.results;
```
v-ducvo: The `dc.activeDialog.state` should now be `step.values` … so this line should be updated to `step.values.profile.name = step.result.`  … oh, and, it should singular form "step.result".

libraries/botbuilder-dialogs/src/dialogSet.ts
```diff
@@ -23,33 +23,34 @@ import { DialogContext, DialogState } from './dialogContext';
...
- *     async function (dc, phone) {
- *         dc.activeDialog.state.profile.phone = phone;
+ *     async (dc, step) => {
+ *         dc.activeDialog.state.profile.phone = step.results;
  *
  *         // Save completed profile to user state
- *         const user = userState.get(context);
+ *         const user = await userState.get(context);
  *         user.profile = dc.activeDialog.state.profile;
+ *         await userState.set(context, user);
```
v-ducvo: For best practice, we should use state property accessor and avoid using the `set()` method because this method will delete the entire state before setting the new value to the state...it could have unexpected behavior if you have other data saved to the conversation or user state...

libraries/botbuilder-dialogs/src/dialogSet.ts
```diff
@@ -115,19 +116,18 @@ import { DialogContext, DialogState } from './dialogContext';
...
  *         if (isMessage) {
  *             const utterance = context.activity.text.trim().toLowerCase();
  *             if (utterance.startsWith('edit profile')) {
- *                 await dc.endAll().begin('fillProfile');
+ *                 await dc.cancelAll().begin('fillProfile');
  *                 return;
  *             } else if (utterance.startsWith('cancel')) {
  *                 if (dc.activeDialog) {
- *                     dc.endAll();
+ *                     dc.cancelAll();
```
v-ducvo: missing `await`

libraries/botbuilder-dialogs/src/dialogSet.ts
```diff
@@ -115,19 +116,18 @@ import { DialogContext, DialogState } from './dialogContext';
...
- *         const dc = dialogs.createContext(context, conversation);
+ *         const dc = dialogs.createContext(context);
  *
  *         // Check for any interruptions
  *         const isMessage = context.activity.type === ActivityType.Message;
  *         if (isMessage) {
  *             const utterance = context.activity.text.trim().toLowerCase();
  *             if (utterance.startsWith('edit profile')) {
- *                 await dc.endAll().begin('fillProfile');
+ *                 await dc.cancelAll().begin('fillProfile');
```
v-ducvo: missing `return` … if you don't return, continued execution may fail with errors. Generally, I found that `begin`, `repeat`, `prompt`, and `end` must be `return` or execution may fail in subsequent turns...

libraries/botbuilder-dialogs/src/dialogSet.ts
```diff
@@ -201,8 +201,7 @@ export class DialogSet {
      * that state.
      *
      * ```JavaScript
-     * const conversation = conversationState.get(context);
-     * const dc = dialogs.createContext(context, conversation);
+     * const dc = dialogs.createContext(context);
```
v-ducvo: should add `await`....exp: `const dc = await dialogs.createContext(context);`

libraries/botbuilder-dialogs/src/dialogSet.ts
```diff
@@ -74,11 +75,11 @@ import { DialogContext, DialogState } from './dialogContext';
  * server.post('/api/messages', (req, res) => {
  *     adapter.processActivity(req, res, async (context) => {
  *         // Get conversation state and create DialogContext object
- *         const conversation = conversationState.get(context);
- *         const dc = dialogs.createContext(context, conversation);
+ *         const dc = dialogs.createContext(context);
```
v-ducvo: missing `await`

v-ducvo:
I've just submitted some change requests...the comment entry in many cases is to only one incident....you should do a full scrub to fix everywhere else where the issue occurs.

libraries/botbuilder-dialogs/src/dialogSet.ts
benbrown: thanks!

libraries/botbuilder-dialogs/src/dialogSet.ts
benbrown: it shouldn't cause problems if you don't return here as long as you are checking the dc.responded and other status fields.  that said, I added returns here to simplify things.

libraries/botbuilder-dialogs/src/dialogSet.ts
benbrown: The variable name could be clearer, but I believe this _is_ showing the use of a state property accessor.  It just happens to be named userState.  Am I wrong?

libraries/botbuilder-dialogs/src/dialogSet.ts
benbrown: I updated this.

libraries/botbuilder-dialogs/src/dialogSet.ts
v-ducvo: This state is not using the accessor...it's using the botstate's get method which returns a cache state...This is generally a bad practice if you have multiple pieces of information stored in the userstate...like a "user info" bag and a "reservation" bag...when you call set in this statement, the user state is cleared then it saves whatever is in "user" to the state...not what a dev would expect (unless that is their intent, to clear all state and start with just the "user" data.

Using the state accessor:
```javascript
const conversationState = new ConversationState(storage);
const convoStateAccessor = conversationState.createProperty('convoState');
```

you can avoid this problem of deleting other data you stored because the accessor only grags you the property you want to work on...

`const convoState = await convoStateAccessor.get(context, {});`

you can then set values by individual properties or call the "set" method to replace whatever is in that property with your new one.

Hope that helps...

libraries/botbuilder-dialogs/src/dialogSet.ts
v-ducvo: Yes it will...try running this bot without the `return` for the begin method ;)

Also, you are chaining `dc.cancelAll().begin('fillProfile')`...this will give you an error:
`BotFrameworkAdapter.processActivity(): 500 ERROR - TypeError: dc.cancelAll(...).begin is not a function`

These are not promisable objects so you can't chain them any more ;(

libraries/botbuilder-dialogs/src/dialogSet.ts
benbrown: I have updated this example to include clear references to setting up memory storage, user state, and a property for the values we are using.

PR #413, opened 2018-09-11
fix more unittests

tomlm:
testAdapter didn't have the actual .startTest() method which returns a promise.


cleemullins:
@tomlm. The build is broken. :(

PR #415, opened 2018-09-11
implement privateConversationState

tomlm:


libraries/botbuilder-core/src/privateConversationState.ts
```diff
@@ -0,0 +1,71 @@
...
+import { BotState } from './botState';
+import { Storage } from './storage';
+import { TurnContext } from './turnContext';
+
+const NO_KEY: string = `PrivateConversationState: channelId and/or PrivateConversation missing from context.request.`;
+
+/**
+ * Reads and writes PrivateConversation state for your bot to storage.
+ *
+ * @remarks
```
cleemullins: This comment rocks! Please (!) cut/paste it over to the C# code as well!!

libraries/botbuilder-core/src/privateConversationState.ts
```diff
@@ -0,0 +1,71 @@
...
+     * Returns the storage key for the current PrivateConversation state.
+     * @param context Context for current turn of PrivateConversation with the user.
+     */
+    public getStorageKey(context: TurnContext): string | undefined {
+        const activity: Activity = context.activity;
+        const channelId: string = activity.channelId;
+        const conversationId: string = activity && activity.conversation && activity.conversation.id ? activity.conversation.id : undefined;
+        const userId: string = activity && activity.from && activity.from.id ? activity.from.id : undefined;
+
+        return channelId && conversationId && userId ? `${channelId}/conversations/${conversationId}/users/${userId}/${this.namespace}` : undefined;
```
cleemullins: Its scary that this returns undefined. That seems like it would corrupt a data store, as it would just start writing things to the null key. 

Consider throwing an exception rather than returning undefined.

libraries/botbuilder-core/tests/privateConversationState.test.js
```diff
@@ -0,0 +1,99 @@
...
+const { TurnContext, ActivityTypes, PrivateConversationState, MemoryStorage, TestAdapter } = require('../');
+
+const receivedMessage = { text: 'received', type: 'message', channelId: 'test', conversation: { id: 'convo' }, from: { id: 'user' }  };
+const missingChannelId = { text: 'received', type: 'message', conversation: { id: 'convo' }, from: { id: 'user' }  };
+const missingConversation = { text: 'received', type: 'message', channelId: 'test', from: { id: 'user' }  };
+const missingFrom = { text: 'received', type: 'message', channelId: 'test', conversation: { id: 'convo' },};
+const endOfConversation = { type: 'endOfConversation', channelId: 'test', conversation: { id: 'convo' } };
+
+describe(`PrivateConversationState`, function () {
+    this.timeout(5000);
```
cleemullins: Why such a long timeout? Seems like 500ms would be enough.

PR #417, opened 2018-09-12
Stevenic/combine context

Stevenic:


libraries/botbuilder-dialogs/src/prompts/activityPrompt.ts
```diff
@@ -45,22 +45,16 @@ export abstract class ActivityPrompt extends Dialog {
         const recognized: PromptRecognizerResult<Activity> = await this.onRecognize(dc.context, state.state, state.options);
 
         // Validate the return value
-        let end: boolean = false;
-        let endResult: any;
-        await this.validator(dc.context, {
+        let isValid = await this.validator({
```
benbrown: could be a const

libraries/botbuilder-dialogs/src/prompts/activityPrompt.ts
```diff
@@ -45,22 +45,16 @@ export abstract class ActivityPrompt extends Dialog {
         const recognized: PromptRecognizerResult<Activity> = await this.onRecognize(dc.context, state.state, state.options);
```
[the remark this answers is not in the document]
Stevenic: The ActivityPrompt **requires** you pass in a validator (its not optional) so we don't need the extra check. I had to double check that myself when going through and updating the code but good eye.

libraries/botbuilder-dialogs/src/prompts/activityPrompt.ts
Stevenic: totally... I'll update that

libraries/botbuilder-dialogs/src/prompts/activityPrompt.ts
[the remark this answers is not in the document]
Stevenic: I added a comment to better call that out

benbrown:
Looks good! I think the single StepContext will make lives easier!

libraries/botbuilder-dialogs/tests/attachmentPrompt.test.js
```diff
@@ -57,10 +57,9 @@ describe('AttachmentPrompt', function() {
 
         const dialogState = convoState.createProperty('dialogState');
         const dialogs = new DialogSet(dialogState);
-        dialogs.add(new AttachmentPrompt('prompt', (context, prompt) => {
-            assert(context);
+        dialogs.add(new AttachmentPrompt('prompt', async (prompt) => {
             assert(prompt);
-            prompt.end(prompt.recognized.value);
+            return prompt.recognized.succeeded;
```
stevengum: @v-ducvo, @kaiqb, here's a delta in how to use validators that needs to be documented.

Issue #390, opened 2018-08-30
The `BotConfiguration`  class in `botframework-config` serializes to JSON with the `internal` property visible.


justinwilaby:
This should be set to `enumerable: false` or specify the enumerable properties in a `toJSON()` override.

PR #423, opened 2018-09-13
Set BotConfiguration.internal as not enumerable

benbrown:
Fix for #390


benbrown:
@cleemullins added a test!

PR #424, opened 2018-09-13
remove middleware component from BotState, rewrite some tests

stevengum:


libraries/botbuilder-core/tests/botStateSet.test.js
```diff
@@ -30,7 +30,7 @@ class BotStateMock {
     }
 }
 
-describe(`BotStateSet`, function () {
+xdescribe(`BotStateSet`, function () {
```
tomlm: >xdescribe [](start = 0, length = 9)

xdescribe?  shouldn't this be describe()?

libraries/botbuilder-dialogs/tests/attachmentPrompt.test.js
```diff
@@ -21,11 +21,12 @@ describe('AttachmentPrompt', function() {
                 const attachment = results.result[0];
                 await turnContext.sendActivity(`${attachment.content}`);
             }
+            await convoState.saveChanges(turnContext);
         });
 
         // Create new ConversationState with MemoryStorage and register the state as middleware.
```
tomlm: >d register the state as middleware. [](start = 61, length = 35)

comment

libraries/botbuilder-dialogs/tests/attachmentPrompt.test.js
```diff
@@ -21,11 +21,12 @@ describe('AttachmentPrompt', function() {
...
         const convoState = new ConversationState(new MemoryStorage());
-        adapter.use(convoState);
+        // adapter.use(convoState);
```
tomlm: delete

libraries/botbuilder-core/src/botState.ts
```diff
@@ -50,7 +49,7 @@ export interface CachedBotState {
  * });
  * ```
  */
-export class BotState implements PropertyManager, Middleware {
+export class BotState implements PropertyManager {
```
tomlm: >BotState [](start = 13, length = 8)

do we have a unit test which tests autoSaveStateMiddleware?

PR #425, opened 2018-09-13
add LuisService.getEndpoint() and unit test

tomlm:


libraries/botframework-config/src/models/luisService.ts
```diff
@@ -16,6 +16,10 @@ export class LuisService extends ConnectedService implements ILuisService {
         super(source, serviceType || ServiceTypes.Luis);
     }
 
+    // get endpoint for the luis service
+    public getEndpoint() { 
+        return `https://${this.region}.api.cognitive.microsoft.com/luis/v2.0/`;
```
stevengum: `'/luis/v2.0/'` needs to be removed per [this line of code](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/luisRecognizer.ts#L111) which defaults to `'https://westus.api.cognitive.microsoft.com'`

benbrown: I concur, see the note here:
https://www.nuget.org/packages/Microsoft.Azure.CognitiveServices.Language.LUIS.Runtime/

>>>    1. Supported customizing service endpoints by assigning the endpoint string to LUISRuntimeClient.Endpoint. The endpoint string can be found on Azure Portal, it should contain only protocol and hostname, for example: https://westus.api.cognitive.microsoft.com/.

Issue #403, opened 2018-09-06
CosmosDb throwing "documentdb_1.ConnectionPolicy is not a constructor" error at compile time


v-ducvo:
**This issue was reported in an email. I'm posting the content of that email here to track the status of this issue.**

I’m running into this error and am not sure what the work around is. I am getting this at compile time.

In my bot, I have created a new CosmosDbStorage object like this:
```javascript
const storage = new CosmosDbStorage({
    serviceEndpoint: "https://lucascosmos.documents.azure.com:443/", 
    authKey: "xCmY4dZqZACCVUoExxAbfa38MwPjShqbm2TTqNhzKnqlTCLSc8mTio1TeCbEu2dCCg8VmNAohRARuUMPmmrRPg==", 
    databaseId: "Tasks",
    collectionId: "Items"
});
```

This failed with the error:
```javascript
c:\Users\v-ducvo\Documents\botdocs\samples\BotFramework-Samples\docs-samples\V4\JS\node_modules\botbuilder-azure\lib\cosmosDbStorage.js:32 
const policy = new documentdb_1.ConnectionPolicy(); 
^ 
TypeError: documentdb_1.ConnectionPolicy is not a constructor 
at new CosmosDbStorage (c:\Users\v-ducvo\Documents\botdocs\samples\BotFramework-Samples\docs-samples\V4\JS\node_modules\botbuilder-azure\lib\cosmosDbStorage.js:32:24) 
at Object.<anonymous> (c:\Users\v-ducvo\Documents\botdocs\samples\BotFramework-Samples\docs-samples\V4\JS\manage-conversation-and-user-state\azure-cosmos-db.js:40:17) 
```

Looking at line 67 of the source file, I see that connectionPolicyConfigurator is an optional field:
```javascript
/**
     * Creates a new ConsmosDbStorage instance.
     *
     * @param settings Setting to configure the provider.
     * @param connectionPolicyConfigurator (Optional) An optional delegate that accepts a ConnectionPolicy for customizing policies. More information at http://azure.github.io/azure-documentdb-node/global.html#ConnectionPolicy
     */

    public constructor(settings: CosmosDbStorageSettings, connectionPolicyConfigurator: (policy: ConnectionPolicy) => void = null)
```

[...]

PR #431, opened 2018-09-13
Workaround issue #403 and re-enabled CosmosDbStorage Tests

JuanAr:
- Re-enabled `cosmosDBStorage.tests` to reproduce #403 issue
- Found workaround for `ConnectionPolicy` constructor in https://github.com/Azure/azure-cosmosdb-node/issues/194#issuecomment-315624414
```
var connectionPolicy = new documentBase.ConnectionPolicy();
```
- This workaround is related to https://github.com/Azure/azure-cosmosdb-node/issues/187 which is not yet implemented in `documentdb`
- Since this is a workaround, we fixed `documentdc` version to `1.14.5`, currently the last version for which this workaround works.
- This might change in the future since from what we've learnt, the `documentdb` had a breaking change at some point.


JuanAr:
Disable CosmosDbTests again.
There seems to be a config problem in the build server that prevent tests to reach the service.
If enabled and the CosmosDb Emulator is running, these tests will pass locally.

Issue #426, opened 2018-09-13
Improve error message for luisrecognizer.recognize 


vishwacsena:
Calling luisrecognize.recognize with a subscription key not assigned to the LUIS application throws the following - 

Ask: can we improve the error here to say something like invalid key or something to that effect? 

FWIW, testing this scenario in [API test console](https://westus.dev.cognitive.microsoft.com/docs/services/5819c76f40a6350ce09de1ac/operations/5819c77140a63516d81aee78/console) gives a more meaningful error 

"401 This application cannot be accessed with the current subscription"

(node:4540) UnhandledPromiseRejectionWarning: Error
warning.js:18
    at client.pipeline (c:\repos\BotBuilder-Samples\javascript_nodejs\50.contoso-cafe-bot\node_modules\azure-cognitiveservices-luis-runtime\lib\operations\prediction.js:179:19)
    at retryCallback (c:\repos\BotBuilder-Samples\javascript_nodejs\50.contoso-cafe-bot\node_modules\ms-rest\lib\filters\systemErrorRetryPolicyFilter.js:89:9)
    at retryCallback (c:\repos\BotBuilder-Samples\javascript_nodejs\50.contoso-cafe-bot\node_modules\ms-rest\lib\filters\exponentialRetryPolicyFilter.js:140:9)
    at c:\repos\BotBuilder-Samples\javascript_nodejs\50.contoso-cafe-bot\node_modules\ms-rest\lib\filters\rpRegistrationFilter.js:59:14
    at handleRedirect (c:\repos\BotBuilder-Samples\javascript_nodejs\50.contoso-cafe-bot\node_modules\ms-rest\lib\filters\redirectFilter.js:39:9)
    at c:\repos\BotBuilder-Samples\javascript_nodejs\50.contoso-cafe-bot\node_modules\ms-rest\lib\filters\formDataFilter.js:23:14
    at Request.defaultRequest [as _callback] (c:\repos\BotBuilder-Samples\javascript_nodejs\50.contoso-cafe-bot\node_modules\ms-rest\lib\requestPipeline.js:125:16)
    at Request.self.callback (c:\repos\BotBuilder-Samples\javascript_nodejs\50.contoso-cafe-bot\node_modules\ms-rest\node_modules\request\request.js:185:22)
    at emitTwo (events.js:126:13)
    at Request.emit (events.js:214:7)
[...]

PR #442, opened 2018-09-16
add useful error messages based on response codes for LuisRecognizer

stevengum:
for #426

### DIFF
--- a/libraries/botbuilder-ai/src/luisRecognizer.ts
+++ b/libraries/botbuilder-ai/src/luisRecognizer.ts
@@ -103,7 +103,8 @@ export class LuisRecognizer {
             includeInstanceData: true,
             log: true,
             spellCheck: false,
-            staging: false, ...options};
+            staging: false, ...options
+        };
         this.includeApiResults = !!includeApiResults;
 
         // Create client
@@ -154,7 +155,7 @@ export class LuisRecognizer {
                     timezoneOffset: this.options.timezoneOffset,
                     verbose: this.options.includeAllIntents,
                     log: this.options.log,
-                    customHeaders: { 'Ocp-Apim-Subscription-Key': this.application.endpointKey}
+                    customHeaders: { 'Ocp-Apim-Subscription-Key': this.application.endpointKey }
                 }
             )
                 .then((luisResult: LuisModels.LuisResult) => {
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -205,6 +210,33 @@ export class LuisRecognizer {
         });
     }
 
+    private prepareErrorMessage(error: Error): Error {
+        switch ((error as any).response.statusCode) {
+            case 401:
```
cleemullins: The text in these responses doesn't match the text found on the LUIS docs site:
https://docs.microsoft.com/en-us/azure/cognitive-services/luis/luis-reference-response-codes. 

Despite that I like your text better, I believe you should cut/paste the text from the docs and use that.

cleemullins: Missing 400. "request's parameters are incorrect meaning the required parameters are missing, malformed, or too large"

cleemullins: Note: If you said "I cut a PR on the docs site to update their text to the next text" that would be the best choice.

cleemullins:
@stevengum Minor changes requested. Should be quick.

libraries/botbuilder-ai/src/luisRecognizer.ts
stevengum: Due to the lack of knowledge around the docs' release cadence (iirc they like to gather a lot and release at once) and how there might be a diff for a while, I changed the 401 error message to more closely follow what was found in the docs' API code explanations. The new error message is this:

> Response 401: The key used is invalid, malformed, empty, or doesn't match the region.

I'm going to add @diberry to this conversation to see if we can update the docs in the future and then add those new explanations to the SDK.

stevengum: Added this, good catch! #Resolved

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -205,6 +210,35 @@ export class LuisRecognizer {
         });
     }
 
+    private prepareErrorMessage(error: Error): Error {
```
carlosscastro: could error be null / undefined?

stevengum: Technically where it's called it could never be `null` or `undefined` unless someone changed the catch  logic to change `error`.

PR #443, opened 2018-09-17
make BotState store plain objects not StoreItem, #297

stevengum:
for #297,
based off of https://github.com/Microsoft/botbuilder-dotnet/commit/148778a19e50e1efe049eb8dd69d6919fbd40b02

### DIFF
--- a/libraries/botbuilder-azure/src/blobStorage.ts
+++ b/libraries/botbuilder-azure/src/blobStorage.ts
@@ -6,7 +6,7 @@
  * Licensed under the MIT License.
  */
 import * as azure from 'azure-storage';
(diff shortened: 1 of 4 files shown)


libraries/botbuilder-azure/src/blobStorage.ts
```diff
@@ -6,7 +6,7 @@
  * Licensed under the MIT License.
  */
 import * as azure from 'azure-storage';
-import { Storage, StoreItem, StoreItems } from 'botbuilder';
+import { Storage, StoreItems } from 'botbuilder';
```
carlosscastro: Unit test using azure storage emulator would be nice for this codebase.

Issue #420, opened 2018-09-12
NumberPrompt returns NaN if a user responses with 0 or "zero"


benbrown:
When a user responds with 0 to a number prompt, prompt.recognized.succeeded is set to true, but prompt.recognized.value is NaN.

benbrown:
This appears to be a problem in the recognizer, the raw value returned by Recognizers.recognizeNumber is:

```
[ { start: 0,
    end: 0,
    resolution: { value: undefined },
    text: '0',
    typeName: 'number' } ]
```

cleemullins:
@JuanAr Hey Juan, any update on this?

johnataylor:
This issue does not reproduce in the C# implementation.

johnataylor:
I'm seeing exactly the same as Ben on the 1.0.1 version of the Node package.

johnataylor:
Incidentally both implementations, Node and C#, behave well and consistently with very large number strings. Its just zero in node that seems broken.

And note its is both '0' and the word 'zero' that produce the undefined value in the resolution.

JuanAr:
This issue was already addressed in Microsoft/Recognizers-Text#508 (merged). The next release of [recognizers-text-number](https://www.npmjs.com/package/@microsoft/recognizers-text-number) will include this fix.

JuanAr:
Latest version of https://www.npmjs.com/package/@microsoft/recognizers-text-number and https://www.npmjs.com/package/@microsoft/recognizers-text-suite includes a fix for this issue.

sgellock:
work around merged and is in 4.0.5 builds.  verified fixed

PR #450, opened 2018-09-18
Potential temporary patch for #420

benbrown:
This patch catches the invalid value from the number recognizer, which would clear our short term dependency on that library's fix being published


iMicknl:
Isnt this fixed in the [release](https://github.com/Microsoft/Recognizers-Text/releases/tag/javascript-v1.1.0) of this morning? :-)

benbrown:
Theoretically yes @iMicknl, but I'm having problems building against the 1.1.0 version of the number recognizer. It is throwing errors about bigNumber missing.

johnataylor:
I just verified the fix in a stand alone Node project. Works great for '0' and 'zero'.

**I don't think we want this work around.**

cleemullins:
@benbrown The build for this PR fails.

benbrown:
@cleemullins yes, but it fails because of the issue in the sub-package that we're trying to fix :O

Sounds like we're going to fix the root issue, so I'm going to close this.

benbrown:
This has been updated with a new test case to validate that we're getting 0s as expected.

PR #456, opened 2018-09-18
fix recognizer packages to 1.1.0 and add bignumber

johnataylor:


JuanAr:
@johnataylor I think as a workaround you can instead `npm install bignumber.js` in the root's `package.json`. That worker for me while testing this locally.

cleemullins:
@johnataylor I'm going to close out this PR as a dupe of #457.

PR #457, opened 2018-09-18
Update recognizer dependencies to latest version

JuanAr:
Updated recognizers dependencies to latest verison: 1.1.0
- @microsoft/recognizers-text-suite
- @microsoft/recognizers-text-choice
- @microsoft/recognizers-text-number
- @microsoft/recognizers-text-date-time

This changes were tested using an adapted version of samples\multiple-prompts-bot-es6.
Additionally, we tweaked the libraries to install bignumber.js which should be fixed in the recognizers as well. This PR can be merged independently of those recognizers fixes.

### DIFF
--- a/libraries/botbuilder-ai/package-lock.json
+++ b/libraries/botbuilder-ai/package-lock.json
@@ -5,91 +5,85 @@
 	"requires": true,
 	"dependencies": {
 		"@microsoft/recognizers-text": {
-			"version": "1.0.1",
-			"resolved": "https://registry.npmjs.org/@microsoft/recognizers-text/-/recognizers-text-1.0.1.tgz",
-			"integrity": "sha512-BZbI6YiLrWRd/jAYgSpLQ+Otl3WYMctefh7Myw6jqMdHhADaj5WpwOdf+yfJHZHOJWQh7lK2v50R6Y0g3TJ18w=="
+			"version": "1.1.2",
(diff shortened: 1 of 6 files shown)


johnataylor:
@JuanAr we're looking at the same issue.
seems the recognizers are missing a dependency. so we tried adding it. but travis stopped working. our latest attempt is to fix the dependency on 1.1.0 for recognizers and on bignumber 7.2.1

johnataylor:
so did you publish these last night?
who can fix the dependency within the recoignizer packages themselves?

JuanAr:
We spot that the number (hence suite) recognizers are missing bignumber.js. The dependency is still in the root package.json (that's why build works leveraging lerna)
I'm about to create a PR with the bignumber.js dependency in the correct package.json in Recognizers. Then I'll ask @tellarin to trigger a new package publish.
I think if new version is 1.1.1, npm will found the fix. Other than that, we'll have to update botbuilder-js dependencies to match the latest version.

JuanAr:
Just created PR Microsoft/Recognizers-Text#863 to fix the `bignumber.js` issue.

johnataylor:
In order to fix the build on travis for other checkins we have current fixed the dependency to recognizers 1.0.1. So when version 1.1.1 is out there we will still need to update the dependencies in botbuilder.

JuanAr:
Closing and favor https://github.com/Microsoft/botbuilder-js/pull/472

PR #464, opened 2018-09-19
[4.1] Merged MainDialog logic into ComponentDialog **DO NOT MERGE**

Stevenic:
Moved the behavior of the `MainDialog` class to the `ComponentDialog` class such that any component dialog can easily be the bots tops level main dialog.  This also adds a single `onRunTurn()` method that derived classes can override to add their interruption logic.

**This is being proposed as a change for version 4.1**


cleemullins:
Closing PR (but keeping branch) after discussion with @Stevenic

PR #472, opened 2018-09-20
Update recognizer dependencies to latest version ^1.1.2

JuanAr:
Updated recognizers dependencies to latest verison: 1.1.2
- @microsoft/recognizers-text-suite
- @microsoft/recognizers-text-choice
- @microsoft/recognizers-text-number
- @microsoft/recognizers-text-date-time

Replaces failing PR: https://github.com/Microsoft/botbuilder-js/pull/457 by updating only package.json file.


benbrown:
We should also remove the patch in NumberPrompt that checked for the result being NaN.

PR #477, opened 2018-09-21
Stevenic/botbuilder core docs

Stevenic:
Updated reference docs for `botbuilder-core`


libraries/botbuilder-core/src/botState.ts
```diff
@@ -88,7 +81,7 @@ export class BotState implements PropertyManager {
      * @param force (Optional) If `true` the cache will be bypassed and the state will always be read in directly from storage. Defaults to `false`.
      */
     public load(context: TurnContext, force: boolean = false): Promise<any> {
-        const cached: any = context.turnState.get(this.stateKey) as CachedBotState;
+        const cached: CachedBotState = context.turnState.get(this.stateKey);
```
sgellock: technically this is a code change.  are we ok with this?

libraries/botbuilder-core/src/botState.ts
```diff
@@ -119,7 +112,7 @@ export class BotState implements PropertyManager {
      * @param force (Optional) if `true` the state will always be written out regardless of its change state. Defaults to `false`.
      */
     public saveChanges(context: TurnContext, force: boolean = false): Promise<void> {
-        let cached: any = context.turnState.get(this.stateKey) as CachedBotState;
+        let cached: CachedBotState = context.turnState.get(this.stateKey);
```
sgellock: technically this is a code change.  are we ok with this?

libraries/botbuilder-core/src/botState.ts
```diff
@@ -165,12 +160,12 @@ export class BotState implements PropertyManager {
...
      *
      * ```JavaScript
-     * const state = await botState.get(context);
+     * const state = botState.get(context);
      * ```
      * @param context Context for current turn of conversation with the user.
      */
     public get(context: TurnContext): any|undefined {
-        const cached: any = context.turnState.get(this.stateKey) as CachedBotState;
+        const cached: CachedBotState = context.turnState.get(this.stateKey);
```
sgellock: code change

libraries/botbuilder-core/src/botState.ts
Stevenic: True but it's just a compile time change. The output code is the same

---
In reply to: [219688302](https://github.com/Microsoft/botbuilder-js/pull/477#discussion_r219688302) [](ancestors = 219688302)

libraries/botbuilder-core/src/botState.ts
Stevenic: same as above

---
In reply to: [219688303](https://github.com/Microsoft/botbuilder-js/pull/477#discussion_r219688303) [](ancestors = 219688303)

libraries/botbuilder-core/src/botState.ts
Stevenic: same

---
In reply to: [219688307](https://github.com/Microsoft/botbuilder-js/pull/477#discussion_r219688307) [](ancestors = 219688307)

libraries/botbuilder-core/src/botStatePropertyAccessor.ts
```diff
@@ -8,15 +8,79 @@
...
+
+    /**
+     * Reads a persisted property from its backing storage object.
+     * 
+     * @remarks
+     * The properties backing storage object SHOULD be loaded into memory on first access. 
+     * 
+     * If the property does not currently exist on the storage object and a `defaultValue` has been
+     * specified, a clone of the `defaultValue` SHOULD be copied to the storage object. If a
+     * `defaultValue` has not been specified then a value of `undefined` SHOULD be returned.
```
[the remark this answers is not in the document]
Stevenic: I was just trying to better call out that while most implementations will clone the default value it's not guaranteed.

PR #483, opened 2018-09-24
Language generation integration 

ahmedlhanafy:
This is the code for language generation resolver along with some tests

### DIFF
--- a/libraries/botbuilder-ai/src/lgResolver.ts
+++ b/libraries/botbuilder-ai/src/lgResolver.ts
@@ -6,26 +6,26 @@
  */
 
 import {
-  Activity,
-  CardAction,
-  CardFactory,
-  MediaCard,
-  OAuthCard,
-  ReceiptCard,
-  ReceiptItem,
-  SigninCard,
-  ThumbnailCard,
+	Activity,
+	CardAction,
+	CardFactory,
+	MediaCard,
+	OAuthCard,
+	ReceiptCard,
+	ReceiptItem,
+	SigninCard,
+	ThumbnailCard,
 } from 'botbuilder';
 import {
-  flatten,
-  isBoolean,
-  isDate,
-  isEmpty,
-  isInteger,
-  isNil,
-  isNumber,
-  isString,
-  replace,
+	flatten,
+	isBoolean,
+	isDate,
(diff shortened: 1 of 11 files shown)


libraries/botbuilder-ai/src/lgResolver.ts
```diff
@@ -0,0 +1,907 @@
...
+
+	/**
+	 * Extracts template references from the activity object, resolves them using the language generation service
+	 * and injects them back in the activity
+	 *
+	 * @param activity botbuilder activity
+	 * @param [entities] map object that contains slots accepted by the LG service
+	 * @returns
+	 */
+	public async resolve(activity: Activity, entities?: Map<string, PrimitiveType>): Promise<void> {
```
Stevenic: >Activity [](start = 32, length = 8)

This should be a Partial<Activity>

libraries/botbuilder-ai/src/lgResolver.ts
```diff
@@ -0,0 +1,907 @@
...
+ * @private
+ */
+export interface LGRequest {
+	scenario: string;
+	locale: string;
+	slots: { [key: string]: LGValue };
+	templateId: string;
+}
+
+class LGRequestBuilder {
```
Stevenic: >LGRequestBuilder [](start = 6, length = 16)

Should mark this @private

libraries/botbuilder-ai/src/lgResolver.ts
Stevenic: >Map<string, PrimitiveType> [](start = 53, length = 26)

Instead of Map I'd just make this a JSON object so { [name: string]: PrimitiveType; }

That will be easier for the caller versus having to new up and populate an actual Map

libraries/botbuilder-ai/src/lgResolver.ts
```diff
@@ -0,0 +1,907 @@
+/**
```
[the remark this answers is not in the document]
ahmedlhanafy: The file name reflects the older class name, I will update it to the new one. Regarding having everything in one file, I think that's the convention in the repo, all the other files follow this convention, so I though I should stick to it too.

PR #487, opened 2018-09-25
export PromptState interface so can be used outside

shmuelgutman:


shmuelgutman:
didn't see that it's private

PR #492, opened 2018-09-25
Use npmjs instead of myget in package-lock.json

JuanAr:
The only change in this PR is updating package-lock.json to use npmjs feed instead of myget


cleemullins:
@tomlm @Stevenic Thoughts on this?

stevengum:
Microsoft/BotBuilder-Samples#549 did something similar for the samples.

PR #499, opened 2018-09-27
[4.1] Added DC.consultDialog() method and plumbed through everywhere. **DO NOT MERGE**

Stevenic:
This change lets a parent dialog ask a child dialog, like a prompt or component dialog, if it understands the current activity before calling `DialogContext.continueDialog()`.  To use:

```JS
// Check for interruptions
const utterance = dc.context.activity.text;
if (/help/i.test(utterance)) {
    const consult = await dc.consultDialog();
    if (consult.status === DialogTurnStatus.empty || consult.score <= 0.5) {
        return await dc.beginDialog('helpDialog');
    }
}

// Continue active dialog
const result = await dc.continueDialog();
```

### DIFF
--- a/libraries/botbuilder-dialogs/src/componentDialog.ts
+++ b/libraries/botbuilder-dialogs/src/componentDialog.ts
@@ -6,7 +6,7 @@
  * Licensed under the MIT License.
  */
 import { TurnContext } from 'botbuilder-core';
-import { Dialog, DialogInstance, DialogReason, DialogTurnResult, DialogTurnStatus } from './dialog';
+import { Dialog, DialogInstance, DialogReason, DialogTurnResult, DialogTurnStatus, DialogConsultResult } from './dialog';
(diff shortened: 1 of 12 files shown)


benbrown:
So far so good! I have used this in a component dialog nested in a component dialog, and have been able to trigger successfully query child-components for interrupts and handlers that supersede parent handlers.

cleemullins:
@Stevenic @sgellock This is tagged "4.1" and "Do Not Merge". Should we take this, or not take this?

sgellock:
I don’t think we should take this for the 4.1 release. At the very least I think we want to write it up and let the community provide some feedback and consider the feedback for a possible inclusion in 4.2 or 4.3

benbrown:
I for one think this is a great addition, and create a much needed way to "test" against multiple dialogs desire to handle a given message.

cleemullins:
Closing PR (but keeping branch) after discussion with @Stevenic

PR #505, opened 2018-10-01
Fix activity timestamp string to Date.

enzocanoo:
- Convert `string` json props from `WebRequest.body` to `Date`.
- Change `cloneActivity` method to maintain the Date properties as Date.

----
These changes are needed to save the timestamp with `transcriptLogger`.


cleemullins:
@tomlm Can you take a quick look? This looks good to me.

tomlm:
@Stevenic can you look at this?  I don't understand why we need to do this?

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -775,6 +775,9 @@ function parseRequest(req: WebRequest): Promise<Activity> {
         function returnActivity(activity: Activity): void {
             if (typeof activity !== 'object') { throw new Error(`BotFrameworkAdapter.parseRequest(): invalid request body.`); }
             if (typeof activity.type !== 'string') { throw new Error(`BotFrameworkAdapter.parseRequest(): missing activity type.`); }
+            if (activity.timestamp && typeof activity.timestamp === 'string') { activity.timestamp = new Date(activity.timestamp); }
```
Stevenic: You don't need the first test. You can just do "if (typeof activity.timestamp === 'string')"

Issue #470, opened 2018-09-20
ShowTypingMiddleware is missing from the JavaScript


johnataylor:
### Versions
4.0.5

### Description
In the C# we have a ShowTypingMiddleware, this is missing from the JavaScript and should be added for parity. Note, it is important that the implementation follows that of the C# around sending the actual Activity because we don't want sending the Typing indicator Activity to flip the Responded flag.

cleemullins:
This exists in Steve's Toybox and should be moved into the primary repo.

PR #507, opened 2018-10-02
Add ShowTypingMiddleware

benbrown:
This adds a ShowTypingMiddleware to botbuilder-core, as well as some unit tests.

This is basically a port of the dotnet version, with some slight changes for Javascript.

Should address #470.

### DIFF
--- a/libraries/botbuilder-core/src/showTypingMiddleware.ts
+++ b/libraries/botbuilder-core/src/showTypingMiddleware.ts
@@ -11,12 +11,25 @@ import { TurnContext } from './turnContext';
 
  /**
   * Middleware that will send a typing indicator autmatically for each message.
+  * 
+  * @remarks
+  * When added, this middleware will send typing activities back to the user when a Message activity
+  * is receieved to let them know that the bot has received the message and is working on the response.
+  * You can specify a delay in milliseconds before the first typing activity is sent and then a frequency,
+  * also in milliseconds which determines how often another typing activity is sent. Typing activities
+  * will continue to be sent until your bot sends another message back to the user
   */
 export class ShowTypingMiddleware implements Middleware {
-        private readonly _delay: number;
-        private readonly _period: number;
-        private _interval: number;
+        private readonly delay: number;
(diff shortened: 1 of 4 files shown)


libraries/botbuilder-core/src/showTypingMiddleware.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        /** Implement middleware signature
+         * @param context {TurnContext} An incoming TurnContext object.
+         * @param next {function} The next delegate function.
+         */
+        public async onTurn(context: TurnContext, next: () => Promise<void>): Promise<void> {
+            if (context.activity.type === ActivityTypes.Message) {
+                 await this.sleep(this.delay);
+                 await this.sendTypingActivity(context);
```
Stevenic: This code will cause every incoming message to block for 500ms. You should just make startInterval() take the context and the amount of time to wait.  You don't want to "await" on anything before calling await next().

benbrown: Changed this as you suggested, and changed it from an interval to a recursive function using setTimeout.

libraries/botbuilder-core/src/showTypingMiddleware.ts
```diff
@@ -0,0 +1,106 @@
...
+
+        /** Implement middleware signature
+         * @param context {TurnContext} An incoming TurnContext object.
+         * @param next {function} The next delegate function.
+         */
+        public async onTurn(context: TurnContext, next: () => Promise<void>): Promise<void> {
+            if (context.activity.type === ActivityTypes.Message) {
+                // Set a property to track whether or not the turn is finished.
+                // When it flips to true, we won't send anymore typing indicators.
+                this.finished = false;
```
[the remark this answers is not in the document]
benbrown: Yeah whoops!
I've updated this to be a local variable which should resolve the rush condition/state issues.

benbrown:
Any more feedback on this, @stevenic?

Issue #522, opened 2018-10-09
Issue with luisRecognizer.js


Kperroch:
The luisRecognizer does not take into account the optional parameters in the query string.

the staging parameter for instance.

The correction can be : 
```
recognize(context) {
        const cached = context.turnState.get(this.cacheKey);
        if (!cached) {
            const utterance = context.activity.text || '';
            return this.luisClient.prediction.resolve(this.application.applicationId, utterance, {
                timezoneOffset: this.options.timezoneOffset,
                verbose: this.options.includeAllIntents,
                log: this.options.log,
		staging: this.options.staging,
                customHeaders: { 'Ocp-Apim-Subscription-Key': this.application.endpointKey }
            })
```

JasonSowers:
@Kperroch would you like to submit a pull request for this change so we can review it?

stevengum:
This was fixed with #530, thanks for reporting the bug.

PR #530, opened 2018-10-11
use object spread instead of explicit assigns in LuisRecognizer

stevengum:
This will address #522 in that all of the LuisPredictionOptions are now passed in to the underlying autorest-generated client.

### DIFF
--- a/libraries/botbuilder-ai/src/luisRecognizer.ts
+++ b/libraries/botbuilder-ai/src/luisRecognizer.ts
@@ -197,6 +197,7 @@ export class LuisRecognizer {
             return this.luisClient.prediction.resolve(
                 this.application.applicationId, utterance,
                 {
+                    verbose: this.options.includeAllIntents,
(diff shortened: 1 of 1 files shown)


libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -197,10 +197,8 @@ export class LuisRecognizer {
             return this.luisClient.prediction.resolve(
                 this.application.applicationId, utterance,
                 {
-                    timezoneOffset: this.options.timezoneOffset,
-                    verbose: this.options.includeAllIntents,
```
iMicknl: In the old version they pass `this.options.includeAllIntents` to `verbose`. In the object spread version this will be mapped to `includeAllIntents`. 

Is this still correct?

benbrown: Good catch!

stevengum: I'll go back in and make sure everything is mapped correctly. Thanks for catching this!

Issue #533, opened 2018-10-12
Builds are failing failing Auth Tests


stevengum:
This affects tests that do not use a valid MSA AppId and AppPassword, but instead use an old generated Auth Token. The problem is that whenever the keys get rotated out, the auth token becomes invalid and needs to be renewed. 

Instead of continuing with this auth token route, JS should take should take a page out the .NET repo and get a new token whenever the tests are run in Travis.

stevengum:
Update on this, for the [`AuthHeader` tests](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botframework-connector/tests/auth.test.js#L10-L65) the tokens obtained trigger [the emulator validation path](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botframework-connector/src/auth/jwtTokenValidation.ts#L60-L64). This might mean there might be an underlying issue in our auth stack.  

What this means in a shallow sense is that the [second test](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botframework-connector/tests/auth.test.js#L26-L36) fails because the auth stack successfully validates the token though the channelId is incorrect (neither 'webchat' nor 'emulator'):

            it('with incorrect ChannelId should not validate', function(done) {
                Connector.ChannelValidation.ToBotFromChannelTokenValidationParameters.ignoreExpiration = true;
[...]

PR #534, opened 2018-10-12
temporarily disable AuthHeader tests to stop build breaks, #533

stevengum:
First step in #533...


johnataylor:
approved... just waiting on build

Issue #531, opened 2018-10-11
Cannot createConversation with my own reference


mecampbellsoup:
## Versions
What package version of the SDK are you using. `4.x`
What nodejs version are you using `8.12.0`
What browser version are you using `N/A`
What os are you using `Mac OS`

## Describe the bug

I am not able to use the [`createConversation`](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder/src/botFrameworkAdapter.ts#L243) SDK function from the `BotbuilderFrameworkAdapter`. 

Here is the code I'm running:
```js
let contextHelper = new ContextHelper(turnContext)
let directMessageReference = contextHelper.buildDirectMessageReference(user.name)
await this.consoleLogAndSendActivity(turnContext, directMessageReference, 'directMessageReference')

// If we have already persisted the current conversation into this.conversationState
// then we continue that conversation; otherwise we create a new one.
if (storageKey.includes(conversationId)) {
    await this.adapter.createConversation(directMessageReference, async (proactiveTurnContext) => {
        await proactiveTurnContext.sendActivity(`(2) Hello (in private), new user!`)
    })
}
```

However, `this.adapter.createConversation` (where `this.adapter` is the BotFrameworkAdapter passed in from `app.js`) fails with `[onTurnError]: Error: Mandatory 'members' property is missing`.

## Expected behavior

I expect to be able to create a new conversation/DM a user for the first time programatically via the `createConversation` bot framework SDK.

mecampbellsoup:
I'm pretty sure the issue is that the `parameters` object [here](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder/src/botFrameworkAdapter.ts#L248) is not being constructed properly, i.e. it is missing required parameters for the REST API endpoint to which it sends the `POST` request.

## Steps to reproduce

### Add bot to an existing Skype chat

![image](https://user-images.githubusercontent.com/2043821/46883699-de616400-ce20-11e8-8852-ab9273515cde.png)

![image](https://user-images.githubusercontent.com/2043821/46883729-f0430700-ce20-11e8-932d-9a52bfac0d6e.png)

### Our bot then tries to invoke `BotFrameworkAdapter.createConversation()`

However the function encounters an error regarding missing a required `members` parameter.

![image](https://user-images.githubusercontent.com/2043821/46883771-0f419900-ce21-11e8-8d76-0fb22b69d489.png)

benbrown:
[...]

PR #535, opened 2018-10-12
Bugfix BotFrameworkAdapter.createConversation

mecampbellsoup:
The API endpoint for creating conversations requires a `Conversation` object parameter, and as such the SDK function requires additional parameters including `members` and potentially `topicName` and `activity` as well per:

https://docs.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-api-reference?view=azure-bot-service-4.0#create-conversation
https://docs.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-api-reference?view=azure-bot-service-4.0#conversation-object


mecampbellsoup:
Addresses #531.

@Stevenic @benbrown @tomlm @stevengum do you think you could take a quick look at this? Seems pretty straightforward but looking forward to your feedback - cheers!

PR #544, opened 2018-10-15
Cleaned up adapter

Stevenic:
- Updated code to use async/await
- Fixed status codes returned by processActivity().
- Added ability to pass ConversationParamters to createConversation().


libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -679,40 +629,32 @@ export class BotFrameworkAdapter extends BotAdapter {
...
-        } catch (err) {
-            return Promise.reject(err);
+    public async updateActivity(context: TurnContext, activity: Partial<Activity>): Promise<void> {
+        if (!activity.serviceUrl) { throw new Error(`BotFrameworkAdapter.updateActivity(): missing serviceUrl`); }
+        if (!activity.conversation || !activity.conversation.id) {
+            throw new Error(`BotFrameworkAdapter.updateActivity(): missing conversation or conversation.id`);
         }
+        if (!activity.id) { throw new Error(`BotFrameworkAdapter.updateActivity(): missing activity.id`); }
+        const client: ConnectorClient = this.createConnectorClient(activity.serviceUrl);
+        await client.conversations.updateActivity(
```
cleemullins: I don't understand why the updateActivity here wouldn't just take an Activity. That would allow all the valiations to be done in there, and it seems quite a bit cleaner. 

It's quite odd that you're passing in the id's off the activity AND the activity itself. Strongly consider refactoring to cleanup the implementation. 

There is no test coverage for a circumstance for where the activity.id and the activity don't match. Is that a valid scenario? If so, we'll have to update the C# code.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -238,34 +237,38 @@ export class BotFrameworkAdapter extends BotAdapter {
...
-
-                // Create context and run middleware
-                const context: TurnContext = this.createContext(request);
-
-                return this.runMiddleware(context, logic as any);
-            });
-        } catch (err) {
-            return Promise.reject(err);
+    public createConversation(reference: Partial<ConversationReference>, logic: (context: TurnContext) => Promise<void>): Promise<void>;
+    public createConversation(reference: Partial<ConversationReference>, params: ConversationParameters, logic: (context: TurnContext) => Promise<void>): Promise<void>;
```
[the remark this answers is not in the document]
cleemullins: This overload is new, but doesn't have a test.

johnataylor:
How about we make this separate PRs?

1. consistent application of async/await

1. alignment with C# adapter methods

1. http returns 200

libraries/botbuilder/src/botFrameworkAdapter.ts
Stevenic: This is from the swagger generated code. It takes an ID because the REST path generated has an id.

---
In reply to: [225354928](https://github.com/Microsoft/botbuilder-js/pull/544#discussion_r225354928) [](ancestors = 225354928)

PR #549, opened 2018-10-17
Add getEndpoint to ILuisService definition

### LINKED ISSUE #546: Does ILuisService need getEndpoint()?
## Describe the bug

The [interface declaration of ILuisService](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botframework-config/src/schema.ts#L183) does not include the getEndpoint() method that is included in the [LuisService implementation](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botframework-config/src/models/luisService.ts#L14).

If a developer casts to an ILuisService, this will cause issues with Typescript complaining that getEndpoint is not defined.  Is getEndpoint actually a required member of this method? Should we add it to the interface?

tomlm:
fixes #546


libraries/botframework-config/src/schema.ts
```diff
@@ -205,6 +205,11 @@ export interface ILuisService extends IConnectedService {
      * Region for luis.
      */
     region: string;
+
+    /**
+     * getEndpoint() returns the full region ednpoint for the luis service
```
benbrown:
```suggestion
     * getEndpoint() Returns the full region endpoint for the LUIS service.
```

PR #553, opened 2018-10-18
Fix #547: don't use select * in queries

### LINKED ISSUE #547: CosmosDB Storage is using "Select *" and should be explicit
The current implemention of CosmosDB uses the following query in two places:
```
{
        query: 'SELECT * FROM root r WHERE r.id=@id',
        parameters: [{ name: '@id', value: collectionId }]
};
```
This should be changed to use explicit fields, as using "Select *" is generally not considered best practice for writing SQL statements.

JuanAr:
- Changed `SELECT * FROM root r WHERE r.id=@id` into `SELECT r._self FROM root r WHERE r.id=@id`

### DIFF
--- a/libraries/botbuilder-azure/src/cosmosDbStorage.ts
+++ b/libraries/botbuilder-azure/src/cosmosDbStorage.ts
@@ -232,7 +232,7 @@ function getOrCreateDatabase(client: DocumentClient, databaseId: string): Promis
             value: string;
         }[];
     } = {
-        query: 'SELECT * FROM root r WHERE r.id = @id',
+        query: 'SELECT r._self FROM root r WHERE r.id = @id',
         parameters: [{ name: '@id', value: databaseId }]
     };
 
@@ -261,7 +261,7 @@ function getOrCreateCollection(client: DocumentClient, databaseLink: string, col
             value: string;
         }[];
     } = {
-        query: 'SELECT * FROM root r WHERE r.id=@id',
+        query: 'SELECT r._self FROM root r WHERE r.id=@id',
         parameters: [{ name: '@id', value: collectionId }]
     };
 


libraries/botbuilder-azure/src/cosmosDbStorage.ts
```diff
@@ -261,7 +261,7 @@ function getOrCreateCollection(client: DocumentClient, databaseLink: string, col
             value: string;
         }[];
     } = {
-        query: 'SELECT * FROM root r WHERE r.id=@id',
+        query: 'SELECT r._self FROM root r WHERE r.id=@id',
```
cleemullins: Line 235 in "getOrCreateDatabase" needs to be fixed as well.

libraries/botbuilder-azure/src/cosmosDbStorage.ts
```diff
@@ -261,7 +261,7 @@ function getOrCreateCollection(client: DocumentClient, databaseLink: string, col
             value: string;
         }[];
     } = {
```
cleemullins: The SQL Query here "Select r._self" is quite different from the [C# Query](https://github.com/Microsoft/botbuilder-dotnet/blob/34c957e998c03b37585dc993b48148e994134c9c/libraries/Microsoft.Bot.Builder.Azure/CosmosDbStorage.cs#L143). 

This seems like an area where the 2 platforms (and soon 3, with Java coming soon) should be *exactly* aligned. 

`
QueryText = $"SELECT c.id, c.realId, c.document, c._etag FROM c WHERE c.id in ({parameterSequence})"
`

libraries/botbuilder-azure/src/cosmosDbStorage.ts
JuanAr: This is done!

libraries/botbuilder-azure/src/cosmosDbStorage.ts
JuanAr: This query is used for something different than checking the database existence.
The [equivalent query in JS](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-azure/src/cosmosDbStorage.ts#L115) is exactly the same:
```
query: `SELECT c.id, c.realId, c.document, c._etag FROM c WHERE c.id in (${parameterSequence})`,
```

PR #556, opened 2018-10-19
Fixes and Updates to the CosmosDB Storage Provider

enzocanoo:
- Extracted Key Escaping into a dedicated class (with tests).
- Added numerous CTor exception tests.
- Fixed race condition around Initialization

Based on https://github.com/Microsoft/botbuilder-dotnet/pull/1014

### DIFF
--- a/libraries/botbuilder-azure/src/cosmosDbKeyEscape.ts
+++ b/libraries/botbuilder-azure/src/cosmosDbKeyEscape.ts
@@ -1,0 +1,41 @@
+/**
+ * @module botbuilder-azure
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export module CosmosDBKeyEscape {
+    const illegalKeys: string[] = ['\\', '?', '/', '#', '\t', '\n', '\r', '*'];
+    const illegalKeyCharacterReplacementMap: Map<string, string> =
+    illegalKeys.reduce<Map<string, string>>(
+        (map : Map<string, string>, c : string) => {
+            map.set(c, `*${c.charCodeAt(0).toString(16)}`);
+
+            return map;
+        },
+        new Map()
+    );
+
+    export function escapeKey(key: string): string {
+        if (!key) {
+            throw new Error('The \'key\' parameter is required.');
+        }
+
+        const keySplitted: string[] = key.split('');
+
+        const firstIllegalCharIndex: number = keySplitted.findIndex(((c : string): boolean => illegalKeys.some((i : string) => i === c)));
+
+        // If there are no illegal characters return immediately and avoid any further processing/allocations
+        if (firstIllegalCharIndex === -1) {
+            return key;
+        }
+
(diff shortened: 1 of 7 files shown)


libraries/botbuilder-azure/src/cosmosDbStorage.ts
```diff
@@ -213,8 +249,16 @@ export class CosmosDbStorage implements Storage {
      */
     private ensureCollectionExists(): Promise<string> {
         if (!this.collectionExists) {
-            this.collectionExists = getOrCreateDatabase(this.client, this.settings.databaseId)
-                .then((databaseLink: string) => getOrCreateCollection(this.client, databaseLink, this.settings.collectionId));
+            this.collectionExists = new Promise((resolve : Function, reject : Function) : void => {
+                _semaphore.take(() => {
```
johnataylor: Assuming this is needed. If not this should be harmless.

libraries/botbuilder-azure/src/cosmosDbKeyEscape.ts
```diff
@@ -0,0 +1,41 @@
...
+    illegalKeys.reduce<Map<string, string>>(
+        (map : Map<string, string>, c : string) => {
+            map.set(c, `*${c.charCodeAt(0).toString(16)}`);
+
+            return map;
+        },
+        new Map()
+    );
+
+    export function escapeKey(key: string): string {
```
Stevenic: This should be marked as @private

libraries/botbuilder-azure/src/cosmosDbKeyEscape.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * @module botbuilder-azure
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export module CosmosDBKeyEscape {
```
Stevenic: >CosmosDBKeyEscape [](start = 14, length = 17)

I'm not sure why this needs to be a module versus just a set of exported functions.  At any rate it should be marked @private

libraries/botbuilder-azure/src/index.ts
```diff
@@ -8,3 +8,4 @@
 export * from './cosmosDbStorage';
 export * from './blobStorage';
 export * from './azureBlobTranscriptStore';
+export * from './cosmosDbKeyEscape';
```
Stevenic: Is this something you want devs to be able to call?  If so ignore the @private comments and flush out the documentation

stevengum: Spoke with @cleemullins, this implementation mirrors the .NET SDK in 
 that [`CosmosDBKeyEscape.EscapeKey`](https://github.com/Microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.Azure/CosmosDBKeyEscape.cs#L31) is public so it can be covered by tests.

For the JS side of things we don't really need it to be public to still test it, but it makes sense from a C# POV.

PR #557, opened 2018-10-22
botframework-connector auth test cleanup

stevengum:
Cleanup around auth tests in the `botframework-connector` library.
1. Updates tests to be more ES6-like
2. Tests now use async/await syntax
3. When tests are expected to error, the errors are actually examined instead of simply seeing if an error exists

This PR does not:
- Address the issues in #533
- Fix the potential new issue in the `ChannelValidator` test: `validateIdentity should fail if no identity` (TypeError is thrown)


stevengum:
Closing and reopening to trigger Travis builds...

PR #566, opened 2018-10-24
use overload instead of optional parameter

mmkal:
using an optional parameter meant the return type was always `T | undefined`, even though when a default value is supplied, the value should never be `undefined`. Using an overload allows being more precise about the return type.


cleemullins:
@Stevenic Steve - can you take a quick look. This looks correct to me, and I'm a big fan of the overloading as it makes C# / Java alignment much easier.

PR #583, opened 2018-10-27
add us gov data centers to getEndpoint()

tomlm:
properly map us gov regions to us gov luis endpoint


cleemullins:
@tomlm This is approved, but we should sync on that datacenter name.

PR #594, opened 2018-10-31
Cleanup test utilities

stevengum:
Fixes N/A

## Description
Cleans up the `tools/` folder by removing a number of unused `azure` dependencies in the package.json and removes unused test utilities.

## Specific Changes
  - Remove unused `azure` dependencies
  - Delete unused test utility files


cleemullins:
@stevengum Do similar tools/dependencies exist in the C# repo?

stevengum:
@cleemullins, I don't think so; the history indicates that the `tools/` folder was added by the Southworks crew when they split out the botframework-connector and botframework-schema libraries. The only reference to the `tools/` contents is in the  tests for the connector ([`conector.test.js`](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botframework-connector/tests/connector.test.js)).

PR #599, opened 2018-11-01
Use language generation production environment 

ahmedlhanafy:
- Use language generation production environment 
- Throw an exception in the middleware if the language resolution fails


ahmedlhanafy:
@vishwacsena This fixes the issue we discussed yesterday. The code now resolves using production environment.

PR #631, opened 2018-11-13
Correct NO_KEY messages in State classes for issue #610

### LINKED ISSUE #610: NO_KEY error messages incorrectly refer to "context.request"
## Versions
All stable versions.

## Describe the bug
Error messages in the `State` classes incorrectly reference `"context.request"`.

## To Reproduce
1. Go to UserState, ConversationState or PrivateConversationState.
2. Examine the `NO_KEY` value and see references to "context.request". For example:
```ts
const NO_KEY: string = `UserState: channelId and/or conversation missing from context.request.`;
```

## Expected behavior
Should refer to `context.activity` instead.

[bug]

v-kydela:
Fixes #610 

## Description

I have corrected the faulty `NO_KEY` messages in three files to more accurately represent the error. While the issue originally called for a simple change from `context.request` to `context.activity`, I took it upon myself to reword the messages based on the conditions I've discovered will trigger the error.

## Specific Changes

The new message `overridden getStorageKey method did not return a key` can be found in:

  - userState.ts
  - conversationState.ts
  - privateConversationState.ts

## Testing

The `NO_KEY` error messages can only be triggered if the `getStorageKey` methods are overridden because the current `getStorageKey` methods cannot return anything that evaluates to false. In order to test the messages you must override the `getStorageKey` methods and have them return `undefined` or an empty string.

### DIFF
--- a/libraries/botbuilder-core/src/conversationState.ts
+++ b/libraries/botbuilder-core/src/conversationState.ts
@@ -10,7 +10,7 @@ import { BotState } from './botState';
 import { Storage } from './storage';
 import { TurnContext } from './turnContext';
 
-const NO_KEY: string = `ConversationState: channelId and/or conversation missing from context.request.`;
+const NO_KEY: string = `ConversationState: overridden getStorageKey method did not return a key.`;
 
 /**
  * Reads and writes conversation state for your bot to storage.
--- a/libraries/botbuilder-core/src/privateConversationState.ts
+++ b/libraries/botbuilder-core/src/privateConversationState.ts
@@ -10,7 +10,7 @@ import { BotState } from './botState';
 import { Storage } from './storage';
 import { TurnContext } from './turnContext';
 
-const NO_KEY: string = `PrivateConversationState: channelId and/or PrivateConversation missing from context.request.`;
+const NO_KEY: string = `PrivateConversationState: overridden getStorageKey method did not return a key.`;
 
 /**
  * Reads and writes PrivateConversation state for your bot to storage.
--- a/libraries/botbuilder-core/src/userState.ts
+++ b/libraries/botbuilder-core/src/userState.ts
@@ -10,7 +10,7 @@ import { BotState } from './botState';
 import { Storage } from './storage';
 import { TurnContext } from './turnContext';
 
-const NO_KEY: string = `UserState: channelId and/or conversation missing from context.request.`;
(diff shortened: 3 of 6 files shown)


cleemullins:
This change is causing tests to fail. Please update the tests as part of this so the build passes. 

```
  1) ConversationState
       should throw NO_KEY error if getStorageKey() returns falsey value.:
      AssertionError [ERR_ASSERTION]: unexpected Error.message received: ConversationState: overridden getStorageKey method did not return a key.
      + expected - actual
      -false
      +true
      
      at Context.<anonymous> (libraries/botbuilder-core/tests/conversationState.test.js:83:13)
  2) PrivateConversationState
       should throw NO_KEY error if getStorageKey() returns falsey value.:
      AssertionError [ERR_ASSERTION]: unexpected Error.message received: PrivateConversationState: overridden getStorageKey method did not return a key.
      + expected - actual
      -false
      +true
      
      at Context.<anonymous> (libraries/botbuilder-core/tests/privateConversationState.test.js:95:13)
  3) UserState
       should throw NO_KEY error if getStorageKey() returns falsey value.:
      AssertionError [ERR_ASSERTION]: unexpected Error.message received: UserState: overridden getStorageKey method did not return a key.
      + expected - actual
      -false
      +true
      
      at Context.<anonymous> (libraries/botbuilder-core/tests/userState.test.js:71:13)
```

v-kydela:
@cleemullins - I see now that the tests are failing because the old error messages have been hard-coded into the tests. To avoid this sort of hard-coding which could cause the tests to fail in the future, would it be acceptable for the tests to just check if any error is thrown rather than an error with that specific message? Or should I just hard-code the new message? A third option may be to throw a specific type of Error and then check for that. Please advise

cleemullins:
@v-kydela For now, I would just hard code the message. That's probably the quickest overall solution, and has the upside as being just as right (or wrong) as the previous solution.

stevengum:
@v-kydela Specific error types in JS are not common to my knowledge (or at the very least not something we’ve done in the JS SDK). 

Checking the error message ensures that the behavior we expect is occurring. The expected behavior is code that we wrote into the SDK. If (for example) a TypeError or RangeError is thrown when we expect another error, then that is a bug.

v-kydela:
Thank you for the advice, @cleemullins and @stevengum. I've submitted the changes

PR #636, opened 2018-11-14
Stevenic/clear state fix

### LINKED ISSUE #473: DCR: BotState.clear() only clears if load() has been called
## Versions
version 4.0.5 of both Node and C# packages

## Describe the bug
There's a couple of issues here around the `BotState.clear()`:

1. On both the C# and JS versions the current implementation assumes that you've already called `load()` because it only clears the cached copy of the backing state object. If you haven't called load() then the "on-disk" version of the state will remain unchanged.
2. The C# version of the method is asynchronous and the JS version isn't. They should both be async.
3. The C# version of the method is named `ClearStateAsync()` and the JS method is named `clear()`. These should be more in-sync name wise.

## Expected behavior
For the naming conflict I believe the JS method should be named `clear()` and the C# version should be named `ClearAsync()`. My rational for dropping "State" from the C# version is that the other CRUD methods are named `LoadAsync()` and `SaveChangesAsync()` without any mention of "State".

For the implementation logic my P0 is to get the signature of the methods in-sync and we can debate the implementation in future version. So to at least work around the issue I'm going to propose an implementation that looks like this:

```TS
    public clear(context: TurnContext): Promise<void> {
        return Promise.resolve(this.load(context)).then(() => {
             // We leave the change hash un-touched which will force the cleared state changes to get persisted.
            const cached: CachedBotState = context.turnState.get(this.stateKey);
            cached.state = {};
            context.turnState.set(this.stateKey, cached);
        });
    }
```

[bug]

Stevenic:
Fixes #473 

## Description
Fixes the issue where `BotState.load()` needed to be called before `BotState.clear()`.

## Specific Changes
- Updated `clear()` to ALLWAYS set the cache to have an empty object and invalid hash.
- Made `clear()` async to match C#.

  - 
  -
  -


libraries/botbuilder-core/src/botState.ts
```diff
@@ -144,13 +144,11 @@ export class BotState implements PropertyManager {
...
-        // We leave the change hash un-touched which will force the cleared state changes to get persisted.
-        const cached: any = context.turnState.get(this.stateKey) as CachedBotState;
-        if (cached) {
-            cached.state = {};
-            context.turnState.set(this.stateKey, cached);
-        }
+    public clear(context: TurnContext): Promise<void> {
+        // Just overwrite cached value with a new object and empty hash. The empty hash will force the
+        // changes to be saved. 
+        context.turnState.set(this.stateKey, { state: {}, hash: '' });
```
drub0y: This looks good to me for JS, but the C# version is not doing the same thing. See comments there to make sure they align.

cleemullins:
@Stevenic Build is failing. If you have a moment, please take a look.

cleemullins:
@Stevenic Build remains failing. Branch is out of date. 

When you have a moment, please clean this up.

PR #638, opened 2018-11-15
Updates to validate emulator JWT tokens from US Gov

Jeffders:
## Description
When the emulator is configured (via the channelService property in the .bot file) to use the US Government Azure AAD endpoint, this causes a different set of issuers and a different open id metadata URL to be needed to validate the emulator's JWT token.

## Testing
Because this involves testing JWT tokens which I cannot check-in, I validated with:
1. Emulator configured to use Gov talking connecting to a local JS bot configured to use Gov running locally 
2. Emulator configured to use Gov talking connecting to an Azure hosted JS bot configured to use Gov running locally 
3. DirectLine talking to the same bot (via US Gov)


libraries/botframework-connector/src/auth/emulatorValidation.ts
```diff
@@ -95,12 +99,16 @@ export module EmulatorValidation {
     export async function authenticateEmulatorToken(
         authHeader: string,
         credentials: ICredentialProvider,
+        channelService: string,
```
cleemullins: Same comment as C# - please update the method comments with the new `@param`.

Jeffders: Done.

Issue #551, opened 2018-10-18
qnaMaker want to add strictFilters for generateAnswer


NoDirective:
**Is your feature request related to a problem? Please describe.**
when i use qnaMaker  by generateAnswer method to get my answer,
sometimes it return many answer depending on  the question.
I want to add  strictFilters  parameter to filter my answer , but there is no options 
for that.

**Describe the solution you'd like**
if possible , please add Optional parameters for generateAnswer  that could send to the body 
of request .
even if qna api adds  more parameters , it could work  well.

PR #649, opened 2018-11-20
added QnA Maker Trace onto generateAnswer(). generateAnswer() paramet…

### LINKED ISSUE #637: qnamaker trace activity is not sent on generateAnswer calls off the qna recognizer
## Describe the bug
Trace activity should be emitted by the SDK (so emulator can render appropriate inspectors) when the generateAnswer method is called. At the moment the JS SDK only sends trace activity on answer() call. .

## To Reproduce
1. Get set up with 51.cafe-bot sample for nodejs
2. Run the bot (with the QnA and LUIS services configured)
3. Connect to the bot using the V4 Emulator
4. Send 'what are your locations'
5. The log section in emulator does not show any QnA trace activity

## Expected behavior
SDK generates trace activity and emulator can show it in logs and when i click on the QnA trace activity, emulator renders the qna inspector.

[bug]

Zerryth:
Fixes #637 

## Description
- generates QnA Maker Trace activity when getAnswers() is called
- deprecate generateAnswers() method to favor getAnswers(), which does not have trace
- can view log in emulator

## Specific Changes

- added `getAnswers()` method, which calls `emitTraceInfo()`
-  `@deprecated` added to `generateAnswer()`
- refactored `qnaMaker.ts` and `qnaMaker.test.js` to use async/await pattern
- added unit tests for updated method `getAnswers()`

### DIFF
--- a/libraries/botbuilder-ai/src/qnaMaker.ts
+++ b/libraries/botbuilder-ai/src/qnaMaker.ts
@@ -199,16 +199,10 @@ export class QnAMaker {
      * @param scoreThreshold (Optional) minimum answer score needed to be considered a match to questions. Defaults to a value of `0.001`.
      */
     public async getAnswers(context: TurnContext, top?: number, scoreThreshold?: number): Promise<QnAMakerResult[]> {
-        let question: string = '';
+        let question: string = context && context.activity ? context.activity.text : '';
+        const trimmedQuestion: string = question ? question.trim() : '';
 
-        if (context) {
-            if (context.activity) 
-                question = context.activity.text;
-        }
-
-        const q: string = question ? question.trim() : '';
-
-        if (q.length > 0) {
+        if (trimmedQuestion.length > 0) {
             const answers = await this.callService(this.endpoint, question, typeof top === 'number' ? top : 1);
             const minScore: number = typeof scoreThreshold === 'number' ? scoreThreshold : 0.001;
             const sortedQnaAnswers = answers.filter(
--- a/libraries/botbuilder-ai/src/qnaMaker.ts
+++ b/libraries/botbuilder-ai/src/qnaMaker.ts
@@ -173,7 +173,7 @@ export class QnAMaker {
     public answer(context: TurnContext): Promise<boolean> {
         const { top, scoreThreshold } = this.options;
 
-        return this.generateAnswer(context.activity.text, top, scoreThreshold).then((answers: QnAMakerResult[]) => {
+        return this.generateAnswer(context, top, scoreThreshold).then((answers: QnAMakerResult[]) => {
             return this.emitTraceInfo(context, answers).then(() => {
                 if (answers.length > 0) {
                     return context.sendActivity({ text: answers[0].answer, type: 'message' }).then(() => true);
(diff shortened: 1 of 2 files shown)


Zerryth:
aye-aye! will do

libraries/botbuilder-ai/src/qnaMaker.ts
```diff
@@ -193,18 +193,25 @@ export class QnAMaker {
      * @param top (Optional) number of answers to return. Defaults to a value of `1`.
```
cleemullins: These @param comments are not in sync with the changed code below. Please make them match.

libraries/botbuilder-ai/src/qnaMaker.ts
```diff
@@ -193,18 +193,25 @@ export class QnAMaker {
...
      * @param scoreThreshold (Optional) minimum answer score needed to be considered a match to questions. Defaults to a value of `0.001`.
      */
-    public generateAnswer(question: string|undefined, top?: number, scoreThreshold?: number): Promise<QnAMakerResult[]> {
+    public generateAnswer(context: TurnContext, top?: number, scoreThreshold?: number): Promise<QnAMakerResult[]> {
+        const question = context.activity.text;
```
cleemullins: I don't think we can make this change, as it's a compatibility break in code that people are likely using. 

The answer is probably: 
1) Update the existing method as best we can. 
2) Introduce a new method with the new signature and have it work correctly.
3) Put a @depricated decerator on the old method. [https://stackoverflow.com/questions/19412660/how-should-i-mark-a-method-as-obsolete-in-js](https://stackoverflow.com/questions/19412660/how-should-i-mark-a-method-as-obsolete-in-js)

stevengum: D'oh! Yes this would be a breaking change. After talking with @Zerryth, she will instead implement the [QnAMaker.GetAnswersAsync()](https://github.com/Microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.AI.QnA/QnAMaker.cs#L108) from the dotnet SDK  in JS and mark QnAMaker.generateAnswer() as deprecated.

Zerryth: Ok. 

@stevengum found that in the C# there's a method `GetAnswersAsync()` that does not exist in the TS version

I'll make a new method `getAnswers()` in TS, deprecate `generateAnswers()`, and write tests for `getAnswers()` method

It `generateAnswer()` will be left as it originally, since it can't `emitTraceInfo()` without accepting the breaking change `TurnContext` parameter type

cleemullins:
@Zerryth LGTM. I've signed off. 

@stevengum Steven?

Zerryth:
@stevengum see latest commit for your requested changes:
* Null checked for context.activity in `getAnswers()`. 
* Edited documentation on `getAnswers()` to include mention of Trace added and modified deprecated message in `generateAnswers()`
* simplified return statements in `getAnswers()` and `generateAnswer()`
* added missing period in comment
* restyled `if`/`else` statements to MS BF's convention for TS

Zerryth:
Ok, pushed changes to clean up nested if statements in `getAnswers()`.

However now Strict Task Master @stevengum wants me to write tests for the `.sort()` in `getAnswers()` and `generateAnswers()`, so ball still in my court pending those changes :)

stevengum:
Right now we're trying to get Microsoft/botbuilder-dotnet#1099 merged in which addresses this issue in Microsoft/botbuilder#5143: *add new GetAnswersAsync method with QnAMakerOptions param*.

@Zerryth, we'll need to get those changes in your PR to avoid any breaking changes. Then your PR will also address #551.

Zerryth:
(Sorry @stevengum -- made new PR 'cause silliness involving reimaged computer and not putting new code into clone of this branch)

See fixes in PR https://github.com/Microsoft/botbuilder-js/pull/683

PR #651, opened 2018-11-20
cherry-pick: add overloads to BotStatePropertAccessor.get()

stevengum:
Fixes bug introduced in #566 for 4.1.5 branch.

(See original PR #647 against master for more information)

### DIFF
--- a/libraries/botbuilder-core/src/botStatePropertyAccessor.ts
+++ b/libraries/botbuilder-core/src/botStatePropertyAccessor.ts
@@ -91,7 +91,9 @@ export class BotStatePropertyAccessor<T = any> implements StatePropertyAccessor<
         }
     }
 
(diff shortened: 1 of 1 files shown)


cleemullins:
@stevengum & I spoke in person, and we are not going to cherrypick this to the 4.1.5. The reason being is that we don't have plans to release a 4.1.6 build, and should such a build be required it would be due to an emergent bug. Adding this PR into that would create un-necessary churn. 

This PR is already merged into master, and will be picked up in the 4.2 release. Anyone needing this now can pull [4.2.0-preview off of the daily build feed](https://botbuilder.myget.org/gallery/botbuilder-v4-js-daily).

sgellock:
Not taking this now causes customers using TypeScript configured to use strict checking to fail.

PR #654, opened 2018-11-21
fix confirm prompt when invalid locale used with number selection

### LINKED ISSUE #648: ConfirmPrompt can throw an exception processing activity with locale string empty.
## Versions
botbuilder-dialogs
4.1.5

## Describe the bug
If the inbound activity locale is string empty the code in the recognize phase of the prompt appears to have an exception if the text of the utterance is not recognized. It attempts to look up the locale in its choice table. The logic should be the same as the prompt phase.

This has recently been fixed in the C# implementation in addition to an internal refactoring that consolidated the two lookup tables into a single table.

johnataylor:
Fixes #648 

## Description
Confirm prompt had exception when invalid locale used with number selection.

### DIFF
--- a/libraries/botbuilder-dialogs/src/prompts/confirmPrompt.ts
+++ b/libraries/botbuilder-dialogs/src/prompts/confirmPrompt.ts
@@ -83,17 +83,13 @@ export class ConfirmPrompt extends Prompt<boolean> {
     }
 
     protected async onPrompt(context: TurnContext, state: any, options: PromptOptions, isRetry: boolean): Promise<void> {
-        // Determine locale
-        let locale: string = context.activity.locale || this.defaultLocale;
-        if (!locale || !ConfirmPrompt.defaultChoiceOptions.hasOwnProperty(locale)) {
-            locale = 'en-us';
-        }
 
         // Format prompt to send
         let prompt: Partial<Activity>;
         const channelId: string = context.activity.channelId;
-        const choiceOptions: ChoiceFactoryOptions = this.choiceOptions || ConfirmPrompt.defaultChoiceOptions[locale];
-        const choices: any[] = this.confirmChoices || ConfirmPrompt.defaultConfirmChoices[locale];
+        const culture = this.DetermineCulture(context.activity);
+        const choiceOptions: ChoiceFactoryOptions = this.choiceOptions || ConfirmPrompt.defaultChoiceOptions[culture];
+        const choices: any[] = this.confirmChoices || ConfirmPrompt.defaultConfirmChoices[culture];
         if (isRetry && options.retryPrompt) {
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-dialogs/src/prompts/confirmPrompt.ts
```diff
@@ -130,4 +126,12 @@ export class ConfirmPrompt extends Prompt<boolean> {
 
         return result;
     }
+
+    private DetermineCulture(activity: Activity) : string {
```
stevengum: The method name should be camel cased, and there is an extra space after `)`

libraries/botbuilder-dialogs/src/prompts/confirmPrompt.ts
```diff
@@ -83,17 +83,13 @@ export class ConfirmPrompt extends Prompt<boolean> {
...
-        if (!locale || !ConfirmPrompt.defaultChoiceOptions.hasOwnProperty(locale)) {
-            locale = 'en-us';
-        }
 
         // Format prompt to send
         let prompt: Partial<Activity>;
         const channelId: string = context.activity.channelId;
-        const choiceOptions: ChoiceFactoryOptions = this.choiceOptions || ConfirmPrompt.defaultChoiceOptions[locale];
-        const choices: any[] = this.confirmChoices || ConfirmPrompt.defaultConfirmChoices[locale];
+        const culture = this.DetermineCulture(context.activity);
```
stevengum: This should be typed, `const culture: string` etc.

libraries/botbuilder-dialogs/src/prompts/confirmPrompt.ts
```diff
@@ -108,17 +104,17 @@ export class ConfirmPrompt extends Prompt<boolean> {
         const result: PromptRecognizerResult<boolean> = { succeeded: false };
         const activity: Activity = context.activity;
         const utterance: string = activity.text;
-        const locale: string = activity.locale || this.defaultLocale || 'en-us';
-        const results: any = Recognizers.recognizeBoolean(utterance, locale);
+        const culture = this.DetermineCulture(context.activity);
```
stevengum: `culture` should be typed

libraries/botbuilder-dialogs/src/prompts/confirmPrompt.ts
stevengum: Could this method be used elsewhere (e.g. ChoicePrompt), if the object we're examining is passed in as an argument?

```ts
public determineCulture(activity: Activity, defaultChoiceOptions: { [culture: string]: any }): any {
```

Or is this too sloppy?

johnataylor: easier to add as private and make public later...

PR #656, opened 2018-11-21
Lock internal dependencies to patch version

### LINKED ISSUE #652: Bot builder dependencies to other builder packages should use ~ versioning instead of ^
## Versions
All 4.x.x releases suffer from this issue

## Describe the bug
Bot builder dependencies in packages.json are currently `^4.0.0` when importing other bot builder package dependencies, e.g. [botbuilder-core](https://github.com/Microsoft/botbuilder-js/blob/66dd368e35a64302563afa66d17134e7c833853f/libraries/botbuilder/package.json#L26). This causes having a hard dependency on exact versions of botbuilder, e.g. 4.0.8, to bring in latest version of botbuilder-core, e.g. 4.1.5, that might have some breaking changes.  

## To Reproduce
Have a dependency on botbuilder: "4.0.8" in your node project and see that botbuilder-core version will be "4.1.5" installed as dependency.

benbrown:
Fixes #652 

## Description

Update all the package files to change dependency on other Botbuilder libraries from any minor release (^4.0.0) to the specific minor release (~4.1.x)

This means a user installing version 4.1.7 will get all 4.1 updates, but will not automatically migrate to 4.2 which could break things.

### DIFF
--- a/libraries/botbuilder-ai/package.json
+++ b/libraries/botbuilder-ai/package.json
@@ -2,7 +2,7 @@
   "name": "botbuilder-ai",
   "author": "Microsoft Corp.",
   "description": "Cognitive services extensions for Microsoft BotBuilder.",
-  "version": "4.0.0",
+  "version": "4.1.6",
   "license": "MIT",
   "keywords": [
     "botbuilder",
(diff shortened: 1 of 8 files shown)


benbrown:
@stevengum yes, we do update them as part of the build.

After consulting with @cleemullins I updated the versions on Github because I thought it better to have working internal consistency -- IE, how can we be pointing to ~4.1.6 if on Github the package file says the current version is 4.0.0?

IMHO we should keep the public record clean and consistent.

stevengum:
I'm all for keeping the public record clean and consistent as long as we stay consistent going forward 😄 

:shipit:!

benbrown:
@stevengum I have made an informal suggestion that we put together a shipping checklist that we run through every time we publish a new version that would include maintenance tasks like this!  I will make roll up my sleeves and _just do it_ soon.

PR #659, opened 2018-11-26
update model_fixes.js to set CardAction.channelData as an optional property

### LINKED ISSUE #639: CardAction requires channelData property
## Versions
4.1.5

v10.13.0

What os are you using
Windows

## Describe the bug
CardAction interface requires unexplained ```channelData``` property. [06.using-cards sample](https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/javascript_nodejs/06.using-cards/bot.js) creates CardActions without channelData, implying that it should not be required: 
```
            CardFactory.actions([
                {
                    type: 'openUrl',
                    title: 'Get started',
                    value: 'https://docs.microsoft.com/en-us/azure/bot-service/'
                }
```

## To Reproduce
Follow [06.using-cards sample](https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/javascript_nodejs/06.using-cards/bot.js) sample to create a HeroCard with a card action. Don't pass in ```channelData```.

## Expected behavior
Card should not require unnecessary properties. If there is a reason for this property, samples/docs should use it and explain it. 

## Screenshots
image

## Additional context
I'm developing in TypeScript, so I have to adhere to more rules. This is a bug nonetheless. 

[bug]

stevengum:
First fix for #639

## Description
Add `"channelData"` as an optional property for the CardAction interface.

Note: This doesn't fix #639 directly as the library needs to be regenerated. This PR sets it up so that then the library is properly regenerated (via the `generateClient` scripts), #639 will be fixed.

### DIFF
--- a/libraries/swagger/model_fixes.js
+++ b/libraries/swagger/model_fixes.js
@@ -11,7 +11,7 @@ var optionalModelProperties = {
     'SigninCard': ['text'],
     'Attachment': ['content', 'contentUrl', 'name', 'thumbnailUrl'],
     'CardImage': ['alt', 'tap'],
-    'CardAction': ['image', 'displayText', 'text'],
+    'CardAction': ['channelData', 'displayText', 'image', 'text'],
     'MediaUrl': ['profile']
 };
 


stevengum:
Pinging @sgellock as FYI

PR #669, opened 2018-12-02
add console.error for async TranscriptLogger.logActivity() impl in TranscriptLoggerMiddleware

### LINKED ISSUE #383: TranscriptLoggerMiddleware should await call to logger.logActivity(...)
TranscriptLoggerMiddleware member `logger` is a `TranscriptLogger` which defines `logActivity` as:

```TypeScript
logActivity(activity: Activity): void | Promise<void>;
```

Because `logActivity` might return a promise, TranscriptLoggerMiddleware should `await` calls to it, to ensure that promises are resolved correctly, and so that rejections are passed downstream.

Currently exceptions are swallowed up when a call to `this.logger.logActivity` returns a rejected Promise:

```TypeScript
try {
  let activity = this.transcript.shift();
  this.logger.logActivity(activity);
} catch (err) {
  console.error('TranscriptLoggerMiddleware logActivity failed', err);
}
```

Should be a simple fix to add `await` in front of `this.logger.logActivity(activity);`. Happy to submit a PR if that is preferred.

Call is at:
https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/transcriptLogger.ts#L90

stevengum:
Fixes #383

## Description
Previously, if asynchronous TranscriptLogger.logActivity() implementations threw an error in TranscriptLgogerMiddleware, an UnhandledPromiseRejectionWarning would occur. This PR changes TranscriptLoggerMiddleware to instead console.error the error thrown via an appended `catch()`.

## Specific Changes
  - Add logic to append catch if TranscriptLoggerMiddleware.logger.logActivity() returns a Promise 
  - Add tests for synchronous and asynchronous error handling in TranscriptLoggerMiddleware.onTurn() when calling this.logger.logActivity()


libraries/botbuilder-core/src/transcriptLogger.ts
```diff
@@ -92,7 +92,22 @@ export class TranscriptLoggerMiddleware implements Middleware {
...
+                // is no innate error handling in the method. This catch prevents
+                // UnhandledPromiseRejectionWarnings from being thrown and prints the error to the
+                // console.
+                if (logActivityResult instanceof Promise) {
+                    logActivityResult.catch(err => {
+                        // tslint:disable-next-line:no-console
+                        console.error('TranscriptLoggerMiddleware logActivity failed', err);
+                    });
+                }
             } catch (err) {
```
johnataylor: This is something of a nit.

If this err object is of type Error (which it _should_ be but isn't necessarily) then this line will do a toString which will write only a subset of the information being carried to the console.

Wouldn't it therefore be more correct to explicitly check if it was of type Error and if it was then print out more of the fields, otherwise just toString()? (Incidentally, if I remember correctly I don't think JSON.stringify will work on Error.)

In this particular case I'm not sure this matters to much because its just the console. But something to consider in general.

stevengum: How about this?

```ts
try {
  const activity: Activity = this.transcript.shift();
  // Comments comments comments...
  const logActivityResult = this.logger.logActivity(activity);
  // Comments comments comments...

  if (logActivityResult.catch(err => {
    this.transcriptLoggerErrorHandler(err);
  });
} catch (err) {
  this.transcriptLoggerErrorHandler(err);
}

private transcriptLoggerErrorHandler(err: Error|Any): void {
  // tslint:disable:no-console
  if (err instanceof Error) {
    console.error(`TranscriptLoggerMiddleware logActivity failed: "${ err.message }"`);
    console.error(err.stack);
  } else {
    console.error(`TranscriptLoggerMiddleware logActivity failed: "${ JSON.stringify(err) }"`);
  }
  // tslint:enable:no-console
}
```

PR #673, opened 2018-12-03
ActivityPrompt wasn't using retryPrompt if specified.

### LINKED ISSUE #495: ActivityPrompt doesn't use PromptOptions.retryPrompt
## Describe the bug
Behavior for `ActivityPrompt` differs from other prompts in that `PromptOptions.retryPrompt` is never used. The only time retryPrompt is handled or used is in [beginDialog()](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-dialogs/src/prompts/activityPrompt.ts#L38-L40) to construct the activity if `options.retryPrompt` is an object/activity.

[Lines 53-90](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-dialogs/src/prompts/activityPrompt.ts#L53-L90)
```ts
    public async continueDialog(dc: DialogContext): Promise<DialogTurnResult> {
        // Perform base recognition
        const state: any = dc.activeDialog.state as ActivityPromptState;
        const recognized: PromptRecognizerResult<Activity> = await this.onRecognize(dc.context, state.state, state.options);

        // Validate the return value
        // - Unlike the other prompts a validator is required for an ActivityPrompt so we don't
        //   need to check for its existence before calling it. 
        const isValid = await this.validator({
            context: dc.context,
            recognized: recognized,
            state: state.state,
            options: state.options
        });

        // Return recognized value or re-prompt
        if (isValid) {
            return await dc.endDialog(recognized.value);
        } else {
            return Dialog.EndOfTurn;
        }
    }

    public async resumeDialog(dc: DialogContext, reason: DialogReason, result?: any): Promise<DialogTurnResult> {
        // Prompts are typically leaf nodes on the stack but the dev is free to push other dialogs
        // on top of the stack which will result in the prompt receiving an unexpected call to
        // resumeDialog() when the pushed on dialog ends.
        // To avoid the prompt prematurely ending we need to implement this method and
        // simply re-prompt the user.
        await this.repromptDialog(dc.context, dc.activeDialog);

        return Dialog.EndOfTurn;
    }

    public async repromptDialog(context: TurnContext, instance: DialogInstance): Promise<void> {
        const state: ActivityPromptState = instance.state as ActivityPromptState;
        await this.onPrompt(context, state.state, state.options);
    }
```

## Expected behavior
`options.retryPrompt` should be used in `continueDialog()` and `repromptDialog()`. This behavior differs from the other `Prompts`.

[bug]

Stevenic:
Fixes #495 

## Description
ActivityPrompt wasn't using the retryPrompt option if it was passed in as part of the prompt options. This made it inconsistent with the other prompts.
 
## Specific Changes
ActivityPrompt now works more like the other prompts and will automatically re-prompt the user if the received activity is a "message" and was un-handled.
  

## Testing
Updated unit tests and added an additional test.

### DIFF
--- a/libraries/botbuilder-dialogs/src/prompts/activityPrompt.ts
+++ b/libraries/botbuilder-dialogs/src/prompts/activityPrompt.ts
@@ -5,7 +5,7 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-import { Activity, InputHints, TurnContext } from 'botbuilder-core';
+import { Activity, InputHints, TurnContext, ActivityTypes } from 'botbuilder-core';
 import { Dialog, DialogInstance, DialogReason, DialogTurnResult } from '../dialog';
 import { DialogContext } from '../dialogContext';
 import { PromptOptions, PromptRecognizerResult, PromptValidator } from './prompt';
@@ -45,7 +45,7 @@ export abstract class ActivityPrompt extends Dialog {
         state.state = {};
 
         // Send initial prompt
-        await this.onPrompt(dc.context, state.state, state.options);
+        await this.onPrompt(dc.context, state.state, state.options, false);
 
         return Dialog.EndOfTurn;
     }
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-dialogs/src/prompts/activityPrompt.ts
```diff
@@ -69,6 +69,10 @@ export abstract class ActivityPrompt extends Dialog {
         if (isValid) {
             return await dc.endDialog(recognized.value);
         } else {
+            if (dc.context.activity.type == ActivityTypes.Message && !dc.context.responded) {
```
benbrown:
```suggestion
            if (dc.context.activity.type === ActivityTypes.Message && !dc.context.responded) {
```

PR #701, opened 2018-12-16
Add more language around daily builds

sgellock:
the README makes it seem like you should use daily builds. this PR adds clarity as to when you would consider using daily bits.


libraries/botbuilder-dialogs/README.md
```diff
@@ -17,7 +17,8 @@ To add the latest version of this package to your bot:
 npm install --save botbuilder-dialogs
 ```
 
-#### Use the Daily Build
+#### How to Use Daily Builds
+If you want to play with the very latest versions of botbuilder, you can opt in to working with the daily builds.  This is not meant to be used in a production environment and is for advanced development.  Quality will vary and you should only use daily builds for exploritory purposes.
```
stevengum: "exploritory" should be "exploratory" but otherwise LGTM

PR #721, opened 2019-01-11
Added choices parameter to DialogContext.prompt method

### LINKED ISSUE #697: Missing typing for DialogContext.prompt()
## Versions
SDK version: 4.2.0
Node version: 10.5.0

## Describe the bug
This is the only signature available for the `DialogContext.prompt()` method:
```
prompt(dialogId: string, promptOrOptions: string | Partial<Activity> | PromptOptions): Promise<DialogTurnResult>
```

which does not contain the `choices` parameter.

## To Reproduce
1. Install `botbuilder-dialogs` and use it according to the docs/samples.
2. Trigger IntelliSense on a `DialogContext` object and observe the options for the `prompt()` method. The `choices` parameter is missing.
3. Confirm by opening `npm_modules/botbuilder-dialogs/lib/dialogContext.d.ts` and looking at the definition of `prompt()`. The `choices` parameter is missing.

## Expected behavior
I expected to see multiple choices for `prompt()` with different sets of parameters. Or at least an optional `choices` parameter.

[bug]

stevkan:
Fixes #697

## Description
Adds the choices parameter to the DialogContext.prompt method that was missing.

## Specific Changes
  - Added "(Optional)" to the promptOrOptions @param declaration
  - Changed choicesOrOptions @param declaration to "choices"
  - Added "choices?: (string|Choice)[]" parameter to prompt method


cleemullins:
@stevkan This change looks reasonable, but the code would need to be fixed in 2 places - both here an in C#. 

The C# code [is linked here](https://github.com/Microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.Dialogs/DialogContext.cs#L96).

libraries/botbuilder-dialogs/src/dialogContext.ts
```diff
@@ -148,10 +148,10 @@ export class DialogContext {
      * return await dc.prompt('confirmPrompt', `Are you sure you'd like to quit?`);
      * ```
      * @param dialogId ID of the prompt to start.
-     * @param promptOrOptions Initial prompt to send the user or a set of options to configure the prompt with..
-     * @param choicesOrOptions (Optional) array of choices associated with the prompt.
+     * @param promptOrOptions (Optional) Initial prompt to send the user or a set of options to configure the prompt with..
+     * @param choices (Optional) array of choices associated with the prompt.
      */
-    public async prompt(dialogId: string, promptOrOptions: string|Partial<Activity>|PromptOptions): Promise<DialogTurnResult>;
+    public async prompt(dialogId: string, promptOrOptions?: string|Partial<Activity>|PromptOptions, choices?: (string|Choice)[]): Promise<DialogTurnResult>;
```
stevengum: Why is `promptOrOptions` now optional? The function marks it as [required](https://github.com/Microsoft/botbuilder-js/pull/721/files#diff-957a2549a16757f7986d71b20d222995R157). 

Additionally, this results in one typing, where choices is optional. Can you keep the old typing and add an overload that includes choices?

> I expected to see multiple choices for prompt() with different sets of parameters. Or at least an optional choices parameter.
> - [navzam](https://github.com/Microsoft/botbuilder-js/issues/697#issue-390784742)

If you need an example, please see PR #647.

cleemullins:
once @stevengum has signed off, we can merge.

PR #722, opened 2019-01-12
Added ComponentDialog.run() method and unit tests **DO NOT MERGE**

Stevenic:
For [DCR 5235](https://github.com/Microsoft/BotBuilder/issues/5235)

## Description
This PR adds new `ComponentDialog.run()` and `ComponentDialog.onRunTurn()` methods which simplify the root dialog driver code for bots.

My implementation gives you two ways of persisting the components state.

1. You can pass a `StatePropertyAccessor` to the components constructor and the state will be automatically persisted to that property when run() is called.
2. You can pass in the backing state object when you call run() and your code will be responsible for persisting the state after run() is called.

## Testing
Added unit tests for new run() method.

### DIFF
--- a/libraries/botbuilder-dialogs/src/componentDialog.ts
+++ b/libraries/botbuilder-dialogs/src/componentDialog.ts
@@ -328,7 +328,7 @@ export class ComponentDialog<O extends object = {}> extends Dialog<O> {
 
         // Start the component if it wasn't already running
         if (result.status === DialogTurnStatus.empty) {
-            result = await dc.beginDialog(this.id, options);
+            result = await dc.beginDialog(this.id, options.dialogOptions);
         }
         return result;
 
--- a/libraries/botbuilder-dialogs/src/componentDialog.ts
+++ b/libraries/botbuilder-dialogs/src/componentDialog.ts
@@ -5,12 +5,28 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-import { TurnContext, BotTelemetryClient, NullTelemetryClient } from 'botbuilder-core';
+import { TurnContext, BotTelemetryClient, NullTelemetryClient, StatePropertyAccessor } from 'botbuilder-core';
 import { Dialog, DialogInstance, DialogReason, DialogTurnResult, DialogTurnStatus } from './dialog';
 import { DialogContext, DialogState } from './dialogContext';
 import { DialogSet } from './dialogSet';
 
 const PERSISTED_DIALOG_STATE: string = 'dialogs';
+
+/**
+ * Additional options that can be passed into the `ComponentDialog.run()` method.
+ */
+export interface ComponentDialogRunOptions {
+    /**
+     * (Optional) options to pass to the component when its first started.
+     */
+    dialogOptions?: object;
+
+    /**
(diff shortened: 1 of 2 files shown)


benbrown:
@Stevenic is this where you would imagine the onMessage and other on* methods would be implemented as well for the mvc work?

libraries/botbuilder-dialogs/src/componentDialog.ts
```diff
@@ -251,4 +302,36 @@ export class ComponentDialog<O extends object = {}> extends Dialog<O> {
         return this._telemetryClient;
     }
 
+    /**
+     * Called from within the bots logic handler to route a received activity to the appropriate
```
cleemullins: bot's

libraries/botbuilder-dialogs/src/componentDialog.ts
```diff
@@ -251,4 +302,36 @@ export class ComponentDialog<O extends object = {}> extends Dialog<O> {
...
+     * Called from within the bots logic handler to route a received activity to the appropriate 
+     * dialog.
+     * 
+     * @remarks
+     * This method lets any component be run in isolation without having to be added to another 
+     * `ComponentDialog` or `DialogSet`. 
+     * @param context Context for the current turn of conversation with the user.
+     * @param options (Optional) options that can be used to control the running of the component.
+     */
+    public async run(context: TurnContext, options?: ComponentDialogRunOptions): Promise<DialogTurnResult> {
```
cleemullins: This is goodness.

libraries/botbuilder-dialogs/src/componentDialog.ts
```diff
@@ -77,7 +93,19 @@ export class ComponentDialog<O extends object = {}> extends Dialog<O> {
      * This defaults to the ID of the first child dialog added using [addDialog()](#adddialog).
      */
     protected initialDialogId: string;
-    private dialogs: DialogSet = new DialogSet(null);
+    private readonly dialogs: DialogSet = new DialogSet(null);
```
tomlm: >readonly [](start = 12, length = 8)

Why make these readonly?  This means you can only set initialize from the constructor?  Makes it non-declaritive

libraries/botbuilder-dialogs/src/componentDialog.ts
```diff
@@ -221,6 +248,30 @@ export class ComponentDialog<O extends object = {}> extends Dialog<O> {
...
+     * and [onContinueDialog()](#oncontinuedialog) calls.
+     * 
+     * SHOULD be overridden by components that wish to perform custom interruption logic. The
+     * default implementation first calls `innerDC.continueDialog()` and if there was no active
+     * dialog it will then call `innerDC.beginDialog()` with the dialog assigned to
+     * [initialDialogId](#initialdialogid).
+     * @param innerDC Dialog context for the components internal `DialogSet`.
+     * @param options (Optional) options that were passed to the component by its parent.
+     */
+    protected async onRunTurn(innerDC: DialogContext, options?: O): Promise<DialogTurnResult> {
```
tomlm: >onRunTurn [](start = 20, length = 9)

Why not onTurn()?  That's what IBot has in the C# integration...

gabog:
The implementation is in samples for 4.4 and documentation matches that.

PR #732, opened 2019-01-29
Prevent Cortana from trying to use suggested actions in prompts

v-kydela:
Fixes  https://github.com/Microsoft/botbuilder-dotnet/issues/1173

## Description
I have modified choiceFactory.ts by removing `hasMessageFeed` and `supportsCardActions` from the `if` condition so that now a channel _must_ support suggested actions in order for the SDK to automatically select them as the list style. I have also deleted those local variables for cleanliness since they aren't being used anywhere else.


v-kydela:
@stevengum - I just pushed another commit

PR #735, opened 2019-01-30
Added ability to call dialogs within a parent component.

Stevenic:
Implements [this DCR](https://github.com/Microsoft/BotBuilder/issues/5242)

## Description
Currently, dialogs can only start sibling dialogs that are defined within the same DialogSet. This PR adds the ability for a dialogs to also start dialogs that are defined in a parents dialog set.

## Specific Changes
- Added a public `DialogContext.parent` property which can be used to assign a parent DC. 
- Added a public `DialogContext.findDialog()` method which will first search the local DialogSet for a dialog and then search its parents DC if not found.
- Updated all dialog lookups in the DialogContext class to use new findDialog() method.
- Updated `ComponentDialog` to wire up a parent DC chain it works through the dialog stack.
- Updated `WaterfallStepContext` class to properly clone parent DC.

## Testing
Added new Unit Tests to verify that a dialog in a component can call a dialog in its parent component.


johnataylor:
@Stevenic nice change! **but don't close the DCR** - its now assigned to me to do the C#.

PR #738, opened 2019-01-31
add UnitTestTranscriptLogger 

tomlm:
Fixes #https://github.com/Microsoft/BotBuilder/issues/5279

## Description
Add FileTranscriptLogger and change unit tests to generate transcripts

## Specific Changes
* add FileTranscriptLogger
* added helper static on TestAdapter.CreateConversation()
* changed unit tests to use FileTranscriptLogger

## Testing
All of the tests are using it.


libraries/botbuilder/src/fileTranscriptLogger.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        const transcriptPath: string = this.getTranscriptPath(activity.conversation.id);
+        let activities: Activity[] = [];
+        if (await fsx.exists(transcriptPath)) {
+            if (this.unitTestMode === true && this.started.has(transcriptPath) === false) {
+                this.started.add(transcriptPath);
+            }
+            else {
+                const json = await fsx.readFile(transcriptPath, 'utf8');
```
cleemullins: This code has the same bug as the C#. Over time, performance will progressively degrade as the file gets larger and larger. 

This code also has bugs in that multiple processes / activities will lead to data loss on the file. 

The design of the file format needs to be revisited.

tomlm: renamed to unitTestTranscriptLogger to clearly delineate this is not for production code.  FileTranscriptStore correctly stores activities in performant way but which is not easy to consume by loading into emulator directly.

cleemullins:
Pushing this until @gabog has completed his Testing Manifesto and we have a concrete overall e2e plan in place.

PR #743, opened 2019-01-31
sort and extend list of channel constants

### LINKED ISSUE #733: Channels list missing Skype for Business
## Describe the bug
Raising this as requested by @cleemullins to track against a recent fix on the .Net side to add Skype for Business to the list of available channels.  This was reported on the .Net side in issue (#1255)[https://github.com/Microsoft/botbuilder-dotnet/issues/1255] and resolved with PR (#1319)[https://github.com/Microsoft/botbuilder-dotnet/pull/1319] .  This PR also made changes to the oautprompt to remove the use of hardcoded channel id strings.

[bug]

johnataylor:
fixes #733

### DIFF
--- a/libraries/botbuilder-dialogs/src/choices/channel.ts
+++ b/libraries/botbuilder-dialogs/src/choices/channel.ts
@@ -11,20 +11,21 @@ import { TurnContext } from 'botbuilder-core';
  * @private
  */
 export const channels: any = {
-    facebook: 'facebook',
-    skype: 'skype',
-    msteams: 'msteams',
-    telegram: 'telegram',
-    kik: 'kik',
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-dialogs/src/choices/channel.ts
```diff
@@ -11,20 +11,21 @@ import { TurnContext } from 'botbuilder-core';
  * @private
  */
 export const channels: any = {
+    console: 'console',
```
stevengum: We don't have an official console adapter, do we still need this? (We just have a sample...)

If someone were developing a bot for the console, I think they would know that cards, buttons, etc are not supported.

PR #752, opened 2019-02-03
Update TokenAPI from new swagger; adds optional channelId property

Jeffders:
Description
1. Updated the TokenAPI.json swagger file based on recent TokenAPI service updates
2. This introduces a new, optional channelId property on many of hte TokenAPIs


cleemullins:
Looking into why the primary build didn't run for this PR. I've tagged @godsanchez to look into it.

PR #766, opened 2019-02-09
Update LUIS to new (Azure) nodejs SDK 

daveta:
In conjunction with https://github.com/Microsoft/botbuilder-dotnet/pull/1367.
- Update to latest Azure LUIS SDK.
- Azure Autorest has updated their version of Autorest which uses Axios
- Nock test updated to reflect


libraries/botbuilder-ai/package.json
```diff
@@ -20,14 +20,17 @@
...
   "dependencies": {
+    "@azure/cognitiveservices-luis-runtime": "^1.2.0",
     "@microsoft/recognizers-text-date-time": "1.1.2",
     "@types/html-entities": "^1.2.16",
     "@types/node": "^10.12.18",
     "@types/request-promise-native": "^1.0.10",
-    "azure-cognitiveservices-luis-runtime": "^1.0.0",
     "botbuilder": "~4.1.6",
+    "build": "^0.1.4",
+    "fs-extra": "^7.0.1",
```
justinwilaby: It appears that including `fs-extra` will introduce 2 dependencies that do the same thing into our production code. `async-file` is used currently and it makes sense to pick on and standardize on it. `fs-extra` is my first choice but `async-file` is fine too.

justinwilaby:
It looks like the unit tests were updated but coveralls is reporting the [entire error block](https://coveralls.io/builds/21548905/source?filename=libraries%2Fbotbuilder-ai%2Fsrc%2FluisRecognizer.ts#L281) missing coverage. Any ideas what happened here?

libraries/botbuilder-ai/package.json
daveta: Thanks Justin for the review!  I'm not sure where you're seeing the dependency on `async-file`.  `fs-extra` appears to be being used by the bot framework, so we should probably keep that.
The new set of bits is adoption of [new ms-rest stuff](https://www.npmjs.com/package/ms-rest) which doesn't appear to have a async-file dependency (?).

daveta:
> It looks like the unit tests were updated but coveralls is reporting the [entire error block](https://coveralls.io/builds/21548905/source?filename=libraries%2Fbotbuilder-ai%2Fsrc%2FluisRecognizer.ts#L281) missing coverage. Any ideas what happened here?

Nice catch, I checked in fixed tests!

PR #767, opened 2019-02-10
Updated fileTranscriptStore.ts and azureBlobTranscriptStore.ts to use async/await

### LINKED ISSUE #768: FileTranscriptStore erroneously reports a continuationToken when no more items exist
FileTranscriptStore erroneously reports a continuationToken when the token specified is the *last* item in the series.

## Repro:

1. Add 100 activities to an instance of the `FileTranscriptStore`
2. Call `getTranscriptActivities()` with the previously reported `continuationToken` until the `pagedResult.continuationToken` is null.

**OBSERVE:** 6 iterations are required with the last iteration yielding exactly 0 items.
**EXPECTED:** With a page size of 20 items, 5 iterations should be required to retrieve 100 items with each iteration yielding 20 items.

justinwilaby:
This PR provides the following:
- Fixes #768 
- updates `fileTranscriptStore.ts` and `azureBlobTranscriptStore.ts` to use `async/await`
- Updates unit tests in idiomatic JS
- simplifies code, removes gluttonous loops and unneeded function calls.


justinwilaby:
I'm seeing mixed formatting in this project. the `.editorconfig` shows this
```bash
[*]
indent_style = tab
indent_size = 4
trim_trailing_whitespace = true
```
but I'm seeing spaces instead of tab characters in places. Is there a formatting preference here? Should I set up an autoformat on commit for this repo?

PR #772, opened 2019-02-11
Fixes #755 - null connection name should not be allowed

### LINKED ISSUE #755: OauthPrompt Fails to Notify Consumer when ConnectionName is invalid
Affects: `botbuilder-dialogs@4.2.0`

[oauthPrompt.ts:137:0](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts#L137) does not throw when the connectionName is null or invalid which produces an OauthCard with an invalid link as described in https://github.com/Microsoft/BotFramework-Emulator/issues/1268

justinwilaby:
Fixes #755 and provides better parity with .net [BotFrameworkAdapter.cs#L534](https://github.com/Microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder/BotFrameworkAdapter.cs#L534)

### DIFF
--- a/libraries/botbuilder/tests/botFrameworkAdapter.test.js
+++ b/libraries/botbuilder/tests/botFrameworkAdapter.test.js
@@ -1,6 +1,6 @@
 const assert = require('assert');
 const { TurnContext } = require('botbuilder-core');
-const { ChannelValidation } = require('botframework-connector');
+const connector = require('botframework-connector');
 const { BotFrameworkAdapter } = require('../');
 const os = require('os');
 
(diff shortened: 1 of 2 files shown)


libraries/botbuilder/tests/botFrameworkAdapter.test.js
```diff
@@ -1,6 +1,6 @@
 const assert = require('assert');
 const { TurnContext } = require('botbuilder-core');
-const { ChannelValidation } = require('botframework-connector');
+const connector = require('botframework-connector');
```
stevengum: Is this to make use of some static value? Combined with line [743](https://github.com/Microsoft/botbuilder-js/pull/772/files#diff-edbd5a55c0b905aed85dba82e657c9c9R743) where you take the `TokenApiClient` out of `connector`, I'm not sure why this seemingly odd syntax is being used.

justinwilaby: This is just to replace the TokenApiClient with my mock for testing. This takes advantage of the `module.exports` object being static and mutable. I can require in the whole thing, replace it with my mock and all other occurrences of `new TokenApiClient()` will be the mock.

stevengum: Ah I see what you mean; no, this looks great and seems cleaner than the alternative.

PR #780, opened 2019-02-15
add support for LUIS endpoint to LuisRecognizer, tests

stevengum:
Fixes Microsoft/BotBuilder#5255

## Specific Changes
  - Adds support for passing in a LUIS Application Endpoint as retrieved from the LUIS.ai portals
  - Adds overload for passing in string instead of `LuisApplication` instance
  - Adds unittests for new feature
 
## Notes:
This PR differs from the C# PR in that `LuisApplication` is an interface in TS, not a class. Therefore the parsing logic for the endpoint is performed inside of the constructor for `LuisRecognizer`.

C# parity PR is located here: Microsoft/botbuilder-dotnet#1378

### DIFF
--- a/libraries/botbuilder-ai/src/luisRecognizer.ts
+++ b/libraries/botbuilder-ai/src/luisRecognizer.ts
@@ -9,6 +9,7 @@ import { LUISRuntimeClient as LuisClient, LUISRuntimeModels as LuisModels } from
 import { RecognizerResult, TurnContext } from 'botbuilder';
 import * as msRest from 'ms-rest';
 import * as os from 'os';
+import * as Url from 'url-parse';
 
 const pjson = require('../package.json');
 
@@ -119,12 +120,37 @@ export class LuisRecognizer {
 
     /**
      * Creates a new LuisRecognizer instance.
-     * @param application An object conforming to the [LuisApplication](#luisapplication) definition.
+     * @param application An object conforming to the [LuisApplication](#luisapplication) definition or a string representing a LUIS application endpoint, usually retrieved from https://luis.ai.
      * @param options (Optional) options object used to control predictions. Should conform to the [LuisPrectionOptions](#luispredictionoptions) definition.
      * @param includeApiResults (Optional) flag that if set to `true` will force the inclusion of LUIS Api call in results returned by [recognize()](#recognize). Defaults to a value of `false`.
      */
-    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean) {
-        this.application = application;
+    constructor(application: string, options?: LuisPredictionOptions, includeApiResults?: boolean);
+    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean);
+    constructor(application: LuisApplication|string, options?: LuisPredictionOptions, includeApiResults?: boolean) {
+        if (typeof(application) === 'string') {
+            this.application = {} as LuisApplication;
+            const parsedEndpoint: any = Url(application);
+            if (parsedEndpoint.origin === 'null') {
+                throw new Error(`Unable to parse \`endpoint\` value (e.g. "https://westus.api.cognitive.microsoft.com") from LUIS endpoint ${application}.\nPlease make sure your endpoint is a valid LUIS Endpoint.`);
+            }
+            this.application.endpoint = parsedEndpoint.origin;
+            this.application.applicationId = parsedEndpoint.pathname.split('apps/')[1];
+            
+            if (!this.application.applicationId) {
+                throw new Error(`Unable to parse \`applicationId\` value (e.g. "b31aeaf3-3511-495b-a07f-571fc873214b") not detected in LUIS endpoint ${application}.\nPlease make sure your endpoint is a valid LUIS Endpoint.`);
+            }
+
+            // Use exposed querystringify to parse the query string for the endpointKey value.
+            const parsedQuery = Url.qs.parse(parsedEndpoint.query);
+            if (parsedQuery && parsedQuery['subscription-key']) {
+                this.application.endpointKey = parsedQuery['subscription-key'];
+            } else {
+                throw new Error(`Unable to parse \`endpointKey\` ("subscription-key") from passed in LUIS endpoint's query string ${application}.\nPlease make sure your endpoint is a valid LUIS Endpoint.`);
+            }
+        } else {
+            this.application = application as LuisApplication;
+        }
+        
         this.options = {
             includeAllIntents: false,
             includeInstanceData: true,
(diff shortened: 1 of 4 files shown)


libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -119,12 +120,37 @@ export class LuisRecognizer {
...
-    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean) {
-        this.application = application;
+    constructor(application: string, options?: LuisPredictionOptions, includeApiResults?: boolean);
+    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean);
+    constructor(application: LuisApplication|string, options?: LuisPredictionOptions, includeApiResults?: boolean) {
+        if (typeof(application) === 'string') {
+            this.application = {} as LuisApplication;
+            const parsedEndpoint: any = Url(application);
+            if (parsedEndpoint.origin === 'null') {
+                throw new Error(`Unable to parse \`endpoint\` value (e.g. "https://westus.api.cognitive.microsoft.com") from LUIS endpoint ${application}.\nPlease make sure your endpoint is a valid LUIS Endpoint.`);
```
justinwilaby: Consider a custom Error class here to clean things up:
```js
class EndpointParseError extends Error {
  public static BAD_ORIGIN = 1;
  public static MISSING_APP_ID = 2;
  public static MISSING_ENDPOINT_KEY = 3;
  constructor(type: number, url: string) {
    super();
    this.message = this.getMessage(type, url);
  }
  private getMessage(type: number, url): string {
    switch (type) {
     case EndpointParseError.BAD_ORIGIN:
      return 'a custom message'
     //...
  }
}
//...
// usage
if (parsedEndpoint.origin === 'null') { 
  throw new EndpointParseError(EndpointParseError.BAD_ORIGIN, luisApplication);
}
```

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -119,12 +120,37 @@ export class LuisRecognizer {
...
+     * @param application An object conforming to the [LuisApplication](#luisapplication) definition or a string representing a LUIS application endpoint, usually retrieved from https://luis.ai.
      * @param options (Optional) options object used to control predictions. Should conform to the [LuisPrectionOptions](#luispredictionoptions) definition.
      * @param includeApiResults (Optional) flag that if set to `true` will force the inclusion of LUIS Api call in results returned by [recognize()](#recognize). Defaults to a value of `false`.
      */
-    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean) {
-        this.application = application;
+    constructor(application: string, options?: LuisPredictionOptions, includeApiResults?: boolean);
+    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean);
+    constructor(application: LuisApplication|string, options?: LuisPredictionOptions, includeApiResults?: boolean) {
+        if (typeof(application) === 'string') {
```
justinwilaby: These validation rules should be applied to the `LuisApplication` object as well since it exposes a potentially invalid state:
```js
// Invalid LuisApplication
{
  applicationId: ''
  endpointKey: ''
} as LuisApplication;
```

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -119,12 +120,37 @@ export class LuisRecognizer {
...
      * Creates a new LuisRecognizer instance.
-     * @param application An object conforming to the [LuisApplication](#luisapplication) definition.
+     * @param application An object conforming to the [LuisApplication](#luisapplication) definition or a string representing a LUIS application endpoint, usually retrieved from https://luis.ai.
      * @param options (Optional) options object used to control predictions. Should conform to the [LuisPrectionOptions](#luispredictionoptions) definition.
      * @param includeApiResults (Optional) flag that if set to `true` will force the inclusion of LUIS Api call in results returned by [recognize()](#recognize). Defaults to a value of `false`.
      */
-    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean) {
-        this.application = application;
+    constructor(application: string, options?: LuisPredictionOptions, includeApiResults?: boolean);
+    constructor(application: LuisApplication, options?: LuisPredictionOptions, includeApiResults?: boolean);
```
justinwilaby: This class allows writes to `this.application` but does not update `this.luisClient` which can lead to unpredictable behavior.
```js
const luisRecognizer = new LuisRecognizer({applicationId: "234", endpoint:"http", endpointKey:"234"}); 
luisRecognizer.application = {}; // oops! no validation, no regeneration of the luisClient. This is writable but should not be.
```
A solution would be to create accessors for the `application` property that performs validation and generates the `luisClient` when set then `Object.freeze()` the passed in `application` to make it immutable.
```js
private _application: LuisApplication;
public set application(value: LuisApplication | string) {
  this._luisApplication = this.validateAndParseLuisApplication(value);
  // validate, create a luisClient then Object.freeze(this._application) to make it immutable 
}
public get application(): LuisApplication {
 return this._luisApplication;
}
```

johnataylor: can't we just make this.application setter private? if you want to change it create a new LuisRecognizer

libraries/botbuilder-ai/src/luisRecognizer.ts
johnataylor: i don't think we use custom Error classes elsewhere at this point. I would have thought we can skip this for now.

johnataylor:
approved - assuming Justin's points are addressed

libraries/botbuilder-ai/src/luisRecognizer.ts
justinwilaby: Sure. The constructor in this case is quite messy. This is more about tidying up a bit.

libraries/botbuilder-ai/src/luisRecognizer.ts
justinwilaby: The visibility of this field is `private` but that does not prevent someone from making changes to a passed in Object.
```js
const config = {applicationId: "234", endpoint:"http", endpointKey:"234"};
const luisRecognizer = new LuisRecognizer(config); 
config.endpoint = ""; // <-- I can change it here
(luisRegognizer as any).application.endpoint = ""; // <--- I can change it here too
```
`Object.freeze()` makes this object immutable and throws at runtime if someone tries.

libraries/botbuilder-ai/src/luisRecognizer.ts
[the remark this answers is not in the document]
stevengum: Too much time in Python 3, fixed 😄

stevengum: Done!

libraries/botbuilder-ai/src/luisRecognizer.ts
stevengum: Per offline discussions with @justinwilaby, instead extracted the values from a passed in LuisApplication instance and then constructed the LuisRecognizer.application instance with primitive values to avoid this problem.

PR #781, opened 2019-02-15
Making DialogSet extendable by making properties protected

algomax:
## Description
The dialogs and dialogState properties of the DialogSet class are private. This makes it difficult to extend the class in order to add additional functionality. I have updated the property modifiers from private to protected. This provides a balance between the class encapsulation and the ability to extend it.  

## Specific Changes
  - properties dialogs and dialogState of the DialogSet class were changed to protected


cleemullins:
@stevenic @johnataylor What's the plan with this issue?

gabog:
Making these protected won't solve the issue because we never pass DialogSet, we always new it up internally. 
Consider building your own DialogSet instead.

PR #796, opened 2019-02-22
add full nock validation for LUIS calls

johnataylor:
additional validation for:
- full path including appId, not just regex match for 'apps'
- query string
- subscription key in headers (twice)
- request content


johnataylor:
Note this is a PR for master not for 4.3 branch. This is test code.

PR #800, opened 2019-02-24
Luis SDK tests (oracle)

daveta:
SDK tests (ported oracle tests) based on recognizer tests..

We can augment with more.  Separate package we can supply to Autorest team.

Based on @chrimc62 oracle tests.


cleemullins:
Should we have these same tests on the C# Side? Seems like we probably should.

cleemullins:
@daveta - Please merge when ready. :)

PR #805, opened 2019-03-05
Replace async-file with fs-extra to match other packages

### LINKED ISSUE #786: Cleanup: Use either async-file or fs-extra in the libraries and tests
## Versions
What package version of the SDK are you using.  __4.2.x-4.3.x__

## Describe the bug
`async-file` and `fs-extra` are both used in the libraries, and they both do the same thing, so we should reduce down to one package.

In `botbuilder` we use `async-file` for the library and for tests. 

In `botframework-config`, the `tools/` folder,  we use `fs-extra` for the library and helper code. `fs-extra` is also used in the tests for `botbuilder-azure` and `botbuilder-ai`, but isn't listed as a devDependency for these two projects. (The tests rely on the hoisted dependencies, otherwise they will fail with import errors) 

@justinwilaby FYI

Edit: added `fs-extra` as a devDependency to botbuilder-ai in #766

[bug]

benbrown:
Fixes #786 

## Description
Replace async-file package with fs-extra.

These packages provide the same functionality with slightly different syntax.

Removing dependence on async-file reduces the packages we need to worry about.

## Specific Changes

* replaced dependence in package file
* replaced calls to .delete with .remove
* replaced calls to .exists with .pathExists

## Testing

Run existing tests

### DIFF
--- a/libraries/botbuilder/package-lock.json
+++ b/libraries/botbuilder/package-lock.json
@@ -1,0 +1,36 @@
+{
+  "name": "botbuilder",
+  "version": "4.1.6",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "fs-extra": {
+      "version": "7.0.1",
+      "resolved": "https://registry.npmjs.org/fs-extra/-/fs-extra-7.0.1.tgz",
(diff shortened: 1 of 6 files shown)


sgellock:
I'd propose we lock any sagger generated dependencies.  Any of our 1st party components, e.g timex, etc.  for 3rd party components, I'd propose we use ~ versioning and *not* ^ versioning.

benbrown:
thank you gentlemen.

Issue #745, opened 2019-02-01
Make a consistent and strict usage of tslint/eslint across bot framework, samples and generators


kichinsky:
**Is your feature request related to a problem? Please describe.**
The usage of linting tools and applied rules is not consistent across the bot framework, generators and samples. It would be helpful from code quality and coding experience perspectives to have one shared approach everywhere. 

E.g., the echobot generator enforces using single quotes!

**Describe the solution you'd like**
Setup some share policy on using linting tools and default rules.

[enhancement] 
Related #728

benbrown:
I think if we implement eslint everwhere, this will be a good start. (See #728)

We'll be applying _typescript_ rules to typescript code, so there won't be a 100% match on the linter rules, but we'll at least be using the same tool and broadly the same set of rules.

benbrown:
We've now implemented eslint on the main botbuilder-js project, which incorporates the same ruleset that is applied to the samples.

benbrown:
Upon review of the samples work in progress, I believe this is done!

PR #806, opened 2019-03-05
Re-enable linting

### LINKED ISSUE #727: tslint doesn't lint botbuilder-js repo
## Versions
4.2.0

## Describe the bug
Running command `npm run tslint` (or more specifically `tslint ./libraries/*/src/*.ts ./libraries/*/src/**/*.ts  -t verbose`) from botbuilder-js root

results in:
`No valid rules have been specified for TypeScript files`

## Expected behavior
For linting rules to run

## Screenshots
From travis:
image

[bug]

benbrown:
Fixes #727 #745 

## Description

This is an attempt to get linting re-enabled with a well tuned set of rules we can apply to both the core libraries and the samples and templates.

## Specific Changes

- enable eslint with typescript extensions
- start with full set of recommended rules
- also apply the same or equivalent rules as those applied to javascript and typescript samples
- apply all automatic fixes
- add `npm run eslint` and `npm run eslint-fix` to root package


benbrown:
for reference:
https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/javascript_typescript/02.a.echobot/tslint.json

https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/javascript_nodejs/02.a.echobot/.eslintrc.js

benbrown:
When/before we merge this, we will need to change the build tests to run eslint instead of tslint.  It currently fails.

benbrown:
Now that the new set of tests are passing which includes running eslint instead of tslint, I think we can merge this and make those tests THE TESTS.

@cleemullins want to do the honors?

Issue #774, opened 2019-02-12
Lerna is a dependency but is not installed automatically nor listed in documentation


chrimc62:
## Describe the bug
If you sync the SDK and do 'npm install' it will complain about lerna not being there.  Either install should work or it should be documented you need to install it. 

[bug]

Aliandi:
Hey @chrimc62! After reviewing your issue we agreed that lerna should be listed bellow the devDependencies to make working of the SDK a smoother experience. We will add it on the PR #802. Meanwhile, we also found some documentation regarding this topic [here](https://github.com/Microsoft/botbuilder-js/wiki/Building-the-SDK#build-and-test-locally), where lerna is described as required to work with the SDK.

PR #811, opened 2019-03-11
Removing lerna and typescript install from travis.yml

Aliandi:
Relates to Issue #774 and PR #802 

## Description
In order to centralize the dependencies and versions, we removed the installing of `lerna` and `typescript` from `travis.yml` and added them to the devDependencies on the `package.json` file. As `mocha` is already a devDependency there are no changes regarding that nor `nyc`, that gets installed when `npm install` is ran inside `.\tools`.

## Specific Changes
  - Remove `npm install --global lerna@3.2.1 nyc mocha typescript@2.7.2` from the `before_install` at `travis.yml`
  - Add `npm install` to the `install` at `travis.yml
 - Add `"typescript": "2.7.2"` to `package.json`

## Testing

Working Travis build after changes:
imagen


GasparAcevedoZainSouthworks:
Replaced with #839

PR #813, opened 2019-03-13
add escape character support

feich-ms:
Fixes #

## Specific Changes

  - 
  -
  -


libraries/botbuilder-ai/src/lg/evaluator.ts
```diff
@@ -144,6 +148,25 @@ export class Evaluator extends AbstractParseTreeVisitor<string> implements LGFil
         return this.evalutationTargetStack[this.evalutationTargetStack.length - 1];
     }
 
+     private EvalEscapeCharacter(exp: string): string
+        {
+            var validCharactersDict = {
```
Danieladu: "const" should be used here

feich-ms: Thx, fixed. Just learnt the differences of var/let/const.

PR #837, opened 2019-03-26
 Add userId parameter to signOutUser

Aliandi:
Parity with C#

## Description
It was found that the TypeScript method signOutUser() differed from the C# implementation of [signOutUserAsync()](https://github.com/Microsoft/botbuilder-dotnet/blob/924eaa286fe3cebfaebae27570f504977c8591e9/libraries/Microsoft.Bot.Builder/BotFrameworkAdapter.cs#L642). The TypeScript version expects less parameters, not letting the user pass a userId.

_TypeScript_
`public async signOutUser(context: TurnContext, connectionName: string): Promise<void>`

_C#_
`public virtual async Task SignOutUserAsync(ITurnContext turnContext, string connectionName = null, string userId = null, CancellationToken cancellationToken = default(CancellationToken))`

## Specific Changes
  - add userId optional parameter to the method
  - add handling for when userId is null


stevengum:
@cleemullins, @BruceHaley any idea on this **BotBuilder-JS-master-4.4-CI** build? I just noticed this isn't in 4.3.x for JS but it is for C#....

PR #839, opened 2019-03-26
Removing lerna and typescript install from travis.yml

GasparAcevedoZainSouthworks:
Relates to Issue #774 and PR #802

## Description
In order to centralize the dependencies and versions, we removed the installing of `lerna` and `typescript` from `travis.yml` and added them to the devDependencies on the `package.json` file. As `mocha` is already a devDependency there are no changes regarding that nor `nyc`, that gets installed when `npm install` is ran inside `.\tools`.

## Specific Changes

- Remove `npm install --global lerna@3.2.1 nyc mocha typescript@2.7.2` from the `before_install` at `travis.yml`
- Add `npm install` to the `install` at `travis.yml`
- Add `"typescript": "2.7.2"` to `package.json`

## Testing
Working Travis build after changes:
image


sgellock:
We're no longer using travis for our builds.  also, lerna should be removed in place of using yarn for mono repo links

GasparAcevedoZainSouthworks:
> I'm not sure what we should do with this. On the one hand, we don't build with Travis anymore as we've moved to DevOps. I _believe_ this means we can effectivly delete this entire file.
> 
> On the other hand, our automated build is doing this:
> image
> 
> ... is there a better way to do this?

Hi @cleemulins, as the `travis.yml` is not used anymore, we updated this PR for removing this file.

Regarding the automated build, we are researching if there is a better approach for doing this.

cleemullins:
Build updated to run "npm test" in place of "npm travis:test". Merging.

PR #842, opened 2019-03-31
Parity with c# - binary compat changes for LuisRecognizer

daveta:
Matching changes to [C# PR for binary compat with LUIS recognizer](https://github.com/Microsoft/botbuilder-dotnet/pull/1561) which moves TelemetryClient into options class.

Also fix broken test and make hook async.

### DIFF
--- a/libraries/botbuilder-ai/src/luisRecognizer.ts
+++ b/libraries/botbuilder-ai/src/luisRecognizer.ts
@@ -102,6 +102,16 @@ export interface LuisPredictionOptions {
      * (Optional) The time zone offset for resolving datetimes.
      */
     timezoneOffset?: number;
+
+    /**
+     * (Optional) Telemetry Client.
+     */
+    telemetryClient?: BotTelemetryClient;
+
+    /**
+     * (Optional) Designates whether personal information should be logged in telemetry.
+     */
+    logPersonalInformation?: boolean;
 }
 
 export interface LuisRecognizerTelemetryClient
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-ai/tests/luisRecognizer.test.js
```diff
@@ -115,7 +115,7 @@ function TestJson(file, done, includeAllIntents, includeInstance, telemetryClien
     var newPath = expectedPath + ".new";
     var context = new TestContext({ text: expected.text });
     var recognizer = new LuisRecognizer({ applicationId: luisAppId, endpointKey: endpointKey }, 
-        { includeAllIntents: includeAllIntents, includeInstanceData: includeInstance }, true, telemetryClient, logPersonalInformation);
+        { includeAllIntents: includeAllIntents, includeInstanceData: includeInstance, telemetryClient: telemetryClient, logPersonalInformation: logPersonalInformation }, true, telemetryClient, logPersonalInformation);
```
Zerryth: It looks like you moved telemetry & PII into the options argument, which is great!
But shouldn't we also remove the extra `telemetryClient` and `logPersonalInformation` that comes after "`true`" (the `includeApiResults` parameter)?

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -316,7 +327,7 @@ export class LuisRecognizer implements LuisRecognizerTelemetryClient{
      * @param telemetryProperties Additional properties to be logged to telemetry with the LuisResult event.
      * @param telemetryMetrics Additional metrics to be logged to telemetry with the LuisResult event.
      */
-    protected onRecognizerResults(recognizerResult: RecognizerResult, turnContext: TurnContext, telemetryProperties?: {[key: string]:string}, telemetryMetrics?: {[key: string]:number}): Promise<void> {
+    protected async onRecognizerResults(recognizerResult: RecognizerResult, turnContext: TurnContext, telemetryProperties?: {[key: string]:string}, telemetryMetrics?: {[key: string]:number}): Promise<void> {
         this.fillTelemetryProperties(recognizerResult, turnContext, telemetryProperties).then(props => {
```
Zerryth: think we need an `await` for `this.fillTelemetryProperties()`

Zerryth: Looks like you did await on the C# side too 
https://github.com/Microsoft/botbuilder-dotnet/blob/5b3075cbac8a2c15d953282f472ced63dd6f931b/libraries/Microsoft.Bot.Builder.AI.LUIS/LuisRecognizer.cs#L190

Zerryth:
Looks parity-fulfilling 👍 
Just a couple things to address, but other than that looks good

libraries/botbuilder-ai/tests/luisRecognizer.test.js
daveta: Awesome!  Thanks for the feedback!

libraries/botbuilder-ai/src/luisRecognizer.ts
daveta: Great catch!!

PR #843, opened 2019-04-01
Upgrade LUIS SDK to latest typescript version

daveta:
The LUIS SDK is generated by the Azure team, and the underlying generated Autorest-based SDK is moving to a new typescript-generated back-end (the Javascript version is going to be deprecated).

This moves us to the new typescript SDK.


daveta:
Ah, looks like there might be more to this.. Closing this

PR #856, opened 2019-04-04
Handle textless messages differently, bringing Node into parity with .NET

v-kydela:
Fixes https://github.com/Microsoft/botbuilder-dotnet/issues/1349

## Description
While there wasn't necessarily a bug in this repo, there was potential for LuisRecognizer to make a needless call to a LUIS endpoint when an utterance was empty. This PR gets the behavior of LuisRecognizer to match a [recent V4 .NET PR](https://github.com/Microsoft/botbuilder-dotnet/pull/1566) which in turn is based off of V3 .NET. Now the `recognize` function will check if an utterance is empty and then create a fake result instead of getting a result from the endpoint in that case.

## Specific Changes

  - `LuisRecognizer.recognize` now creates a `RecognizerResult` with a blank intent if the activity's text is null or whitespace
  - The EmptyText.json test in luisRecognizer.test.js now checks the actual `RecognizerResult` instead of just the `topIntent` function because `topIntent` converts the empty intent to None

## Testing
The EmptyText test has been modified to fit the new behavior.


v-kydela:
@cleemullins - All right, I've gone ahead and did the linting in its own PR and requested you as a reviewer: https://github.com/Microsoft/botbuilder-js/pull/867

I must have idly auto-formatted the document without realizing how much actually got changed

cleemullins:
@v-kydela When I look at the Diff of this PR, I still see all the linting changes. I did merge in the separate PR with those changes earlier this evening, so I'm surprised to see them still in this PR. 

Can you take a look?

v-kydela:
@cleemullins - It looks to me like your merge successfully removed the linting changes from this PR:
image

Can you look again? Maybe clear your cache or something? If you need me to change this PR then I could try a rebase but that could get messy. Then of course the last resort would be to discard this PR and just create a new one.

Issue #840, opened 2019-03-28
PartitionKey value must be supplied for this operation


GasparAcevedoZainSouthworks:
## Versions
botbuilder  4.3.0
botbuilder-azure  4.3.4
nodejs version v10.14.2
Windows 10 pro v1803

## Describe the bug
When performing a delete operation in a partitioned cosmosDB collection, the following error is thrown: 
```
“{  
   “code”:“BadRequest”,
   “message”:   “PartitionKey value must be supplied for this operation.\r\nActivityId:b46cf62e-d77f-4911-97d4-b8e86186a348,
   \r\nRequestStartTime:2019-03   -27   T20:17:05.5558598   Z,
   RequestEndTime:2019-03   -27   T20:17:05.5558598   Z,
   Number of regions attempted:1   \r\n,
   Microsoft.Azure.Documents.Common/2.2.0.0”
}”
```
 

## To Reproduce
1. Install the AzureCosmoDb emulator

2. Create a partitioned collection and insert a new document (Sample image below)

2.1 Use the following information to create a new Cosmos Database in the emulator:
````
Database name: bot-cosmos-DB
Collection name: bot-collection
PartitionKey: /city
````
2.2 Add a new document using the following information
````
{
    "id": "001",
    "name": "contoso",
    "city": "EEUU"
}
````

3. Use the [echobot-with-counter](https://github.com/Microsoft/BotBuilder-Samples/tree/master/samples/javascript_typescript/02.b.echobot-with-counter)

4. Connect the bot with the CosmosDb created 
```
const policyConfigurator = (policy) => policy.DisableSSLVerification = true;
const getSettings = () => ({
    serviceEndpoint: 'https://localhost:8081',
    authKey: 'C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==',
    databaseId: 'bot-cosmos-db',
    collectionId: 'bot-collection'
});
```

5. Update the code to attempt to delete the document 
5.1 Insert the next code in the onTurn method of the echoBot class
```
const storage = new CosmosDbStorage(getSettings(), policyConfigurator);

        if (turnContext.activity.type === ActivityTypes.Message) {

            try {
                await storage.delete([turnContext.activity.text]);     
            } catch (ex) {
                await turnContext.sendActivity(`Error message when trying to delete in cosmos DB storage`);
[...]

PR #860, opened 2019-04-05
Add PartitionKey value to Read and Delete operations

Bill7zz:
## Description
This Pr will Fix the issue #840 
Also, we updated the _Read_ method to handle a partitioned CosmosDb collection.

## Testing
Added the Test for Reading and Deleting operations on partitioned collections. 
- Read with partitionKey
- Read without partitionKey
- Delete with partitionKey
- Delete without partitionKey

Note: To get the nock responses files, we ran the test in `record` mode after that, we set it back to `lockdown` mode. 

image


cleemullins:
@Bill7zz Could you "touch" the PR in some way to retrigger the build? We changed the build names and branch protection rules on Friday and it looks like this PR was caught in the middle.

PR #879, opened 2019-04-17
Dependency: update node.js app insights sdk to 1.2.0

markwolff:
## Description

Update version of node.js app insights SDK. Versions >= 1.1.0 are much more stable as it deprecates usage of zone.js.
https://github.com/Microsoft/ApplicationInsights-node.js/releases


jspruance:
If we're updating an npm package we probably want to commit the corresponding 'package-lock.json' change as well.

PR #880, opened 2019-04-17
Changed en/decryption cipher to aes-256-cbc to work with Electron 4+.

tonyanziano:
Fixes #https://github.com/Microsoft/BotFramework-Emulator/issues/1431#issuecomment-484228177

## Description

The Emulator recently upgraded from Electron 2.0.12 to Electron 4.1.1. Electron 4 ships with only one SSL library (BoringSSL) instead of the two (BoringSSL & OpenSSL) in previous versions. [(Source #1)](https://github.com/electron/electron/issues/16195#issuecomment-450990183)

This means that many of the recognized hashing functions used in the `crypto` library have been cut, one of which is the `aes256` function which is used by `botframework-config`. This breaks any encryption & decryption in the emulator, so encrypted bots can no longer be used.

However, `aes256` was just an alias for `aes-256-cbc`. [(Source #2)](https://github.com/electron/electron/issues/16195#issuecomment-453273919) [(Source #3)](https://github.com/nodejs/node/blob/master/deps/openssl/openssl/crypto/evp/c_allc.c#L188) This means that we can just change the cipher name to `aes-256-cbc` and we will achieve backwards compatibility while no longer breaking the Emulator.

## Specific Changes

  - changed the cipher name used in en/decryption from `aes256` to `aes-256-cbc`

## Testing

image


cleemullins:
This looks good to me, but I would like @jspruance to signoff.

cleemullins:
... I assume this same change needs to be made to the V3 JS SDK?

PR #886, opened 2019-04-22
Add default middleware to patch location of tenantId for MS Teams

### LINKED ISSUE #884: MS Teams tenant ID location needs mapping in BotFrameworkAdapter.createConversation
## Versions
4.3

## Describe the bug
The tenant id of a given MS Teams tenant is _currently_ located in the legacy location of `activity.channelData.tenant.id`, but we expect it to be in `activity.conversation.tenantId`.

We need to handle BOTH possible locations until the MS Teams API is updated.

## To Reproduce

Using teams, call adapter.createConversation.

An error will occur about missing tenant id.

benbrown:
Fixes #884 

## Description
This change handles the in-transition locaiton of the tenant ID field for MS Teams.

## Specific Changes
Load a middleware in the constructor for BotFrameworkAdapter that will look for channelData.tenant.id and copy the value to conversation.tenantId for use in createConversation.

## Testing
* Unit tests have been updated
* Connect a bot to MS Teams, then use `adapter.createConversation` to create a DM with a user.


benbrown:
NOTE: I fixed this with a middleware because it seemed like this was the LEAST disruptive place to put code that will only be needed until the protocol change happens.

The other possible place to do this would be inside processActivity, which seemed less good.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -145,6 +145,17 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
+
+        // Relocate the tenantId field used by MS Teams to a new location (from channelData to conversation)
+        // This will only occur on actities from teams that include tenant info in channelData but NOT in conversation,
+        // thus should be future friendly.  However, once the the transition is complete. we can remove this.
+        this.use(async(context, next) => {
+            if (context.activity && context.activity.conversation && !context.activity.conversation.tenantId && context.activity.channelData && context.activity.channelData.tenant) {
+                context.activity.conversation.tenantId = context.activity.channelData.tenant.id;
+            }
+            await next();
+        });
```
tomlm: Do we need this delta in C# SDK too?

tomlm: I think you should add constraint on channelId first
**context.channelId == ChannelTypes.Teams**

---
In reply to: [277836819](https://github.com/Microsoft/botbuilder-js/pull/886#discussion_r277836819) [](ancestors = 277836819)

benbrown: @tomlm - @gabog has done the C# version of thi.

i'll add the extra constraint now!

benbrown: @tomlm - there is no ChannelTypes object in the JS SDK -- the closest I could find is part of the prompt definitions in dialogs.  I did a test against channelId === 'msteams'

PR #897, opened 2019-04-26
Johtaylo/sidecar

johnataylor:
- Fully working port of C# implementation.
- Manual tested, needs more testing.
- Unit tests to follow.

Discrepancies between the implementations is down to the subtle differences in the state implementations.


johnataylor:
@benbrown this temporarily includes more in TestBot for the moment. (We like the state but not the inspection middleware.)

cleemullins:
There's a build error with this:
```
src/inspectionMiddleware.ts(9,58): error TS2307: Cannot find module 'botframework-connector'.
```

PR #904, opened 2019-04-30
Sync Change with C#, Allow on WhiteSpace in IF ELSE

cosmicshuai:
Fixes #

## Description

This PR is a synchronization of BotBuilder-dotnet repo 4.next branch. commits of 1822, 1823 and 1824.
Add handling of IF ELSEIF ELSE conditional cases.
And allowing at most 1 whitespace in IF ELSEIF ELSE statement.

## Specific Changes

StaticCheck.ts: handle checking of IF ELSEIF ELSE conditional cases.
Extractor.ts: modify conditionLabel variable of IF ELSEIF ELSE conditional cases.
LGFileLexer.g4, LGFileParser.g4: edit to IF ELSEIF ELSE and their generated .ts files.

  - 
  -
  -


libraries/botbuilder-lg/src/extractor.ts
```diff
@@ -65,7 +65,14 @@ export class Extractor extends AbstractParseTreeVisitor<Map<string, any>> implem
         const ifRules: lp.IfConditionRuleContext[] = context.conditionalTemplateBody().ifConditionRule();
         for (const ifRule of ifRules) {
             const expressions: TerminalNode[] = ifRule.ifCondition().EXPRESSION();
-            const conditionLabel: string = ifRule.ifCondition().IFELSE().text.toLowerCase();
+            const conditionNode = ifRule.ifCondition();
+            const ifExpr = conditionNode.IF() != null;
+            const elseIfExpr = conditionNode.ELSEIF() != null;
```
Danieladu: should use `undefined` instead of 'null'

Danieladu: use `!==` and `===` instead of `!=` and `==`

Danieladu: install TSLint extension and fix all code format warinings

PR #905, opened 2019-04-30
Support ordinal.relative entity in LUIS

moabba:
## Description
This adds support for the "ordinal.relative" entity in LUIS and allows handling entities that don't have the "value" or "values" property in the resolution.

## Specific Changes
  - Support ordinal.relative
  - Support any resolution that doesn't have the "value"/"values" property in the resolution.


johnataylor:
We're not taking this. This is breaking change, actually a breaking change in the service. Going forward we need to take a slightly more sophisticated approach to "typing." Take a look at what the Timex datatypes library does.

PR #908, opened 2019-05-02
Added support for Teams mentions & re-enable Connector SDK tests

mdrichardson:
## Description

@EricDahlvang mentioned that this file is probably auto-generated. If anybody reviewing this can point me to the right place to make the change, that would be appreciated.

BotBuilder JS SDK doesn't support Teams @mentions. It return an "Invalid Mention Entity" error because Connector strips out `mentioned: {}` from `activity.Entities[0]`.

See here: https://github.com/OfficeDev/BotBuilder-MicrosoftTeams-node/blob/master/botbuilder-teams-js/src/teamsContext.ts#L164

You can see that this is already supported in the Dotnet version as [it allows any `JObject`](https://github.com/Microsoft/botbuilder-dotnet/blob/c9d46423153dfec62a3d0ffaeb5a08ef53d2a9fa/libraries/Microsoft.Bot.Schema/EntityEx.cs) (I tested to confirm).

## Specific Changes

Added the following fields to `Entity`:

- mentioned: ChannelAccount
- text: String

## Testing

I don't know a better way of testing this other than the normal SDK test as I'm not *super* familiar with connector. However, the @mentions work now:

image

And the Connector tests pass:

image


stevengum:
@mdrichardson let's sync in person and talk about writing a test for this code.

mdrichardson:
Note: the connector.js tests have been disabled for over a year (due to a lot of them failing): https://github.com/microsoft/botbuilder-js/blame/master/libraries/botframework-connector/tests/connector.test.js#L77

I'll be working next week to get all of the tests working and adding mine to it.

mdrichardson:
Update: I think I'm going to fully rewrite all of the Connector Conversations tests.

1. They haven't been used in over a year
2. I'm fairly certain they don't use the connectorApi properly, due to how difficult it is to make and use a connection/token
3. They rely on recordings that are difficult to deal with and add to

Instead, I'll write them all from scratch and make it easy to mock/unmock them. I'm fairly certain this will free up a bunch of files in https://github.com/microsoft/botbuilder-js/tree/master/tools/framework, as well since I don't believe they're used outside of this test file

mdrichardson:
Progress update: Finally figured out why the old tests wouldn't work: they needed `trustServiceUrl` to prevent the 401 errors. Should've known. 

Current tests include tests for `oAuthApiClient`, which doesn't exist, so I'm writing full tests for `tokenApiClient`, which does.

Still in progress..

mdrichardson:
Progress update: I've created all of the necessary tests and recordings. It took me awhile to figure out how to get it all working, since it's been pretty broken for awhile and is probably more complex than it needs to be.

All of the tests work live and mocked, except for one. Should be done tomorrow.

mdrichardson:
Update: Still working on the `getAttachment` test. It works fine live, but is proving difficult to record the nock due to it being a stream. I'd just stub it with sinon, but I keep feeling like I'm sooo close. If this is blocking for anybody, let me know and I'll do a simple stub and call it a day.

mdrichardson:
@stevengum @jspruance @tomlm 	

I've added a new commit for the tests that:

1. Fixed old tests that used `xdescribe` to disable them
2. Wrote new tests for `TokenApiClient` that match the Dotnet `OauthClient` tests
3. Wrote a test for the Teams Mention (the original point of this PR)

Notes:

* I'm not sure why the BotBuilder-JS-master-CI check failed. When I click "Details", it says I'm unauthorized. Happy to fix with mroe details
* Sorry for the no-op changes. My linter was overly aggressive
* The Connector tests use SuiteBase, which is pretty complex. I made some changes to it because as far as I could tell, nothing else uses it. They shouldn't be breaking, but that's an area worth additional review
* I added notes to the top of connector.test.js to make running the tests easier
* Sorry it took so long. I should've known the fix for the GetAttachment test would be to use fs.createReadStream() in the recording. The Newly recorded tests should do that automatically.

cleemullins:
@mdrichardson  The build for this is failing. Are you able to take a look and see what's wrong?

mdrichardson:
@cleemullins 	

Negative:

image

mdrichardson:
@cleemullins 	

Tests passed!

Issues were:

* Tester wouldn't read .env file (I think it uses its own app .env?)
* Nock recordings weren't appropriately being adjusted post-recording
* Relative image paths weren't working (maybe the tester uses an older version of Node?)

Relevant tracking issue: https://github.com/microsoft/botframework-sdk/issues/5430

I removed what I felt were irrelevant checkboxes. Let me know if I need to adjust/change anything.

libraries/botframework-connector/tests/connector.test.js
```diff
@@ -4,420 +4,595 @@
...
+*       * NOCK_OFF=false
+*
+*   To run live or unmocked:
+*    1. Supply .env with (all must be valid): 
+*       * CLIENT_ID (appID)
+*       * CLIENT_SECRET (appPass)
+*       * USER_ID (from Slack)
+*       * BOT_ID (from slack)
+*       * AZURE_SUBSCRIPTION_ID
+*       * NOCK_OFF=true
```
stevengum: I love this.

@mdrichardson do we need to update the build config in Azure DevOps for this added configuration? I'd love to have these tests run live nightly and for when we do RC builds.

mdrichardson: Assuming that by "run live", you mean unmocked, we'd need to add the environment variables to the build config. At least when committing this PR, it didn't seem that the test server used the .env file at all and instead fell back on defaults. You don't seem to need a slack-enabled bot running, but you do need a valid and existing appId/appPass that has the Slack channel enabled, so it would be good to use one that's somewhat "permanent"

PR #909, opened 2019-05-02
[BotBuilder-AI] Remove 'html-entities' from QnA Maker class 

GasparAcevedoZainSouthworks:
## Description
- Remove `html-entities` library and its decoding from the answer formatting.

## Specific Changes
At this moment when the QnA Maker class in BotBuilder AI receives an answer that contains enconded HTML entities, said entities are [decoded before returning the answer](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/qnaMaker.ts#L582).

Remove this behaviour from the QnA Maker answer formatting, as this responsibility should be delegated to the user consuming the answer.

## Testing
We did a couple of tests using both the NodeJS and .NET version of the class.
A knowledge base was created with a single answer containing several encoded HTML entities.

image

We tested the answer from the knowledge base by adding it to a `console.log()` and the `sendActivity()` using the [Echo Browser](https://github.com/Microsoft/BotBuilder-Samples/tree/master/samples/javascript_es6/01.a.browser-echo) bot sample.

image
As seen in the image above, the answer returned by the QnA Maker class does not match the answer stored in the knowledge base.

We also compared the behaviour of the JavaScript class against the .NET class replicating the same test.
image

This test confirms that the behaviour of the QnA Maker class in .NET does not decode the HTML entities before returning the answer.


cleemullins:
This looks reasonable to me, as that delegation seems odd. 

@zerryth?

Zerryth:
Seems good to me

PR #913, opened 2019-05-05
Update the oracles to match LUIS.

chrimc62:
Add roles to testing.
Revert node-only change that caused passing through raw resolutions.
In particular for phone instead of phone:["425-765-6532"] the bug
would cause phonenumber: [{score:0.9, value:"425-555-1212"].


stevengum:
Closing as we're not needing this in 4.4, same changes were merged into master via #911

Issue #620, opened 2018-11-09
Make botbuilder-ai browser compatible


stevengum:
**Is your feature request related to a problem? Please describe.**
botbuilder-ai has a dependency on request which is only used in the QnAMaker class. Additionally, there exists the azure-sdk-for-js repo which contains an isomorphic luis runtime client which can be used instead of the current Node.js-specific client. We can use `node-fetch` instead of `request`, and `@azure/cognitiveservices-luis-runtime` instead of `azure-cognitiveservices-luis-runtime` to remove the Node.js-specific dependencies.

(Related to #600)

[enhancement]

stevengum:
Additionally, it looks like there are some Node.js libraries like `html-entities` that would need to be replaced.

cleemullins:
@juanar Any update on this?

gabog:
@JuanAr Any update on this?

JuanAr:
@cleemullins @gabog we still haven't had the chance to start digging into this one. However, we realized #818 is related to this one and should be tackled first.
As a ballpark ETA, we are going to fix this issue early next week.

stevengum:
Fixed with #914

PR #914, opened 2019-05-06
[BotBuilder-AI] Make BotBuilder AI web compatible

GasparAcevedoZainSouthworks:
Fixes issue #620.

Must be merged with PR #909.

## Description
Remove or replace NodeJS dependent libraries, and make the **BotBuilder-AI** library fully web-compatible.

## Changes
- Replace `request-promise-native` with `node-fetch`
- `@azure/cognitiveservices-luis-runtime` was fixed in pr # 844
- `html-entities` should be removed to keep parity with the .NET version. We will be creating a Pull Request with this change and more detailed information regarding this decision.
- Remove `request` since it was unused.

## Details
We searched through the library looking for any incompatible dependencies to remove.

 The libraries already identified by the original issue report as ones to be removed were:
•	**request-promise-native** - Suggested to be replaced by node-fetch
•	**azure-cognitiveservices-luis-runtime** - Suggested to be replaced by @azure/cognitiveservices-luis-runtime
•	**html-entities**
 
_request-promise-native_ was removed and replaced by node-fetch
_html-entities_ was straight out removed and its removal documented at `PR 909`
_azure-cognitiveservices-luis-runtime_ had been already removed by a previous commit at `e6016d4226656378aa1a370e022a8755e25e3609`
 
Furthermore, we researched the compatibility of the rest of the dependencies in **BotBuilder-AI** looking for some more libraries that could break compatibility with web deployment. The remaining dependencies then were:
 
image
 
## Testing
We used the dependencies' documentation, when specified, to check which libraries were ready for web. 
For a better, practical way, to test the web-compatibility of the library we set to deploy a bot to the web and make calls to the BotBuilder-AI. That would both allow us to test any changes we would make as well as the current state of compatibility of the library.
 
Looking at the index.ts file, several of **BotBuilder-AI** exports are internally-used structures; and aside from those, the two major exports are **LuisRecognizer** and **QnAMaker**. We made test calls to both in order for the bot to query these services.
 
image
 
As our bot we chose to use the browser **echobot project** (since it's a bot that can be deployed to web and run on a browser) and modify part of its code to make calls to **BotBuilder-AI**. We made some dummy calls at the beginning of the processActivity method, meaning the code would be triggered any time a conversation state or a message update arrived.
 
image
 
To check the response, we added the result to the existing echobot response

image
 
We got correct responses both querying the QnA service 
 
image

and the LUIS service:
 
image
 
Since the library is apparently already correctly working when deployed online (after the removals specified at the top of this post) we only ended up removing _**request**_ as a dependency since we found it was unused. The rest of the dependencies performed correctly when calling the library from a web bot.


cleemullins:
@stevengum Can you review + merge this in today?

libraries/botbuilder-ai/src/qnaMaker.ts
```diff
@@ -458,18 +458,19 @@ export class QnAMaker implements QnAMakerTelemetryClient {
 
         this.validateOptions(queryOptions);
 
-        const qnaResult: any = await request({
-            url: url,
+        const qnaResult: any = await request( url, {
```
stevengum: extra space after `(`

libraries/botbuilder-ai/src/qnaMaker.ts
```diff
@@ -458,18 +458,19 @@ export class QnAMaker implements QnAMakerTelemetryClient {
...
             method: 'POST',
             headers: headers,
             timeout: queryOptions.timeout,
-            json: {
+            body: JSON.stringify( {
```
stevengum: Another extra space after `(`

stevengum:
@cleemullins there's some minor code cleanup and the package-lock.json for the library needs to be updated, but after those changes are in this one is good to go.

@GasparAcevedoZainSouthworks

PR #915, opened 2019-05-07
Fix case of imported files for 4.next botbuilder-lg and botbuilder-expression-parser

benbrown:
There are a few instances where the case of the included file is incorrect, which breaks the library on some operating systems.

As a result, the nightly build version of `botbuilder-lg` and `botbuilder-expression-parser` don't work on Macs!


benbrown:
Do we have to kick off a special process to get these rebuilt and published to nuget?

boydc2014:
We don't have an automatic process to publish those bits now, so we will manually update the packages in myget, will let you know once it's done.

feich-ms:
@benbrown, just manually updated the npm packages of botbuilder-lg and botbuilder-expression-parser with your fixes. You can find them [here](https://botbuilder.myget.org/gallery/botbuilder-declarative). To use these packages, please add the dependency in your package.json like "botbuilder-lg": "https://botbuilder.myget.org/F/botbuilder-declarative/npm/botbuilder-lg/-/4.5.0.tgz" to consume the feed and then run npm install to install the packages.

benbrown:
I’ll try it tomorrow morning!

benbrown:
Looks like it is working now! Thanks for the help.

PR #917, opened 2019-05-08
[BotBuilder-AI] Update @azure/ms-rest-js to version 1.8.4

### LINKED ISSUE #916: [BotBuilder-AI] build failing
## Versions
BotBuilder-js latest package version.
nodejs version 10.14.2
npm version 6.9.0

## Describe the bug
The BotBuilder-ai library is not installing the latest version of the `@azure/ms-rest-js` package. 
This makes some conflicts with the `cognitiveservices-luis-runtime` library, as this is installing the latest version of the `@azure/ms-rest-js` package (1.8.4)

## More info
According to the NPM CLI documentation, one of the [package.lock.json](https://docs.npmjs.com/files/package-lock.json) goals are:
 
```Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.```

This suggests that the version defined in the package-lock.json will have priority over the ones defined in the package.json when the first one is present. 

Additionally, we found an example of the usage of the package-lock.json that clarify the current issue in the `botbuilder-ai` library. More info [Here](https://medium.com/coinmonks/everything-you-wanted-to-know-about-package-lock-json-b81911aa8ab8). 

The problem is that, by one hand the `cognitiveservices-luis-runtime` does not have a package-lock.json (that is because it must not be published, and it will be ignored. More info [here](https://docs.npmjs.com/files/package-lock.json)) and, when the `npm install` command runs, the latest version of `@azure/ms-rest-js` (v1.8.4) will be installed; by the other hand, the `botbuilder-ai` library has a package-lock.json that specify to install the version (1.8.2) of the same dependency, which will cause some dependency issues regarding the versions.

## To Reproduce
1. Clone the botbuilder-js repository
2. run `npm install`
3. run `npm run build`
4. You will see the build of `botbuilder-ai` library broken. 

## Expected behavior
Build the solution successfully.

## Screenshots
Build error:
image

## Proposed Solution
1. Update the BotBuilder-ai package.json to use @azure/ms-rest-js (~1.8.4) version, which will force the update of the package-lock.json 
2. Delete the package-lock.json and `npm install` to create it.

ParadoxARG:
Fixes #916  

## Description
Update `@azure/ms-rest-js` library in `botbuilder-ai` from version `1.8.2` to `1.8.4`(latest). This is a required change because the `@azure/cognitiveservices-luis-runtime` library is using this version and they are not compatible. 

We would also suggest changing the package version from `~1.8.4`(1.8.x) to `^1.8.4`(1.x) to provide the same compatibility as  `@azure/cognitiveservices-luis-runtime`. If we don't we may have issues if a new minor version is released. Also, we should remove `@azure/ms-rest-js` from the `package-lock.json` cause with every patch release the build may break, as `@azure/cognitiveservices-luis-runtime` is always updating to the last version.

## Specific Changes
 - change package.json to use @azure/ms-rest-js `~1.8.4`

## Testing
Build passing:
image

Tests passing:
image

### DIFF
--- a/libraries/botbuilder-ai/package-lock.json
+++ b/libraries/botbuilder-ai/package-lock.json
@@ -5,9 +5,9 @@
 	"requires": true,
 	"dependencies": {
 		"@azure/ms-rest-js": {
-			"version": "1.8.2",
-			"resolved": "https://registry.npmjs.org/@azure/ms-rest-js/-/ms-rest-js-1.8.2.tgz",
-			"integrity": "sha512-xpXLTUztG/oYcyieN0GiS9l33nFTYFm/xetaHef+wbEKevEhEfKzp7Q94nwovNwjhteNSRL2+PjY1uYOQlv3yQ==",
+			"version": "1.8.4",
+			"resolved": "https://registry.npmjs.org/@azure/ms-rest-js/-/ms-rest-js-1.8.4.tgz",
+			"integrity": "sha512-eoyRARiaf+U90T2Bs2uJ2Zqqk/cLvQujOw42Y9ZB2eRJZoGjJd7Xjh4oKuhnuVGwy8lvMM/Ffj7I77E/CJ9drg==",
 			"requires": {
 				"@types/tunnel": "0.0.0",
 				"axios": "^0.18.0",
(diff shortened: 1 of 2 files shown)


stevengum:
Closing in favor of #922.

PR #918, opened 2019-05-09
Feich/decouple lg parser

feich-ms:


libraries/botbuilder-lg/src/LGFileParser.ts
```diff
@@ -0,0 +1,52 @@
+/**
```
boydc2014: Let's just use plain javascript code, like class LGFileParser

PR #925, opened 2019-05-14
Modularize scripts

BruceHaley:
This separates the coveralls upload script from the coverage reporting script so the build can keep the coveralls repo token secret until after building and testing. Increases the security of the token.


package.json
```diff
@@ -1,49 +1,51 @@
...
-            "**/tools/**",
-            "**/internal.*"
-        ]
-    }
+  "private": true,
+  "name": "botbuilder-packages",
+  "scripts": {
+    "postinstall": "lerna bootstrap --hoist",
+    "build": "lerna run build",
+    "build-test": "lerna run build && nyc mocha \"libraries/bot*/tests/*.test.js\"",
```
stevengum: We should retain the `"test"` script instead of renaming it to `"build-test"`, as `"test"` is a [default script](https://docs.npmjs.com/misc/scripts) (meaning you can call `npm test` instead of `npm run test`). I think across all our (FUSE's) projects we use `"test"`.

@benbrown any input?

package.json
```diff
@@ -1,49 +1,51 @@
...
+    "clean": "lerna run clean",
+    "test:coveralls": "lerna run build && nyc mocha \"libraries/bot*/tests/*.test.js\" && nyc report --reporter=text-lcov | coveralls",
+    "test-coverage": "nyc mocha \"libraries/bot*/tests/*.test.js\" ",
+    "upload-coverage": "nyc report --reporter=text-lcov | coveralls",
+    "build-docs": "lerna run build-docs",
+    "eslint": "./node_modules/.bin/eslint  ./libraries/*/src/*.ts ./libraries/*/src/**/*.ts",
```
stevengum: This command can actually be shortened to `"eslint  ./libraries/*/src/*.ts ./libraries/*/src/**/*.ts",` as npm will resolve to local packages before looking for a global package.

Same for the other scripts that use `"./node_modules/.bin"`.

([docs for npm-run-script](https://docs.npmjs.com/cli/run-script))

package.json
```diff
@@ -1,49 +1,51 @@
...
+    "eslint-fix": "./node_modules/.bin/eslint  ./libraries/*/src/*.ts ./libraries/*/src/**/*.ts --fix",
+    "set-dependency-versions": "node tools/util/updateDependenciesInPackageJsons.js ./libraries ^${Version} botbuilder botbuilder-choices botbuilder-core botbuilder-prompts botframework-connector botframework-config botframework-schema && node tools/util/updateDependenciesInPackageJsons.js ./transcripts ^${Version} botbuilder botbuilder-ai botbuilder-dialogs",
+    "update-versions": "lerna run set-version && npm run set-dependency-versions"
+  },
+  "dependencies": {
+    "ms-rest-azure": "^2.5.9",
```
stevengum: Do we still need this dependency here? What is this and `"ms-rest-js"` for?

package.json
```diff
@@ -1,49 +1,51 @@
...
+    "@types/lodash": "^4.14.116",
+    "@typescript-eslint/eslint-plugin": "^1.4.2",
+    "@typescript-eslint/parser": "^1.4.2",
+    "coveralls": "^3.0.2",
+    "eslint": "^5.15.1",
+    "eslint-plugin-only-warn": "^1.0.1",
+    "lerna": "^3.2.1",
+    "mocha": "^5.2.0",
+    "tslint-microsoft-contrib": "^5.2.1",
+    "typescript": "^2.7.2"
```
stevengum: Why do we have different versions of `"typescript"` between the regular dependencies and devDependencies? This can be cleaned up I think.

Also, except for subjective cleanliness; I'm not sure I see the point between having dependencies and devDependencies in a package.json that is only used in our build processes or when a developer forks and clones the repo. @benbrown @johnataylor @cleemullins

stevengum:
If someone wants to make a separate issue and drive my feedback in through that instead of piling onto this PR I think that's fine as well.

package.json
BruceHaley: I am OK with renaming script "build-test" back to "test". I was operating under the assumption that a script named "test" would only test, not also build.

BruceHaley: Renamed. "./node_modules/.bin/" also removed.

package.json
BruceHaley: All good points to raise. I am not able to address them, however. My sole interest was in adding the 2 new scripts on lines 10 and 11. Which changes were unfortunately obscured in the diff listing by my editor's choice to reformat all white space in the document! My apologies. (I did find clicking the "Diff settings" button gives the option to ignore white space -- not the default.)

package.json
BruceHaley: Done.

package.json
BruceHaley: Done.

package.json
BruceHaley: I don't know. Per our discussion I suggest these items be handled in a separate PR or issue. (I am afraid my title "modularize scripts" was way broader than I intended.)

PR #931, opened 2019-05-20
[Expression]add more functions

feich-ms:
Add more expression functions including R1 and R2

### DIFF
--- a/libraries/botbuilder-expression-parser/tests/expression.test.js
+++ b/libraries/botbuilder-expression-parser/tests/expression.test.js
@@ -111,8 +111,10 @@ const dataSource = [
   ["count(guid())", 36],
   ["guid().indexOf('-')", 8],
   ["indexOf(guid(), '-')", 8],
+  ["indexOf(hello, '-')", -1],
   ["guid().lastIndexOf('-')", 23],
   ["lastIndexOf(guid(), '-')", 23],
+  ["lastIndexOf(hello, '-')", -1],
 
   // Logical comparison functions tests
   ["and(1 == 1, 1 < 2, 1 > 2)", false],
--- a/libraries/botbuilder-expression-parser/tests/expression.test.js
+++ b/libraries/botbuilder-expression-parser/tests/expression.test.js
@@ -315,6 +315,8 @@ const dataSource = [
   ["take(createArray('a', 'b', 'c', 'd'), one)", ['a']],
   ["subArray(createArray('a', 'b', 'c', 'd'), 1, 3)", ['b', 'c']],
   ["subArray(createArray('a', 'b', 'c', 'd'), 1)", ['b', 'c', 'd']],
+  ["range(1, 4)", [1, 2, 3, 4]],
+  ["range(-1, 3)", [-1, 0, 1]],
 
   // Object manipulation and construction functions tests
   ["string(addProperty(json('{\"key1\":\"value1\"}'), 'key2','value2'))", "{\"key1\":\"value1\",\"key2\":\"value2\"}"],
--- a/libraries/botbuilder-expression-parser/tests/expression.test.js
+++ b/libraries/botbuilder-expression-parser/tests/expression.test.js
@@ -299,6 +299,11 @@ const dataSource = [
   ["count(union(createArray('a', 'b'), createArray('b', 'c'), createArray('b', 'd')))", 4],
   ["count(intersection(createArray('a', 'b')))", 2],
   ["count(intersection(createArray('a', 'b'), createArray('b', 'c'), createArray('b', 'd')))", 1],
+  ["skip(createArray('a', 'b', 'c', 'd'), 2)", ['c', 'd']],
+  ["take(hello, two)", 'he'],
+  ["take(createArray('a', 'b', 'c', 'd'), one)", ['a']],
+  ["subArray(createArray('a', 'b', 'c', 'd'), 1, 3)", ['b', 'c']],
+  ["subArray(createArray('a', 'b', 'c', 'd'), 1)", ['b', 'c', 'd']],
 
   // Object manipulation and construction functions tests
   ["string(addProperty(json('{\"key1\":\"value1\"}'), 'key2','value2'))", "{\"key1\":\"value1\",\"key2\":\"value2\"}"],
--- a/libraries/botbuilder-expression-parser/tests/expression.test.js
+++ b/libraries/botbuilder-expression-parser/tests/expression.test.js
@@ -1,6 +1,7 @@
-const { ExpressionEngine} =  require('../');
+const { ExpressionEngine } = require('../');
 const { Extensions } = require('botbuilder-expression');
 const assert = require('assert');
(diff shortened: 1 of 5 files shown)


libraries/botbuilder-expression/src/buildInFunction.ts
```diff
@@ -970,6 +1137,76 @@ export class BuiltInFunctions {
...
+                ExpressionType.Skip,
+                BuiltInFunctions.Skip,
+                ReturnType.Object,
+                (expression: Expression): void => BuiltInFunctions.ValidateOrder(expression, [], ReturnType.Object, ReturnType.Number),
+            ),
+            new ExpressionEvaluator(
+                ExpressionType.Take,
+                BuiltInFunctions.Take,
+                ReturnType.Object,
+                (expression: Expression): void => BuiltInFunctions.ValidateArityAndAnyType(expression, 2, 2, ReturnType.String, ReturnType.Object, ReturnType.Number),
```
Danieladu: maybe `ValidateOrder` is more reasonable?

libraries/botbuilder-expression/src/buildInFunction.ts
```diff
@@ -1068,6 +1305,48 @@ export class BuiltInFunctions {
...
+            new ExpressionEvaluator(
+                ExpressionType.CountWord,
+                BuiltInFunctions.Apply((args: ReadonlyArray<any>) => args[0].trim().split(/\s+/).length, BuiltInFunctions.VerifyString),
+                ReturnType.Number,
+                BuiltInFunctions.ValidateUnaryString
+            ),
+            new ExpressionEvaluator(
+                ExpressionType.AddOrdinal,
+                BuiltInFunctions.Apply((args: ReadonlyArray<any>) => this.AddOrdinal(args[0]), BuiltInFunctions.VerifyInteger),
+                ReturnType.Number,
```
Danieladu: return type of `AddOrdinal` is string?

libraries/botbuilder-expression/src/buildInFunction.ts
```diff
@@ -1068,6 +1305,48 @@ export class BuiltInFunctions {
...
+                (expression: Expression): void => BuiltInFunctions.ValidateArityAndAnyType(expression, 1, 1, ReturnType.Number),
+            ),
+            new ExpressionEvaluator(
+                ExpressionType.Guid,
+                BuiltInFunctions.Apply(() => BuiltInFunctions.newGuid()),
+                ReturnType.String,
+                (expression: Expression): void => BuiltInFunctions.ValidateArityAndAnyType(expression, 0, 0, ReturnType.String),
```
Danieladu: there is no need to write `ReturnType.String`

libraries/botbuilder-expression/src/buildInFunction.ts
feich-ms: yes. fixed

libraries/botbuilder-expression/src/buildInFunction.ts
feich-ms: my mistake, fixed.

libraries/botbuilder-expression/src/buildInFunction.ts
feich-ms: Right as there is no param. Fixed

libraries/botbuilder-expression-parser/tests/expression.test.js
```diff
@@ -91,6 +92,27 @@ const dataSource = [
...
+  ["addOrdinal(11 + 10)", "21st"],
+  ["addOrdinal(11 + 11)", "22nd"],
+  ["addOrdinal(11 + 12)", "23rd"],
+  ["addOrdinal(11 + 13)", "24th"],
+  ["addOrdinal(-1)", "-1"],//original string value
+  ["count(guid())", 36],
+  ["guid().indexOf('-')", 8],
+  ["indexOf(guid(), '-')", 8],
+  ["guid().lastIndexOf('-')", 23],
+  ["lastIndexOf(guid(), '-')", 23],
```
Danieladu: add `indexof` and `lastIndexOf` tests that return -1

feich-ms: fixed.

libraries/botbuilder-expression/src/buildInFunction.ts
```diff
@@ -1293,6 +1626,62 @@ export class BuiltInFunctions {
...
+                BuiltInFunctions.ValidateUnary),
+            new ExpressionEvaluator(
+                ExpressionType.Binary,
+                BuiltInFunctions.Apply((args: ReadonlyArray<any>) => this.ToBinary(args[0]), BuiltInFunctions.VerifyString),
+                ReturnType.String,
+                BuiltInFunctions.ValidateUnary),
+            new ExpressionEvaluator(
+                ExpressionType.DataUri,
+                BuiltInFunctions.Apply(
+                    (args: Readonly<any>) => 'data:text/plain;charset=utf-8;base64,'.concat(Buffer.from(args[0]).toString('base64')), BuiltInFunctions.VerifyString),
```
Danieladu: personal opinion， is there a exist library to handle `datauri`, `dataUriToBinary` and `dataUriToString`. Maybe we can use a library to do more param checks. can [datauri](https://www.npmjs.com/package/datauri) help?

libraries/botbuilder-expression/src/buildInFunction.ts
```diff
@@ -970,6 +1137,76 @@ export class BuiltInFunctions {
                     BuiltInFunctions.VerifyContainer),
                 ReturnType.Number,
                 BuiltInFunctions.ValidateUnary),
+            new ExpressionEvaluator(
+                ExpressionType.Range,
+                BuiltInFunctions.ApplyWithError(
+                    (args: ReadonlyArray<any>) => {
+                        let error: string;
+                        if (args[1] < 1) {
```
cosmicshuai: should the condition be args[1] <= 0, according to the error messgae

libraries/botbuilder-expression/src/buildInFunction.ts
feich-ms: datauri package accept a file path as input which is not consistent with the one in WDL doc https://docs.microsoft.com/en-us/azure/logic-apps/workflow-definition-language-functions-reference#dataUri.

libraries/botbuilder-expression/src/buildInFunction.ts
```diff
@@ -970,6 +1137,76 @@ export class BuiltInFunctions {
...
+                            error = 'Second paramter must be more than zero';
+                        }
+
+                        let num: number;
+                        let result: number[] = [];
+                        for (num = 0; num < args[1]; num++) {
+                            result = result.concat(Number(args[0]) + num);
```
Danieladu: This method can be simplified:
```
var length: number = 4;
var start: number = 1;
var s: number[] = [...Array(length).keys()].map(e => e + start);
```

PR #935, opened 2019-05-21
make attach command a Trace activity

johnataylor:
This makes the '/INSPECT attach {etc etc}' a trace activity.

### DIFF
--- a/libraries/botbuilder/src/inspectionMiddleware.ts
+++ b/libraries/botbuilder/src/inspectionMiddleware.ts
@@ -12,6 +12,17 @@ import { Middleware, TurnContext, BotState, StatePropertyAccessor, UserState, Co
 /** @private */
 class TraceActivity {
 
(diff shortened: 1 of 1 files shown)


libraries/botbuilder/src/inspectionMiddleware.ts
```diff
@@ -12,6 +12,17 @@ import { Middleware, TurnContext, BotState, StatePropertyAccessor, UserState, Co
 /** @private */
 class TraceActivity {
 
+    public static fromCommand(command: string): Partial<Activity> {
```
tonyanziano: Might want to change this function name to `makeCommandActivity` to align with [what is currently in the 4.4 branch](https://github.com/microsoft/botbuilder-js/blob/4.4/libraries/botbuilder/src/inspectionMiddleware.ts#L15)

johnataylor: updated function name.

PR #938, opened 2019-05-22
Stevenic/4.next events

Stevenic:
## Description
Removes `DialogContext.consultDialog()` method and instead uses events to achieve the same functionality.

## Specific Changes
- Added a new `DialogContainer` class that `ComponentDialog` and `AdaptiveDialog` derive from. This lets you factory a DialogContext for the containers active child dialog/step using a `DialogContext.child` property.
- Removed `consultDialog()` method and related structures. This resulted in reverting many classes like prompts and waterfalls back to their original implementations.
- Re-worked `emitEvent()` to support bubbling events from the leaf dialog up to the root and then back down to the leaf.
- Re-worked `AdaptiveDialog` to process events on either the leading or trailing edge of the bubbling event as needed.
- Re-worked the way `AdaptiveDialog` executes steps to go along with these new changes.
- Added support for a `ConversationMembersAdded` event to simplify greeting users.
- Tweaked the way recognizer results get persisted for the turn.
- Updated rule and step classes to go along with changes.
- Tweaked some of the class names. Tried to remove all traces of the word planning.
 
## Testing
Haven't added unit tests yet but I did verify my samples are working.


libraries/botbuilder-dialogs-adaptive/src/adaptiveDialog.ts
```diff
@@ -200,93 +159,115 @@ export class AdaptiveDialog<O extends object = {}> extends Dialog<O> {
...
+                    // Recognize utterance
+                    const recognized = await this.onRecognize(sequence.context);
+                    sequence.state.setValue('turn.recognized', recognized);
+
+                    // Emit leading RecognizedIntent event
+                    handled = await this.processEvent(sequence, { name: AdaptiveEventNames.recognizedIntent, value: recognized, bubble: false }, true);
+                } else if (activity.type === ActivityTypes.Event) {
+                    // Emit leading edge named event that was received
+                    handled = await this.processEvent(sequence, { name: activity.name, value: activity.value, bubble: false }, true);
+                } else if (activity.type === ActivityTypes.ConversationUpdate && Array.isArray(activity.membersAdded)) {
```
carlosscastro: I feel a bit uneasy about the adaptive dialog having specific knowledge about things like members added and such. Seems like in an ideal world, the event handling precedence along with specific handling could be in satellite classes. If this code will be somewhat stable this shouldn't be a problem though.

libraries/botbuilder-dialogs-adaptive/src/adaptiveDialog.ts
```diff
@@ -344,237 +325,89 @@ export class AdaptiveDialog<O extends object = {}> extends Dialog<O> {
...
-                planning.queueChanges(changes[0]);
+                sequence.queueChanges(changes[0]);
                 return true;
             }
         }
 
         return false;
     }
 
-    private async queueBestMatches(planning: PlanningContext, event: DialogEvent): Promise<boolean> {
```
carlosscastro: Love deleting this code :) it was annoying to debug

libraries/botbuilder-dialogs-adaptive/src/sequenceContext.ts
```diff
@@ -0,0 +1,203 @@
...
+     * Applies any queued up changes.
+     * 
+     * @remarks
+     * Applying a set of changes can result in additional plan changes being queued. The method
+     * will loop and apply any additional plan changes until there are no more changes left to 
+     * apply.
+     * @returns true if there were any changes to apply. 
+     */
+    public async applyChanges(): Promise<boolean> {
+        const queue: StepChangeList[] = this.context.turnState.get(this.changeKey) || [];;
```
carlosscastro: double semicolon

libraries/botbuilder-dialogs-adaptive/src/sequenceContext.ts
```diff
@@ -0,0 +1,203 @@
...
+        switch (change.changeType) {
+            case StepChangeType.insertSteps:
+                Array.prototype.unshift.apply(this.steps, change.steps);
+                break;
+            case StepChangeType.insertStepsBeforeTags:
+                let inserted = false;
+                if (Array.isArray(change.tags)) {
+                    for (let i = 0; i < this.steps.length; i++) {
+                        if (this.stepHasTags(this.steps[i], change.tags)) {
+                            const args = ([i, 0] as any[]).concat(change.steps);
```
carlosscastro: I (and probably future developers) would appreciate some comments on the intention of lines 167 thru 170

libraries/botbuilder-dialogs-adaptive/src/adaptiveDialog.ts
Stevenic: I think this will eventually get externalized to another component

libraries/botbuilder-dialogs-adaptive/src/sequenceContext.ts
Stevenic: thanks

libraries/botbuilder-dialogs-adaptive/src/sequenceContext.ts
Stevenic: sure

libraries/botbuilder-dialogs-adaptive/src/adaptiveDialog.ts
Stevenic: me too :)

PR #942, opened 2019-05-24
Johtaylo/inspection unit test

johnataylor:
- unit test
- roll up fix from 4.4 on state
- upgrade nock to consistent version across code base

### DIFF
--- a/libraries/botbuilder/src/inspectionMiddleware.ts
+++ b/libraries/botbuilder/src/inspectionMiddleware.ts
@@ -34,14 +34,13 @@ class TraceActivity {
         };
     }
 
-    public static fromState(botState: BotState, turnContext: TurnContext, name: string, label: string): Partial<Activity> {
-        var obj = botState.get(turnContext);
+    public static fromState(botState: BotState): Partial<Activity> {
         return {
             type: ActivityTypes.Trace,
             timestamp: new Date(),
-            name: name,
-            label: label,
-            value: obj,
+            name: 'BotState',
+            label: 'Bot State',
+            value: botState,
             valueType: 'https://www.botframework.com/schemas/botState'
         };
     }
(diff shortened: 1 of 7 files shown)


libraries/botbuilder/src/inspectionMiddleware.ts
```diff
@@ -34,14 +34,13 @@ class TraceActivity {
         };
     }
 
-    public static fromState(botState: BotState, turnContext: TurnContext, name: string, label: string): Partial<Activity> {
-        var obj = botState.get(turnContext);
+    public static fromState(botState: BotState): Partial<Activity> {
```
cleemullins: You changed the signature here. In C# this would be a breaking change. 

IS this something we could avoid?

libraries/botbuilder/src/inspectionMiddleware.ts
```diff
@@ -253,12 +252,17 @@ export class InspectionMiddleware extends InterceptionMiddleware {
...
-                await this.invokeSend(turnContext, session, TraceActivity.fromState(this.userState, turnContext, 'UserState', 'User State'));
+                botState.userState = this.userState.get(turnContext);
             }
+
             if (this.conversationState !== undefined) {
-                await this.invokeSend(turnContext, session, TraceActivity.fromState(this.conversationState, turnContext, 'ConversationState', 'Conversation State'));
+                botState.conversationState = this.conversationState.get(turnContext);
             }
+
+            await this.invokeSend(turnContext, session, TraceActivity.fromState(botState));
```
cleemullins: MUCH cleaner code. Nice.

libraries/botbuilder/src/inspectionMiddleware.ts
johnataylor: this in 4.4 this is essentially a cherry-pick

PR #945, opened 2019-05-24
remove property name existance check from botstate

### LINKED ISSUE #894: Accessor Semantics enforce incorrect usage patterns
All Versions. 

In [Botstate.ts](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/botState.ts) there is a problem with how accessors are created:

```ts
   public createProperty<T = any>(name: string): StatePropertyAccessor<T> {
        if (this.properties.has(name)) { throw new Error(`BotState.createProperty(): a property named '${ name }' already exists.`); }
        const prop: BotStatePropertyAccessor<T> = new BotStatePropertyAccessor<T>(this, name);
        this.properties.set(name, prop);

        return prop;
    }
```

The dictionary that's used to gate the accessor creation is not needed. This prevents multiple accessors from being created, and forces bots to use global static accessors. 

This *signifigantly* differs from the C# code, which allows these to be created at-will:
```cs
        public IStatePropertyAccessor<T> CreateProperty<T>(string name)
        {
            if (string.IsNullOrWhiteSpace(name))
            {
                throw new ArgumentNullException(nameof(name));
            }

            return new BotStatePropertyAccessor<T>(this, name);
        }
```

This difference bleeds into the samples in unexpected ways, and has caused @johnataylor to :boom: :boom: :boom:. He's been :pouting_cat: as he tries to get the sidecar worked wrapped up.

johnataylor:
Fixes #894

### DIFF
--- a/libraries/botbuilder-core/src/botState.ts
+++ b/libraries/botbuilder-core/src/botState.ts
@@ -51,7 +51,7 @@ export class BotState implements PropertyManager {
      * Creates a new property accessor for reading and writing an individual property to the bot
      * states storage object.
      * @param T (Optional) type of property to create. Defaults to `any` type.
-     * @param name Name of the property to add. Must be unique within the set.
+     * @param name Name of the property to add.
      */
     public createProperty<T = any>(name: string): StatePropertyAccessor<T> {
         const prop: BotStatePropertyAccessor<T> = new BotStatePropertyAccessor<T>(this, name);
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-core/src/botState.ts
```diff
@@ -58,10 +54,7 @@ export class BotState implements PropertyManager {
      * @param name Name of the property to add. Must be unique within the set.
```
cleemullins: Is this comment still true? I don't think so.

PR #967, opened 2019-06-05
Declarative Type Loader first iteration

carlosscastro:
Provides support for loading declarative dialog format into fully hydrated dialog objects, including advanced features such as heterogeneous polymorphic collections, implicit references and so on.

What is ready:

- Single file loading
- First batch of unit tests
- First batch of declarative end-to-end tests
- Multiple file loading using basic FileResourceProvider
- Implicit references, i.e. just put a string with a resource id when a dialog object is expected
- Dynamic type factory
- Type builders: different ways of registering classes in the factory, control class re-hydration using different primitives
    - Default type builder using constructor
    - Configurable type builder for Configurable classes, which uses the configure method to hydrate the object from the json 
    - Custom type builder for folks that want to use lambda based factory registrations

What is missing 

- FileResourceProvider: Change notifications. Right now the dialog needs to be reloaded to reflect changes. This  only affects editing of dialogs.
- ResourceProvider caching: We are not doing any caching in resource providers. Arguably it is not a problem at this point because it is only required at load time. However for large bots or future network-based resource providers this will be a problem
- Async resource loading in FileREsource: Originally wrote using fs.promises but it is experimental so went with a synchronous solution until I discuss with stevenic on the best approach. Trivial to rewrite with promises
- Explicit copy: Merge objects when asking for an explicit $copy and want to override properties


libraries/botbuilder-dialogs-adaptive/src/input/choiceInput.ts
```diff
@@ -40,6 +40,14 @@ export class ChoiceInput extends DialogCommand implements DialogDependencies {
     /**
      * Activity to send the user.
      */
+    public set prompt(value: string|Partial<Activity>) {
```
carlosscastro: This was needed because the original implementation was constructor based, but we need property based setting of this from a string.

libraries/botbuilder-dialogs-adaptive/src/input/confirmInput.ts
```diff
@@ -38,6 +38,14 @@ export class ConfirmInput extends DialogCommand implements DialogDependencies {
     /**
      * Activity to send the user.
      */
+    public set prompt(value: string|Partial<Activity>) {
```
carlosscastro: This was needed because the original implementation was constructor based, but we need property based setting of this to allow declarative patterns

libraries/botbuilder-dialogs-adaptive/src/input/numberInput.ts
```diff
@@ -35,6 +35,14 @@ export class NumberInput extends DialogCommand implements DialogDependencies {
     /**
      * Activity to send the user.
      */
+    public set prompt(value: string|Partial<Activity>) {
```
carlosscastro: This was needed because the original implementation was constructor based, but we need property based setting of this to allow declarative patterns

libraries/botbuilder-dialogs-adaptive/src/input/textInput.ts
```diff
@@ -39,7 +39,15 @@ export class TextInput extends DialogCommand implements DialogDependencies {
     /**
      * Activity to send the user.
      */
-    public activity = new ActivityProperty();
+    public set prompt(value: string|Partial<Activity>) {
```
carlosscastro: This was needed because the original implementation was constructor based, but we need property based setting of this to allow declarative patterns

libraries/botbuilder-dialogs-adaptive/src/steps/sendActivity.ts
```diff
@@ -46,19 +46,30 @@ export class SendActivity extends DialogCommand {
...
     /**
      * Activity to send the user.
      */
-    public activity = new ActivityProperty();
+    private activityProperty = new ActivityProperty();
+
+    /**
+     * Public getter and setter for declarative activity configuration
+     */
+    public get activity(): Partial<Activity>|string {
```
carlosscastro: This was needed because the original implementation was constructor based, but we need property based setting of this to allow declarative patterns

carlosscastro:
@Stevenic Could you please take a look?

libraries/botbuilder-dialogs-declarative/src/resources/fileResource.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public id(): string {
+        return this.resourceId;
+    }
+
+    public readText() : Promise<string> {
+        const filePath = this.path;
+        return new Promise<string>(function(resolve, reject) {
+            fs.readFile(filePath, 'utf8', (err, data) => {
```
tomlm: >readFile [](start = 15, length = 8)

there are async versions of this available, shouldn't we use them?

libraries/botbuilder-dialogs-declarative/src/resources/resource.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Licensed under the MIT License.
+ */
+
+const path = require('path');
+const fs = require('fs');
+
+export interface IResource {
+    id(): string;
+    readText() : Promise<string>;
+}
```
tomlm: C# interface looks like this:
  public interface IResource
    {
        /// <summary>
        /// Resource name
        /// </summary>
        string Id { get; }

        /// <summary>
        /// Get resource as text async
        /// </summary>
        /// <returns></returns>
        Task<string> ReadTextAsync();

        /// <summary>
        /// Get resource as text
        /// </summary>
        /// <returns></returns>
        string ReadText();

        /// <summary>
        /// Get readonly stream
        /// </summary>
        /// <returns></returns>
        Stream OpenStream();
    }
???

tomlm:
> 

oh I don't think we want to keep this csx stuff do wee?

---
Refers to: libraries/botbuilder-dialogs-declarative/tests/resources/13 - CustomStep/catyears.csx:7 in 3ffae9f. [](commit_id = 3ffae9f42565edec71c3f5ce70d3da6d53bf36d1, deletion_comment = False)

PR #980, opened 2019-06-12
Swagger connector api updates

jspruance:
Fixes #943 - 'Add State field to Semantic Action in Schema'

## Description
This pull request contains updates for the Connector API, based on changes in the swagger spec since Nov 2018:
https://github.com/microsoft/botframework-sdk/blob/master/specs/botframework-protocol/botframework-channel.json

## Specific Changes

**Change log:**

'localTimestamp' - update description
'localTimezone - update description
botbuilder-js\libraries\botframework-schema\src\index.ts
botbuilder-js\libraries\botframework-schema\src\activityInterfaces.ts

SemanticActions / SemanticActionStates - add 'state' field:
botbuilder-js\libraries\botframework-schema\src\index.ts
botbuilder-js\libraries\botframework-connector\src\connectorApi\models\mappers.ts

ConversationParameters / add 'tenantId' field
botbuilder-js\libraries\botframework-schema\src\index.ts (ALREADY EXISTS - NO CHANGE)
botbuilder-js\libraries\botframework-connector\src\connectorApi\models\mappers.ts

ConversationAccount / add 'tenantId' field
botbuilder-js\libraries\botframework-schema\src\index.ts (ALREADY EXISTS - NO CHANGE)
botbuilder-js\libraries\botframework-connector\src\connectorApi\models\mappers.ts

TokenResponse / add 'channelId' field

botbuilder-js\libraries\botframework-connector\src\tokenApi\models\index.ts (ALREADY EXISTS - NO CHANGE)
botbuilder-js\libraries\botframework-connector\src\tokenApi\models\mappers.ts (ALREADY EXISTS - NO CHANGE)

botbuilder-js\libraries\botframework-schema\src\index.ts
botbuilder-js\libraries\botframework-connector\src\connectorApi\models\mappers.ts

AttachmentView viewId / size descriptions
botbuilder-js\libraries\botframework-schema\src\index.ts

Activity / add 'callerId' field
botbuilder-js\libraries\botframework-schema\src\index.ts
botbuilder-js\libraries\botframework-connector\src\connectorApi\models\mappers.ts

## Testing
1) Manual testing with local bots
2) botbuilder-js\libraries\botbuilder\tests\botFrameworkAdapter.test.js
3) Additional tests could be added to 'connector.test.js', once that test suite is refactored and back i a working state (in progress - see https://github.com/microsoft/botbuilder-js/pull/908).

PR #984, opened 2019-06-13
fix import in inspectionMiddleware.ts

stevengum:
## Description
Stops importing directly from botframework-schema in botbuilder, which does not have a direct dependency on this library.

Instead import from botbuilder-core, which exposes botframework-schema

This makes the InspectionMiddleware class usable outside of the repo/project.


stevengum:
These changes were fixed via #980

PR #988, opened 2019-06-13
Add assert contains

baruchiro:
From https://github.com/microsoft/botbuilder-dotnet/pull/1922

## Specific Changes
Add option to assert that message **contain** a substring.

### DIFF
--- a/libraries/botbuilder-core/src/testAdapter.ts
+++ b/libraries/botbuilder-core/src/testAdapter.ts
@@ -482,13 +482,17 @@ export class TestFlow {
      * @param description (Optional) Description of the test case. If not provided one will be generated.
      * @param timeout (Optional) number of milliseconds to wait for a response from bot. Defaults to a value of `3000`.
      */
-    public assertReply(expected: string | Partial<Activity> | TestActivityInspector, description?: string, timeout?: number): TestFlow {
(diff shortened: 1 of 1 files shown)


stevengum:
@baruchiro the build failed with this error:

```bash
src/testAdapter.ts(492,51): error TS2345: Argument of type 'string | TestActivityInspector' is not assignable to parameter of type 'string'.
  Type 'TestActivityInspector' is not assignable to type 'string'.
```

libraries/botbuilder-core/src/testAdapter.ts
```diff
@@ -482,13 +482,17 @@ export class TestFlow {
      * @param description (Optional) Description of the test case. If not provided one will be generated.
      * @param timeout (Optional) number of milliseconds to wait for a response from bot. Defaults to a value of `3000`.
      */
-    public assertReply(expected: string | Partial<Activity> | TestActivityInspector, description?: string, timeout?: number): TestFlow {
+    public assertReply(expected: string | Partial<Activity> | TestActivityInspector, description?: string, timeout?: number, isSubstring: boolean = false): TestFlow {
```
cleemullins: I'm probably missing the obvious. Can you help me understand? 

The C# has the following signature:
```cs
public TestFlow AssertReplyContains(string expected, string description = null, uint timeout = 3000)
```

Why would the JS have a different signature?

PR #1002, opened 2019-06-18
963 update dialog cancelled return value

### LINKED ISSUE #963: Updated ComponentDialog so it returns DialogTurnStatus.Cancelled when the dialog is cancelled
Need to add parity with this PR:

https://github.com/microsoft/botbuilder-dotnet/pull/1993

jspruance:
Fixes #963

## Description
Return a DialogTurnResult with status: DialogTurnStatus.cancelled when dialog cancelled.

## Specific Changes
Changes to 'beginDialog' and 'continueDialog' methods to return the proper 'cancelled' status:
botbuilder-js\libraries\botbuilder-dialogs\lib\componentDialog.js

## Testing
Modified existing unit tests to check for DialogTurnStatus.cancelled when dialog cancelled.


libraries/botbuilder-dialogs/tests/componentDialog.test.js
```diff
@@ -236,7 +236,7 @@ describe('ComponentDialog', function () {
         const adapter = new TestAdapter(async turnContext => {
             const dc = await dialogs.createContext(turnContext);
             const results = await dc.beginDialog('component');
```
jspruance: Updated test to check for new 'cancelled' status (previously it was 'complete')

libraries/botbuilder-dialogs/tests/componentDialog.test.js
```diff
@@ -87,7 +87,7 @@ describe('ComponentDialog', function () {
         const startDialog = new WaterfallDialog('start', [
             async (step) => {
                 simpleStepContextCheck(step);
```
jspruance: Believe the intent of this test is to ensure status of 'complete' is returned when calling endComponent(). Since we now always return 'cancelled' when the purpose of calling endComponent() is to cancel, updating this test to cover endComponent() when cancellation is not explicit.

cleemullins:
+johnataylor - could you take a quick look?

PR #1032, opened 2019-07-03
add at mention removal code and used it in inspection

johnataylor:
JavaScript equivalent of C# https://github.com/Microsoft/BotBuilder-dotnet/issues/2060

Unit tests included - however this needs more manual testing with Teams tomorrow. Specifically need to validate the exact shape of the activity coming in from Teams regarding the entities that carry additional at mention metadata.

I think the at mention removal code looks a little sketchy (looks like it could stamp on data accidentally?) but for now this code is exactly the same as the C# right down to the use of look-ahead/behind regex.


johnataylor:
Note I haven't included update package-lock files here.

cleemullins:
Holding off Merging, as the comment indicates more testing is required.

johnataylor:
tested end to end with Teams - found and fixed one issue see later commits

libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -79,6 +79,70 @@ export class TurnContext {
...
+    }
+
+    /**
+     * Returns the mentions on an activity.
+     *
+     * ```JavaScript
+     * const mentions = TurnContext.getMentions(context.request);
+     * ```
+     * @param activity The activity to alter the text on
+     * @param id The recipient id of the at mention
```
stevengum: This docstring is incorrect

PR #1034, opened 2019-07-03
DialogTestClient updates

gabog:
* Removed callback from constructor (we don't want to expose that one for now).
* Removed adapterOptions from constructor.
* Refactored constructor to take a channeldId or a TestAdapter (required).
* Added optional conversationState parameter to constructor
* Added conversationUpdate property to allow devs to manipulate state during tests.

Updated Readme to match current changes
Updated tests to pass.
Updated .editorconfig to enforce spaces instead of tabs

### DIFF
--- a/libraries/botbuilder-testing/README.md
+++ b/libraries/botbuilder-testing/README.md
@@ -18,7 +18,7 @@ Use the DialogTestClient to drive unit tests of your dialogs.
 
 To create a test client:
 ```javascript
-let client = new DialogTestClient(dialog_to_test, dialog_options, OptionalMiddlewares);
+const client = new DialogTestClient('test', dialog_to_test, dialog_options, OptionalMiddlewares);
 ```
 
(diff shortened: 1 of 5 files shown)


libraries/botbuilder-testing/README.md
```diff
@@ -37,21 +37,21 @@ Here is a sample unit test using assert:
...
 
 // send a test message, catch the reply
 let reply = await client.sendActivity('hello');
-assert(reply.text == 'hello yourself', 'first message was wrong');
+assert.strictEqual(reply.text, 'hello yourself', 'first message was wrong');
 // expecting 2 messages in a row?
-reply = await client.getNextReply();
-assert(reply.text == 'second message', 'second message as wrong');
+reply = client.getNextReply();
+assert.strictEqual(reply.tex, 'second message', 'second message as wrong');
```
benbrown: typo reply.tex

PR #1035, opened 2019-07-03
QnAMakerService correctly builds hostname URL & throws error w/o URL

### LINKED ISSUE #939: QnaMakerService constructor throws
[this line](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botframework-config/src/models/qnaMakerService.ts#L43) throws when the `hostname` is omitted from the `source` argument but `hostname` is not write protected and should be allowed to be set at a later time. 

Also, if `hostname = 'https://myqna.azurewebsites.net/qnamaker` in the `source` param passed to the constructor, `hostname` then becomes `https://myqna.azurewebsites.net/qnamaker/qnamaker` which is wrong.

Zerryth:
Fixes #939 

## Description
* Issue 1: URL built for hostname would _always_ append "`/qnamaker`" to `QnAMakerService.hostname`, even when the `hostname` already had `/qnamaker` in the hostname string
  * **QnAMakerService will now only append `/qnamaker` to hostname, if it does not end in `/qnamaker` already**
  * Ex 1: `https://myservice.azurewebsites.net` => `https://myservice.azurewebsites.net/qnamaker`
  * Ex 2: `https://MyServiceThatDoesntNeedAppending.azurewebsites.net/qnamaker` remains the same

* Issue 2: previously if `QnAMakerService`'s first parameter receives a `source` that does not have a `hostname`, in the URL class's logic, it would throw invalid URL error when constructing URL + base to create `QnAMakerService.hostname`
  * As per Steven's advice, **QnAMakerService now throws an error, clearly stating that the hostname is required in the source to construct the QnAMakerService properly**
  * this is because nothing actually can be done w/regards to QnAMaker unless you have a hostname

### DIFF
--- a/libraries/botframework-config/src/models/qnaMakerService.ts
+++ b/libraries/botframework-config/src/models/qnaMakerService.ts
@@ -39,8 +39,14 @@ export class QnaMakerService extends ConnectedService implements IQnAService {
      */
     constructor(source: IQnAService = {} as IQnAService) {
         super(source, ServiceTypes.QnA);
+        
+        if (!source.hostname) {
+            throw TypeError('QnAMakerService requires source parameter to have a hostname.')
+        }
 
-        this.hostname = new URL('/qnamaker', this.hostname).href;
+        if (!this.hostname.endsWith('/qnamaker')) {
+            this.hostname = new URL('/qnamaker', this.hostname).href;
+        }
     }
 
     // encrypt keys in service
--- a/libraries/botframework-config/tests/service.test.js
+++ b/libraries/botframework-config/tests/service.test.js
@@ -29,5 +29,22 @@ describe("Service Tests", () => {
         });
         assert.equal(qna.hostname, "https://myservice.azurewebsites.net/qnamaker");
     });
+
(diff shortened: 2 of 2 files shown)


libraries/botframework-config/src/models/qnaMakerService.ts
```diff
@@ -39,8 +39,14 @@ export class QnaMakerService extends ConnectedService implements IQnAService {
      */
     constructor(source: IQnAService = {} as IQnAService) {
         super(source, ServiceTypes.QnA);
+        
+        if (!source.hostname) {
+            throw TypeError('QnAMakerService requires source parameter to have a hostname.')
+        }
 
-        this.hostname = new URL('/qnamaker', this.hostname).href;
+        if (!this.hostname.endsWith('/qnamaker')) {
```
cleemullins: If I'm being picky, "/qnamaker" isn't a host name. 

"/qnamaker" is a path segment. 

In this case, I think the underlying IQnAService is messed up, so we can't change it.

PR #1044, opened 2019-07-10
Streaming Extensions Preview Package

DDEfromOR:
## Description
This PR is needed because the people demand the ability to stream near real time content to and from bots and we aren't going to tell them no. This library introduces the ability for bots to communicate over NamedPipes and WebSockets in addition to their existing ability to use HTTP. This allows for exciting use cases like streaming audio and authorization at the connection level instead of per request.

## Specific Changes
The following classes have been added to the surface area:
	• BrowserWebSocket
	• CancellationToken
	• IPayloadReceiver
	• IPayloadSender
	• ISocket
	• IStreamingTransportServer
	• IStreamingTransportClient
	• NamedPipeClient
	• NamedPipeServer
	• NamedPipeTransport
	• NodeWebSocket
	• PayloadReceiver
	• PayloadSender
	• ProtocolAdapter
	• ReceiveRequest
	• ReceiveResponse
	• RequestHandler
	• StreamingRequest
	• StreamingResponse
	• WebSocketClient
	• WebSocketServer 
        • WebSocketTransport

## Testing
There are slightly over 100 tests added in the library's Tests dir. Currently there is a deficiency in coverage over the Integration namespace, but this will be addressed before GA.


DDEfromOR:
I'm troubleshooting the build issues, it looks like a dependency issue that for some reason didn't show up locally. I should have a fix pushed later today.

libraries/botframework-streaming-extensions/src/Transport/TransportConstants.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @module botframework-streaming-extensions
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+export class TransportContants {
```
DDEfromOR: Spelling!

PR #1052, opened 2019-07-12
Version fix: "botbuilder-testing": "4.5.0" from invalid version 4.1.6

BruceHaley:
Version 4.1.6 does not exist. 
This is a regression from PR "Lock exact versions across dependencies"  https://github.com/microsoft/botbuilder-js/pull/1040 which broke the Run-JS-Functional-Tests-Linux build.

This fix should immediately fix the build. However, there are a lot of other dependencies specifying an exact version match of 4.1.6 instead of 4.5.0, which forces the code to run on 8-month-old packages.

I guess that means we will have to get good at rolling all dependencies every time new package versions are released. (I recall there is a tool to do this.)

### DIFF
--- a/libraries/testbot/package.json
+++ b/libraries/testbot/package.json
@@ -13,7 +13,7 @@
     "botbuilder": "4.1.6",
     "botbuilder-ai": "4.1.6",
     "botbuilder-dialogs": "4.1.6",
-    "botbuilder-testing": "4.1.6",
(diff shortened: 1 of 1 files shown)


cleemullins:
This won't fix the issue. 

We need to update the build to include the version stamping. We do that stamping during the nightly builds, but not during the functional builds. 

That way every package has the same version, and it's "new" each time the build is run.

PR #1053, opened 2019-07-12
Fix ImportResolverDelegate bug

Danieladu:
Fix: https://github.com/microsoft/botbuilder-dotnet/issues/2233

Add source property in  `ImportResolverDelegate(string resourceId)` interface.

C# version: https://github.com/microsoft/botbuilder-dotnet/pull/2232

### DIFF
--- a/libraries/botbuilder-lg/src/ImportResolver.ts
+++ b/libraries/botbuilder-lg/src/ImportResolver.ts
@@ -12,17 +12,16 @@ import * as path from 'path';
 /**
  * Delegate for resolving resource id of imported lg file.
  */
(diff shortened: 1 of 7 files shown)


libraries/botbuilder-lg/src/ImportResolver.ts
```diff
@@ -12,17 +12,16 @@ import * as path from 'path';
 /**
  * Delegate for resolving resource id of imported lg file.
  */
-export declare type ImportResolverDelegate = (resourceId: string) => { content: string; id: string };
+export declare type ImportResolverDelegate = (source: string, resourceId: string) => { content: string; id: string };
```
feich-ms: >source [](start = 46, length = 6)

You used sourceId in C# version

PR #1056, opened 2019-07-12
[DO NOT MERGE] Remove spread operators from cardFactory.ts

corinagum:
Fixes https://github.com/microsoft/BotBuilder-Samples/issues/1618
This bug was verified on Edge ver. 44.18362.1.0

## Description
Edge does not support the spread operator, so I the replaced instances causing error with `Object.assign({}, varname)` in `cardFactory.ts, which is what the error in the above issue specifies.

However, this doesn't remove the spread operator from the entire repo. Is this something that we want removed everywhere? The spread operator is used quite frequently throughout the project. A quick search of `...` in all `.ts` files produces 79 other 

## Specific Changes
  - `cardFactory.ts` no longer uses spread operator in `mediaCard` and `signinCard`
  - This has **not** removed the spread operator from the rest of the project

## Testing

I did not add new tests.

### DIFF
--- a/libraries/botbuilder-core/src/cardFactory.ts
+++ b/libraries/botbuilder-core/src/cardFactory.ts
@@ -245,7 +245,7 @@ export class CardFactory {
             images = text;
             text = undefined;
         }
-        const card: Partial<ThumbnailCard> = { ...other };
+        const card: Partial<ThumbnailCard> = Object.assign({}, other);
         if (title) { card.title = title; }
         if (text) { card.text = text; }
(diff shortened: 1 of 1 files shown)


sgellock:
DO NOT MERGE.  we're discussing the issue and I think we may want to fix in a different manner

compulim:
I am fixing the sample `javascript_es6/01.browser-echo` so Babel will transpile `node_modules/botbuilder-sdk`. Don't need this fix (but we need a few more because of other issues).

PR #1059, opened 2019-07-15
[Expression] Change accessProperty to case insensitive

feich-ms:
Before the change, the accessProperty function is case sensitive. After the change, the accessProperty function is case insensitive, which means getProperty(A, 'name') equals to getProperty(A, 'name'). This change is to alige with the C# version which is a key issue there.

### DIFF
--- a/libraries/botbuilder-expression-parser/tests/expression.test.js
+++ b/libraries/botbuilder-expression-parser/tests/expression.test.js
@@ -386,6 +386,7 @@ const dataSource = [
 
   // Memory access tests
   ["getProperty(bag, concat('na','me'))", "mybag"],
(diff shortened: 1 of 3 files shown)


libraries/botbuilder-expression-parser/tests/expression.test.js
```diff
@@ -386,6 +386,7 @@ const dataSource = [
 
   // Memory access tests
   ["getProperty(bag, concat('na','me'))", "mybag"],
+  ["getProperty(bag, 'Name')", "mybag"],
```
boydc2014: can we have a little bit more test case
like bag.List[bag.Index - 2]

PR #1061, opened 2019-07-15
Merge master into 4.future

Stevenic:
Merged master into 4.future


libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -79,6 +79,72 @@ export class TurnContext {
...
+        var result: Mention[] = [];
+        if (activity.entities !== undefined) {
+            for (var i=0; i<activity.entities.length; i++) {
+                if (activity.entities[i].type.toLowerCase() === 'mention') {
+                    result.push(activity.entities[i] as Mention);
+                }
+            }
+        }
+        return result;
+    }
```
tomlm: these methods aren't in dotnet...?

PR #1064, opened 2019-07-16
Add botbuilder-dialogs-adaptive to script set-dependency-versions.

BruceHaley:
This adds the dependency botbuilder-dialogs-adaptive to the set-dependency-versions script.

Should fix the bad dependency reference that is breaking build BotBuilder-JS-4.future-daily.


BruceHaley:
Build is still broken. Looking...

PR #1077, opened 2019-07-17
[DO NOT MERGE] Fixed removeMention regex for IE

### LINKED ISSUE #1075: RegExp in `removeMentionText` is not supported in IE11
## Versions
What package version of the SDK are you using: `4.5.0`
What nodejs version are you using: N/A
What browser version are you using: IE11

## Describe the bug

[The regular expression pattern on this line of `turnContext.ts`](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/turnContext.ts#L117) is not supported in IE11.

## To Reproduce

1. Open IE11 F12
2. In the "Console" tab, type `/(?<=<at.*>)(.*?)(?=<\/at>)/i`

## Expected behavior

IE11 should returns the `RegExp` object.

In fact, it say `Unexpected quantifier`.

## Screenshots

image

## Additional context

Found this while fixing `javascript_es6/01.browser-echo` sample.

[bug]

mdrichardson:
Fixes #1075 	

## Description

IE doesn't support negative/positive lookaheads/lookbehinds in RegExp, so this removes it.

## Specific Changes

RegExp changed from:

```js
/(?<=<at.*>)(.*?)(?=<\/at>)/i
```

To:

```
/<at.*>(.*?)<\/at>/i
```

This also removes two of the capture groups, which is fine, because we only need one.

## Testing

Tests already existed:

image

### DIFF
--- a/libraries/botbuilder-core/src/turnContext.ts
+++ b/libraries/botbuilder-core/src/turnContext.ts
@@ -114,7 +114,7 @@ export class TurnContext {
         var mentions = TurnContext.getMentions(activity);
         for (var i=0; i<mentions.length; i++) {
             if (mentions[i].mentioned.id === id) {
-                var mentionNameMatch = mentions[i].text.match(/(?<=<at.*>)(.*?)(?=<\/at>)/i);
+                var mentionNameMatch = mentions[i].text.match(/<at.*>(.*?)<\/at>/i);
                 if (mentionNameMatch.length > 0) {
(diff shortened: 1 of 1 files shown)


stevengum:
Do not merge this, @johnataylor probably has some choice words on the implementation of the removeMention code. 

The removeMention's implementation needs to be reevaluated across languages.

mdrichardson:
@stevengum Understood. Should we just close this PR, then?

PR #1078, opened 2019-07-17
Fix oAuthPrompt Catch Block

### LINKED ISSUE #1054: Oauth Prompt: Bug in try...catch block in recognizeToken function
## Versions
v4.5.0

v8.9.4

What os are you using
MSFT Windows

## Describe the bug
App terminates with an "Unexpected token error (this token ' { ')" on line 235 in the oauthPrompt.js during the oauth flow.

## Expected behavior
App to start and stay alive.

[bug]

mdrichardson:
Fixes #1054 

## Description

Per [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#Syntax), in JavaScript, the `catch` in try/catch blocks *must* have the exception statement (something like `catch (err)`.

## Specific Changes

Changes `catch {...}` to `catch (e) {...}`

## Testing

All existing tests pass. I can add one for this, but that seems a bit overkill.

image


mdrichardson:
@cleemullins 	Per linked issue, looks like lots of OAuthPrompt users are running into this. Can we roll this into the next patch release?

PR #1082, opened 2019-07-18
Fix tests in 4.future

BruceHaley:
This fixes 5 tests in build BotBuilder-JS-4.future-daily.

That leaves another 11 tests still failing.


BruceHaley:
4.future branch abandoned.

PR #1083, opened 2019-07-19
Prevent Common LUIS Endpoint User Error

mdrichardson:
Prevents: 

https://github.com/microsoft/BotBuilder-Samples/issues/1659
https://stackoverflow.com/questions/57105966/botbuilder-nlp-with-dispatch-error-no-such-host-is-known

I know I've seen this in past issues, as well. I also hit this occasionally, myself.

## Description

When constructing `LuisRecognizer`, user needs to supply an Endpoint. It isn't *super* clear whether the format is supposed to be any of the following:

* `https://westus.api.cognitive.microsoft.com/luis/v2.0/apps/...`
* `westus`
* `westus.api.cognitive.microsoft.com`

This problem occurs mostly in the NLP samples because:

1. `.env` calls for `LuisAPIHostName=`
2. And it builds the Recognizer with `endpoint: https://${ process.env.LuisAPIHostName }.api.cognitive.microsoft.com`

**The samples aren't consistent with this, either**

[**C# CoreBot**](https://github.com/microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/13.core-bot/FlightBookingRecognizer.cs#L24)

```csharp
$"https://" + configuration["LuisAPIHostName"]);
```

[**C# NLP**](https://github.com/microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/14.nlp-with-dispatch/BotServices.cs#L20)

```csharp
$"https://{configuration["LuisAPIHostName"]}.api.cognitive.microsoft.com"),
```

So, the user MUST supply only `westus` and everything else results in an unclear error (depending on what they supply).

## Specific Changes

Uses regex to allow for any of the above formats.

## Alternatives Considered

The actual `LuisRecognizer` hints what is supposed to be supplied:

```js
export interface LuisApplication {
    [...]
    /**
     *  (Optional) LUIS endpoint with a default of https://westus.api.cognitive.microsoft.com
     */
    endpoint?: string;
    [...]
}
```

Instead of this PR, we could maybe make the sample more descriptive (maybe call it `LuisRegion` instead of `LuisAPIHostName` - **this might actually be best**), do a validation check on the environment variables, or change it to use the whole endpoint URL.

@v-kydela mentioned this might break LUIS hosted on custom domains. So, maybe this isn't as good of a solution as changing the `.env` variable name.

**Let me know if we like this idea and I'll write up the C# version**

## Testing

See tests in file changes.

Testing just LuisRecognizer.tests.js:

image

### DIFF
--- a/libraries/botbuilder-ai/src/luisRecognizer.ts
+++ b/libraries/botbuilder-ai/src/luisRecognizer.ts
@@ -180,7 +180,7 @@ export class LuisRecognizer implements LuisRecognizerTelemetryClient {
             const { applicationId, endpoint, endpointKey } = application;
             this.application = {
                 applicationId: applicationId,
-                endpoint: endpoint,
+                endpoint: this.getValidEndpoint(endpoint),
                 endpointKey: endpointKey
             };
         }
@@ -750,5 +750,17 @@ export class LuisRecognizer implements LuisRecognizerTelemetryClient {
         if (!this.application.endpointKey) {
             throw new Error(`Invalid \`endpointKey\` value detected: ${this.application.endpointKey}\nPlease make sure your endpointKey is a valid LUIS Endpoint Key, e.g. "048ec46dc58e495482b0c447cfdbd291".`);
         }
+    }
+
+    /**
+     * Allows users to pass in either Machine Name (LUIS region or custom <name>.api...) or the full URL
+     * 
+     * @param endpoint https://<region/custom>.api.cognitive.microsoft.com
+     */
+    private getValidEndpoint(endpoint: string): string|null {
+        if (!endpoint) return null;
+
+        const machineName = endpoint.match(/(?:https?:\/\/)?([A-Z]*)(?:.*)?/i)[1];
+        return `https://${ machineName }.api.cognitive.microsoft.com`;
     }
 }
--- a/libraries/botbuilder-ai/tests/luisRecognizer.test.js
+++ b/libraries/botbuilder-ai/tests/luisRecognizer.test.js
@@ -520,7 +520,7 @@ describe('LuisRecognizer', function () {
         });
     });
 
-    it('should successfully construct with valid endpoint.', () => {
+    it('should successfully construct with valid endpoint when given the full application', () => {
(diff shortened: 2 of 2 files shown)


mdrichardson:
@cleemullins @stevengum 	Any thoughts on this? I'm leaning towards just changing the samples vs. merging this PR.

jwiley84:
It was apparently named to mimic QnA? But all it does is confuse people.

image

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -751,4 +751,16 @@ export class LuisRecognizer implements LuisRecognizerTelemetryClient {
             throw new Error(`Invalid \`endpointKey\` value detected: ${this.application.endpointKey}\nPlease make sure your endpointKey is a valid LUIS Endpoint Key, e.g. "048ec46dc58e495482b0c447cfdbd291".`);
         }
     }
+
+    /**
+     * Allows users to pass in either Machine Name (LUIS region or custom <name>.api...) or the full URL
+     * 
+     * @param endpoint https://<region/custom>.api.cognitive.microsoft.com
+     */
+    private getValidEndpoint(endpoint: string): string|null {
```
cleemullins: I can envision an endless series of strings that pass the Regex, but don't work with the pattern. 

My opinion is we should take the full URI endpoint from the dev and just use it "as-is". The parsing game here is a losing battle, and will only cause issues...

mdrichardson: My understanding of @diberry 's email regarding the new endpoint is that it changes from `<region>.api...` to `<custom>.api`, which this still covers.

That being said, if we at least sync up the samples and make it clear what it's asking for, I'm happy. To poorly summarize a discussion with @Kaiqb today, "I've stopped writing LUIS bots because I always forget how to pass the endpoint". We see maybe an issue every week or so regarding this, too.

cleemullins:
Given the impending migration to the LUIS v3 APIs, and the changes to the endpoint format that comes with that, we want to hold off on this at this time.

PR #1088, opened 2019-07-22
Added SkypeMentionNormalizeMiddleware and tests

### LINKED ISSUE #2181: botbuilder-ai depends on moment without using it
The library botbuilder-ai depends on the package moment, which is not even used in the code. There's an import statement in types, but then not references.

[Code Search](https://github.com/microsoft/botbuilder-js/search?q=path%3A%2Flibraries%2Fbotbuilder-ai+moment&unscoped_q=path%3A%2Flibraries%2Fbotbuilder-ai+moment)

tracyboehrer:
Fixes #2181

Added SkypeMentionNormalizeMiddleware to correct the Skype mention text so that TurnContext.removeMentionText operates correctly.

### DIFF
--- a/libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
+++ b/libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
@@ -1,0 +1,40 @@
+/**
+ * @module botbuilder
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import { Activity, ActivityTypes, Middleware, TurnContext, Entity} from 'botbuilder-core';
+
+
+/**
+ * Middleware to patch mention Entities from Skype since they don't conform to expected values.
+ * Bots that interact with Skype should use this middleware if mentions are used.
+ *
+ * @remarks
+ * A Skype mention "text" field is of the format:
+ *   <at id="28:2bc5b54d-5d48-4ff1-bd25-03dcbb5ce918">botname</at>
+ * But Activity.Text doesn't contain those tags and RemoveMentionText can't remove
+ * the entity from Activity.Text.
+ * This will remove the <at> nodes, leaving just the name.
+ */
+export class SkypeMentionNormalizeMiddleware implements Middleware {
+    public static normalizeSkypeMentionText(activity: Activity){
+        if (activity.channelId === 'skype' && activity.type === 'message'){
+            for(var i = 0; i < activity.entities.length; i++){
+                var element = activity.entities[i];
+                if(element.type === 'mention'){
+                    var mentionNameMatch = element['text'].match(/(?<=<at.*>)(.*?)(?=<\/at>)/i);
+                    if (mentionNameMatch.length > 0) {
+                        element['text'] = mentionNameMatch[0];
+                    }
+                }
+            }
+        }
+    }
+
+    public async onTurn(turnContext: TurnContext, next: () => Promise<void>): Promise<void> {
+        SkypeMentionNormalizeMiddleware.normalizeSkypeMentionText(turnContext.activity);
+    }
+}
--- a/libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
+++ b/libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
@@ -20,20 +20,18 @@ import { Activity, Middleware, TurnContext } from 'botbuilder-core';
  * This will remove the <at> nodes, leaving just the name.
  */
 export class SkypeMentionNormalizeMiddleware implements Middleware {
-    public static normalizeSkypeMentionText(activity: Activity){
+    public static normalizeSkypeMentionText(activity: Activity): void {
         if (activity.channelId === 'skype' && activity.type === 'message'){
-            for(var i = 0; i < activity.entities.length; i++){
-                var element = activity.entities[i];
+            activity.entities.map((element): void => {
                 if(element.type === 'mention'){
-                    var text = element['text'];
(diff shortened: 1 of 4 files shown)


tracyboehrer:
These are the same changes made to dotnet.

stevengum:
Let's also keep in mind potential browser-compat issues by continuing to use the super-advanced Node.js RegExp patterns in our SDK: https://github.com/microsoft/botbuilder-js/issues/1075

@compulim, @johnataylor

tracyboehrer:
I'll just remove the use of that regex.  I moved it from the original removeMentionText impl, but it's easy enough to just grab everything between the tags instead.

libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @remarks
+ * A Skype mention "text" field is of the format:
+ *   <at id="28:2bc5b54d-5d48-4ff1-bd25-03dcbb5ce918">botname</at>
+ * But Activity.Text doesn't contain those tags and RemoveMentionText can't remove
+ * the entity from Activity.Text.
+ * This will remove the <at> nodes, leaving just the name.
+ */
+export class SkypeMentionNormalizeMiddleware implements Middleware {
+    public static normalizeSkypeMentionText(activity: Activity){
```
cleemullins: I don't think this code will pass the eslint rules. 

Can you run this through the linter and cleanup the issues it brings up?

libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -114,11 +122,7 @@ export class TurnContext {
         var mentions = TurnContext.getMentions(activity);
         for (var i=0; i<mentions.length; i++) {
             if (mentions[i].mentioned.id === id) {
-                var mentionNameMatch = mentions[i].text.match(/(?<=<at.*>)(.*?)(?=<\/at>)/i);
-                if (mentionNameMatch.length > 0) {
-                    activity.text = activity.text.replace(mentionNameMatch[0], '');
-                    activity.text = activity.text.replace(/<at><\/at>/g, '');
-                }
+                activity.text = activity.text.replace(mentions[i].text, '').trim();
```
cleemullins: I.... don't understand how this works. In C# you've still got a Regex in there. 

Here, I'm probably missing something obvious...

tracyboehrer: I removed RegEx from the Javascript version because stevengum and Scott said that pattern isn't supported on all browsers.  So I am just grabbing everything between the XML tags.  Rather brute force but should work.

libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
stevengum: @jspruance, this method blows up if there is no activity.text: message activities can be sent to the bot that have an attachment and no text. Can you make this more resilient and add more tests?

libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
```diff
@@ -0,0 +1,42 @@
...
+ *   <at id="28:2bc5b54d-5d48-4ff1-bd25-03dcbb5ce918">botname</at>
+ * But Activity.Text doesn't contain those tags and RemoveMentionText can't remove
+ * the entity from Activity.Text.
+ * This will remove the <at> nodes, leaving just the name.
+ */
+export class SkypeMentionNormalizeMiddleware implements Middleware {
+    public static normalizeSkypeMentionText(activity: Activity): void {
+        if (activity.channelId === 'skype' && activity.type === 'message'){
+            activity.entities.map((element): void => {
+                if(element.type === 'mention'){
```
jspruance: I think this is fine if we can't use regex

libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -97,29 +97,32 @@ export class TurnContext {
...
-        var mentions = TurnContext.getMentions(activity);
-        for (var i=0; i<mentions.length; i++) {
-            if (mentions[i].mentioned.id === id) {
-                var mentionNameMatch = mentions[i].text.match(/(?<=<at.*>)(.*?)(?=<\/at>)/i);
-                if (mentionNameMatch.length > 0) {
-                    activity.text = activity.text.replace(mentionNameMatch[0], '');
-                    activity.text = activity.text.replace(/<at><\/at>/g, '');
-                }
-            }
+        const mentions = TurnContext.getMentions(activity);
```
jspruance: I'd like to use 'Array.prototype.find' here but it's not supported by IE

libraries/botbuilder-core/tests/mention.test.js
```diff
@@ -0,0 +1,119 @@
+const assert = require('assert');
+const { MessageFactory, SkypeMentionNormalizeMiddleware, TurnContext } = require('../lib');
+
+describe(`Mention`, function () {
+    this.timeout(5000);
+
```
jspruance: I added 3 negative tests.

jspruance:
Updated to es6 where possible, fixed lint issues, unit tests passing with 3 added negative tests. Ready for re-review.

libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
jspruance: this is done

libraries/botbuilder-core/src/turnContext.ts
jspruance: yep, this should be fine if we can't use regex

PR #1102, opened 2019-08-06
change import in TestAdapter, add @types/node

### LINKED ISSUE #1076: Use `import from` instead of `require` in `testAdapter.ts`
## Versions
What package version of the SDK are you using: `4.5.0`
What nodejs version are you using: N/A
What browser version are you using: IE11

## Describe the bug

In [`testAdapter.ts` line 9](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/testAdapter.ts#L9), it should use `import assert from 'assert';` instead of `import assert = require('assert');`.

## To Reproduce

1. Clone `botbuilder-samples`
1. Go to `samples/javascript_es6/01.browser-echo`
1. Run `npm ci`
1. Run `npm run build`

## Expected behavior

It build successfully.

Instead, Webpack is reporting:

```
SyntaxError: C:\Users\JohnDoe\Source\Repos\BotBuilder-Samples\samples\javascript_es6\01.browser-echo\node_modules\botbuilder-core\src\testAdapter.ts: `import =` is not supported by @babel/plugin-transform-typescript
Please consider using `import <moduleName> from '<moduleName>';` alongside Typescript's --allowSyntheticDefaultImports option.
```

## Additional context

Found this while fixing `samples/javascript_es6/01.browser-echo` sample.

Also, please consider not exposing/publishing `TestAdapter` if it is not used in production code.

[bug]

stevengum:
Fixes #1076 (in [4.5 branch](https://github.com/microsoft/botbuilder-js/tree/4.5))

## Description
- Update import pattern in `TestAdapter`

## Specific Changes
  - Adds `"esModuleInterop": true` to tsconfig.json in botbuilder-core
  - Adds `"@types/node": "^12.6.9"` as `"devDependency"` to botbuilder-core

## Testing
- Locally reran unit tests
- Installed changes via relative-path installation in [browser-echo sample](https://github.com/microsoft/BotBuilder-Samples/tree/master/samples/javascript_es6/01.browser-echo)
@compulim, I ran into this issue after starting the browser sample:

```bash
Uncaught TypeError: Class constructor BotAdapter cannot be invoked without 'new'
    at new WebChatAdapter (webChatAdapter.ts:14)
    at Module../src/app.ts (app.ts:16)
    at __webpack_require__ (bootstrap:726)
    at fn (bootstrap:100)
    at Object.0 (app.js:191399)
    at __webpack_require__ (bootstrap:726)
    at bootstrap:793
    at bootstrap:793
```

Have you run into this issue before?

image


stevengum:
@compulim thinks this is due to the output of `tsc` when we transpile and target [`"ESNext"`](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/tsconfig.json#L3)  instead of something like [`"ES5"` or `"ES6"`](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

compulim:
Thanks @stevengum, let me elaborate a bit more on the topic.

- BotBuilder-JS package is published as ESNext
- When customer build their page, they use Webpack
   - By default, Webpack is configured to only transpile JS code inside userland, not under `/node_modules/`
   - Thus, Webpack is bundling the BotBuilder-JS inside the mono JS file as-is, i.e. in ESNext
   - Webpack is also bundling user code inside the mono JS file, transpiled
- Then, in ES `class`, there is one restriction
   - Do not call a class object, you can only call it thru `new YourClass()`
   - User code get transpiled. In ES5, where we don't have `class` at that ages, the transpiled code turns a `new YourClass()` into a simple function call (remember the old day of `YourClass.prototype.constructor`?)
   - BotBuilder-JS code is not transpiled, the class is still a class, so it get called like a function

I think we should not publish BotBuilder-JS as ESNext. The de facto standard today:

- NPM published code (entrypoint-ed by `package.json/main`), should be ES5, usually kept under `/lib/` folder
- You are free to include source code, non-transpiled, usually kept under `/src/` folder

stevengum:
@compulim, is changing the output of `tsc` a breaking change?

cleemullins:
Is this still a PR we need to cherrypick to 4.5?

stevengum:
note to self, make the changes against master, and if @benbrown, @cwhitten, @a-b-r-o-w-n could chime in later that would be great.

a-b-r-o-w-n:
@stevengum I think that the strategy @compulim outlined is typical among js libs. In terms of if it is a breaking change... maybe. If there is good test coverage, I would say the change could go with a minor version.

stevengum:
Closing in favor of #1232 which should pick up where this PR left off.

PR #1106, opened 2019-08-09
NumberPrompt: Added parsing that takes culture into consideration using Globalize

### LINKED ISSUE #1086: [parity] NumberPrompt doesn't take culture into account while parsing value
## Versions
v4.5

## Describe the bug
See equivalent [botbuilder-dotnet issue #2288](https://github.com/microsoft/botbuilder-dotnet/issues/2288) for details

**Test**
```js
    it('should consider culture when parsing.', async function () {
        // Initialize Testawait adapter.
        const adapter = new TestAdapter(async (turnContext) => {
            const dc = await dialogs.createContext(turnContext);

            const results = await dc.continueDialog();
            if (results.status === DialogTurnStatus.empty) {
                await dc.prompt('prompt', 'Please send a number.');
            } else if (results.status === DialogTurnStatus.complete) {
                const reply = results.result.toString();
                await turnContext.sendActivity(reply);
            }
            await convoState.saveChanges(turnContext);
        });
        // Create new ConversationState with MemoryStorage and register the state as middleware.
        const convoState = new ConversationState(new MemoryStorage());

        // Create a DialogState property, DialogSet and NumberPrompt.
        const dialogState = convoState.createProperty('dialogState');
        const dialogs = new DialogSet(dialogState);
        dialogs.add(new NumberPrompt('prompt', null, 'nl-nl'));

        await adapter.send('Hello')
            .assertReply('Please send a number.')
            .send('3,14')
            .assertReply('3,14');
    });
```

image

[bug]

Zerryth:
Fixes #1086 

## Description
* Numbers are parsed from string result returned from recognizers-text with taking culture into account now (previously it never parsed with internationalization)

## Specific Changes
  -  Used open source library, Globalize, to adding parsing with internationalization
  - Added tests to ensure culture is considered in scenarios when specified in:
    1. Activity
    2. Default Locale via `NumberPrompt` constructor
    3. Uses '`en-us`' as default locale if none is specified
  - Edited tsconfig in dialogs to allow import of json required for use of Globalize


cleemullins:
@stevengum Can you review this? Tomorrow is the deadline, and if it's not in by then it'll probably never go in .

Zerryth:
use full locale

Zerryth:
Update again:

In PR you'll note that we only use first 2 letters for locale identifier instead of the full string value that we have with locales provided via recognizers-text. For example:

* recognizers-text culture: `en-us`
* value we use in `NumberPrompts`: `en`

This is because using globalize to parse numbers from strings, the regions that we support are represented without the regional suffix ("-us"); if we try to use the regional suffix, it'll break.

Therefore in the code, we have a helper function formatting to only use first 2 char of locale identifier.

See [Globalize Locales ](https://github.com/globalizejs/globalize#locales) for more info regarding their identifiers.

___

Per Steven Gum, making note of this in PR and I have a short comment helper function

libraries/botbuilder-dialogs/tests/numberPrompt.test.js
```diff
@@ -1,4 +1,5 @@
 const { ConversationState, MemoryStorage, TestAdapter } = require('botbuilder-core');
+const { ActivityTypes } = require('botframework-schema');
```
stevengum: This should come through botbuilder-core because the schema is exposed through this library.

If someone forked the repo and then only installed the dependencies for botbuilder-dialogs and then ran the tests, these tests would fail on an import error.

libraries/botbuilder-dialogs/tests/numberPrompt.test.js
```diff
@@ -249,4 +250,92 @@ describe('NumberPrompt', function () {
             .send('0')
             .assertReply('ok')
     });
+
+    it('should consider culture specified in constructor', async function () {
+        // Initialize Testawait adapter.
```
stevengum: Testawait adapter?

Zerryth: Hmmm...definitely copied the first unit test as a scaffoled for other tests.

Removed the offending comments!

stevengum:
@EricDahlvang (and @cleemullins and @johnataylor) looked at this and decided that it shouldn't be taken into the core product. If we want to have a sample that shows how to do this, by all means. 

We should have a sample that handles inbound and outbound locales, and open a bug on the recognizers to fix the underlying issue.

In the 4.7 timeframe, MSLG should land and that includes a more robust locale story for language generation.

PR #1107, opened 2019-08-12
[Do Not Merge][BotBuilder-AI] Add custom Luis client

### LINKED ISSUE #1071: [Botbuilder-AI] Remove @azure-ms-rest dependency
## Issue description
**Botbuilder-ai** has a pinned version of `@azure/ms-rest` [v1.8.13](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/package.json#L24). This version is being used to create a credential that’s used by the LUIS client. 
Right now, the versions are compatibles, but some time ago we had an issue because the LUIS client upgraded its version of `@azure/ms-rest`, which caused some differences between the interfaces.
To avoid this happening again, we want to eliminate the `@azure/ms-rest` dependency. However, we can’t do this because of the coupling between LUIS client and `@azure/ms-rest`. 

Currently, the [LuisRecognizer](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/luisRecognizer.ts) class depends on `@azure-ms-rest` to create a [LUISRuntimeClient](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/cognitiveservices/cognitiveservices-luis-runtime/src/lUISRuntimeClient.ts) object.

```
const creds: msRest.TokenCredentials = new msRest.TokenCredentials(this.application.endpointKey);
const baseUri: string = this.application.endpoint || 'https://westus.api.cognitive.microsoft.com';
this.luisClient = new LuisClient(creds as any, baseUri);
```
This implementation caused build issues. The solution was to cast the credentials as `any` to avoid the build to break.

## Possible solution
- Create a LUIS client manually that does not depend on `@azure/ms-rest`.

Bill7zz:
Fixes #1071 

## Description
We created a custom Luis client that does not depend on `@azure/ms-rest`. We used [OpenApi ](https://openapi-generator.tech/)Generator to create a base client from using the Luis API definition. Then we modified it to make it compliant with the current implementation of Luis Recognizer

## Specific Changes
- Create a custom Luis runtime client
- Replace the `@azure/congnitiveservice-luis-runtime` client with a custom Luis runtime client
- Remove the `@azure` dependencies from the BotBuilder-AI library
- Update the _LuisRecognizer_ and _LuisSdk_ tests to make them pass using the custom Luis client

## Testing
The next image shows the result of the local tests.
image

Also, this one shows the result of running the test in a custom build pipeline. 
image

Additionally, we tested web-compatibility by doing a webpack of the new client and using it in a browser bot. We used the [browser](https://github.com/microsoft/BotBuilder-Samples/tree/master/samples/javascript_es6/01.browser-echo) echo bot project (since it's a bot that can be deployed to web and run on a browser). 

The next image shows the result of testing the bot that uses the library
image

## Conclusion

The library is compatible with the browsers: 
    - Google Chrome, 
    - Mozilla Firefox 
    - and Microsoft Edge (Beta based on Chromium)

Microsoft Edge native is not compatible because some latest features of JavaScript are not supported.
The **Recognizer** class is using **Spread syntax in object literals** ([here](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/luisRecognizer.ts#L189-L196)), which is a [feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Browser_compatibility) implemented in ECMAScript 2018.


libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,289 @@
...
+            this.basePath = basePath + luisVersion;
+        }
+    }
+
+    public set useQuerystring(value: boolean) {
+        this._useQuerystring = value;
+    }
+
+    public set basePath(basePath: string) {
+        this._basePath = basePath;
```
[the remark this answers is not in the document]
Bill7zz: @stevengum feedback applied.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+/**
+ * NOTE: This class was auto generated by OpenAPI Generator (https://openapi-generator.tech)
+ * and was manually modified to make it compliant with the current implementation of the library.
+ */
+
+import Request = require('request');
```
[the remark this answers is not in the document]
Bill7zz: Thanks, We will be reviewing this approach and once it is implemented we will notify you.

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -287,31 +289,34 @@ export class LuisRecognizer implements LuisRecognizerTelemetryClient {
...
+                    utterance,
+                    this.application.applicationId,
                     {
                         verbose: luisPredictionOptions.includeAllIntents,
                         customHeaders: {
-                            'Ocp-Apim-Subscription-Key': this.application.endpointKey,
-                            'User-Agent': this.getUserAgent()
+                            headers:{                        
+                                'User-Agent': this.getUserAgent(),
+                                'authorization': `Bearer ${this.application.endpointKey}`,
```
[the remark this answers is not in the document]
Bill7zz: In this case, the mentioned header is set [here](
https://github.com/microsoft/botbuilder-js/blob/f58d025352ded2ab2235e6d9292d76edde3a4c12/libraries/botbuilder-ai/src/luis-client/luisClient.ts#L63).

The next image shows a bot running with those changes in which you can see the header value before making the request. 
image

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,277 @@
...
+    customHeaders: {headers: {[name: string]: string}};
+}
+
+export enum LuisApikeys {
+    apiKeyHeader,
+}
+
+export class LuisClient {
+    protected _basePath: string = '';
+    protected _useQuerystring: boolean = false;
```
stevengum: This defaults to false, is not exposed in the constructor and has no docstring; what's the story around this member for users?

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,277 @@
...
+    bingSpellCheckSubscriptionKey?: string;
+    /**
+     * Log query (default is true)
+     */
+    log?: boolean;
+
+    customHeaders: {headers: {[name: string]: string}};
+}
+
+export enum LuisApikeys {
```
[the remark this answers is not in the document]
GasparAcevedoZainSouthworks: @stevengum We added the necessary comment for this enum.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,277 @@
...
+    apiKeyHeader,
+}
+
+export class LuisClient {
```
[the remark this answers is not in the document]
GasparAcevedoZainSouthworks: Class documented.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,277 @@
...
+    protected _basePath: string = '';
```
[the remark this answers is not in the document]
GasparAcevedoZainSouthworks: This member was auto-generated as `protected`. We change it to `private` and added the necessary comments.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,277 @@
...
+        });
+    }
+    /**
+     * Gets predictions for a given utterance, in the form of intents and entities. The current maximum query size is 500 characters.
+     * @param appId The LUIS application ID (guid).
+     * @param query The utterance to predict.
+     * @param [options] The optional parameters
+     * @returns Promise<LuisResult>
+     */
+    public async predictionResolveGet(appId: string, query: string, options: PredictionResolveOptionalParams): Promise<LuisResult> {
```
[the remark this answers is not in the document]
GasparAcevedoZainSouthworks: Method remove.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
GasparAcevedoZainSouthworks: This is another auto-generated member. As this wasn't reference in any other class (nor in the .LuisClient), we just removed it.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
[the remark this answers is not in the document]
GasparAcevedoZainSouthworks: @stevengum @Bill7zz We replaced the `require` library with `node-fetch` in the `luisClient.tst` file.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,277 @@
...
+        return new Promise<LuisResult>((resolve, reject) => {
+            Request(localRequestOptions, (error, response, luisResult) => {
+                if (error) {
+                    reject(error);
+                } else {
+                    luisResult = ObjectSerializer.deserialize(luisResult, 'LuisResult');
+                    if (response.statusCode && response.statusCode >= HttpStatus.OK && response.statusCode < HttpStatus.MULTIPLE_CHOICES) {
+                        resolve(luisResult);
+                    } else {
+                        reject({
```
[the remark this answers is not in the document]
GasparAcevedoZainSouthworks: We'll be applying these changes in the next days. Once they are applied, we'll notify you.

GasparAcevedoZainSouthworks:
> What's the ETA for completing and merging this?

Hi @cleemullins. 
As this PR has the coverage over 80%, we consider it's ready to merge, unless you have any more feedback.

Bill7zz:
Hi @cleemullins, @stevengum

we tested the new client by using the browser echo bot. We configured the webpack.config file to include the directory `node_modules/botbuilder-ai` to the transpile process.
 
The next image shows the code included in the webpack.config
image

The next images show the bot running on the browsers chosen for the test.
image
image

As you can see, the library and the new LUIS client are browser compatible.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,200 @@
...
+
+const fetch = (new Function('require', 'if (!this.hasOwnProperty("fetch")) { return require("node-fetch"); } else { return this.fetch; }'))(require);
+
+import * as HttpStatus from 'http-status-codes';
+
+/* tslint:disable:no-unused-locals */
+import { LuisResult } from './model/luisResult';
+
+import { ObjectSerializer, Authentication, VoidAuth } from './model/models';
+import { ApiKeyAuth } from './model/models';
```
stevengum: This can be on one line...

libraries/botbuilder-ai/src/luis-client/model/models.ts
```diff
@@ -0,0 +1,215 @@
...
+    public password: string = '';
+
+    applyToRequest(requestOptions: any): void {
+        requestOptions.auth = {
+            username: this.username, password: this.password
+        }
+    }
+}
+
+export class ApiKeyAuth implements Authentication {
```
stevengum: It looks like this is the only class (with `VoidAuth`) that we currently use. Can you get rid of the HttpBasicAuth and OAuth authentication classes?

libraries/botbuilder-ai/src/luis-client/model/models.ts
```diff
@@ -0,0 +1,215 @@
...
+ * and was manually modified to make it compliant with the current implementation of the library.
+ */
+
+export * from './aPIError';
+export * from './compositeChildModel';
+export * from './compositeEntityModel';
+export * from './entityModel';
+export * from './intentModel';
+export * from './luisResult';
+export * from './sentiment';
```
stevengum: Why are these being exported from this file? What's the difference from [`model/index.ts`](https://github.com/microsoft/botbuilder-js/pull/1107/files#diff-0432c52ed06478492a158582768cefb5R13-R19)?

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,200 @@
...
+
+let luisVersion = '/luis/v2.0';
```
stevengum: This should be a `const` for now; it might change in the future but otherwise it should be a `const`

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,200 @@
...
+    /** Supported Cognitive Services endpoint (protocol and hostname, for example: https://westus.api.cognitive.microsoft.com) */
+    private _basePath: string = '';
+
+    protected authentications = {
+        'default': new VoidAuth() as Authentication,
+        'apiKeyHeader': new ApiKeyAuth('header', 'Ocp-Apim-Subscription-Key'),
+    }
+
+    /** Creates a new instance of LuisClient.
+     * @param basePath Supported Cognitive Services endpoint.
```
stevengum: Can you add an example of a valid `basePath` value to this doc string?

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,200 @@
...
+     */
+    public constructor(basePath: string){
+        if (basePath) {
+            this._basePath = basePath + luisVersion;
+        }
+    }
+
+    public setDefaultAuthentication(auth: Authentication): void {
```
stevengum: There are no tests for this method, is this ever used?

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Luis Client Class
+ *
+ * @remarks
+ * This is a helper class for all the http request operations against the LUIS API.
+ * https://www.luis.ai/
+ */
+export class LuisClient {
+    /** Supported Cognitive Services endpoint (protocol and hostname, for example: https://westus.api.cognitive.microsoft.com) */
```
stevengum: This string should move down to the constructor, and is also inaccurate when you look at the constructor. The constructor for the `LuisClient` takes the `basePath` param and adds `luisVersion` to it.

The example is valid for `basePath`, but not for `_basePath`.

libraries/botbuilder-ai/src/luis-client/model/models.ts
```diff
@@ -0,0 +1,215 @@
...
+    'APIError': APIError,
+    'CompositeChildModel': CompositeChildModel,
+    'CompositeEntityModel': CompositeEntityModel,
+    'EntityModel': EntityModel,
+    'IntentModel': IntentModel,
+    'LuisResult': LuisResult,
+    'Sentiment': Sentiment,
+}
+
+export class ObjectSerializer {
```
stevengum: We need tests for this class as well; it looks like it's [exposed](https://github.com/microsoft/botbuilder-js/pull/1107/files#diff-583a03304a5e52260ab6e44f50ecd0c5R1-R14).

libraries/botbuilder-ai/src/luis-client/luisClient.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /** Creates a new instance of LuisClient.
+     * @param basePath Supported Cognitive Services endpoint.
+     */
+    public constructor(basePath: string){
+        if (basePath) {
+            this._basePath = basePath + luisVersion;
+        }
+    }
```
stevengum: This logic doesn't quite support custom host names and endpoints. @chrimc62 what's the proper logic to use here?

libraries/botbuilder-ai/src/luisRecognizer.ts
```diff
@@ -196,13 +201,10 @@ export class LuisRecognizer implements LuisRecognizerTelemetryClient {
...
-        // - We have to cast "creds as any" to avoid a build break relating to different versions
-        //   of autorest being used by our various components.  This is just a build issue and
-        //   shouldn't effect production bots.
-        const creds: msRest.TokenCredentials = new msRest.TokenCredentials(this.application.endpointKey);
-        const baseUri: string = this.application.endpoint || 'https://westus.api.cognitive.microsoft.com';
-        this.luisClient = new LuisClient(creds as any, baseUri);
+        // Create Luis Runtime Client
+        const basePath = 'https://westus.api.cognitive.microsoft.com';
+        this.luisClient = new LuisClient(basePath);
+        this.luisClient.setApiKey(LuisApikeys.apiKeyHeader, this.application.endpointKey)
```
stevengum: This looks like a horrendous bug.... We're not only supporting the `"westus"` region, are we? 😄 

Perhaps some other tests hitting different endpoints are required...

stevengum:
@darrenj FYI, when this PR gets merged it will likely cause some TypeScript projects to not build in your [repo](https://github.com/microsoft/botframework-solutions/search?l=TypeScript&q=botbuilder-ai)...

libraries/botbuilder-ai/src/luis-client/luisClient.ts
Aliandi: We did a little refactor here and now the exports come from `model/index.ts`. This is just one line now.

libraries/botbuilder-ai/src/luis-client/model/models.ts
Aliandi: Indeed, we deleted them!

libraries/botbuilder-ai/src/luis-client/model/models.ts
Aliandi: There's no difference actually. We are keeping index for exporting and adding a new helper class to handle the ObjectSerializer class and a types file to handle types.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
Aliandi: Done!

libraries/botbuilder-ai/src/luis-client/model/models.ts
Aliandi: Now it's not, it shouldn't have been exposed

libraries/botbuilder-ai/src/luis-client/luisClient.ts
Aliandi: We'll add this ASAP.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
Aliandi: It's not, we are removing it.

libraries/botbuilder-ai/src/luis-client/luisClient.ts
Aliandi: We'll add this ASAP.

libraries/botbuilder-ai/src/luisRecognizer.ts
Aliandi: We totally missed this, nice catch. We'll take a look into this ASAP.

Aliandi: Fixed this to the [original code](https://github.com/microsoft/botbuilder-js/blob/fa7318bf9be24164b990a2afc28b3accce130121/libraries/botbuilder-ai/src/luisRecognizer.ts#L204)

Aliandi:
> 
> 
> @Bill7zz what does the upgrade story look like for users who use TypeScript and an old version of the `botbuilder-ai` package? I'm assuming they have to rewrite the components that use the LUIS model types?

@stevengum, we'd expect that not to be the case, we are running some tests right now with the Virtual Assistant's team and will let you know. We'll try to have no breaking changes or as little as possible, properly documenting them.

cleemullins:
We need to revisit this as part of the V3 LUIS Recognizer that hits the endpoints direction using standard REST calls.

PR #1112, opened 2019-08-16
updated docstring in LuisRecognizer to newly reorg'd dispatch sample

### LINKED ISSUE #779: Update LuisRecognizer description after sample reorganization
## Tracking Issue:
Update LuisRecognizer description after sample reorganization efforts. 

The LuisRecognizer currently points to the 12.nlp-with-luis sample which may be renumbered in the upcoming sample work. If it is renumbered, we will need to change the link in the [doc string](https://github.com/Microsoft/botbuilder-js/commit/6acf62fe67533738712b35a155be43626a79741b#diff-d746df635d42aa7e8228fc3af83d6852R105):

```ts
/**
 * Recognize intents in a user utterance using a configured LUIS model.
 *
 * @remarks
 * This class is used to recognize intents and extract entities from incoming messages.
 * See this class in action [in this sample application](https://github.com/Microsoft/BotBuilder-Samples/tree/master/samples/javascript_nodejs/12.nlp-with-luis).
 *
 * This component can be used within your bots logic by calling [recognize()](#recognize).
 */
export class LuisRecognizer {
```

[Link to code in master](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/luisRecognizer.ts#L103-L112)

Zerryth:
Fixes #779 

## Description
Updates link in `LuisRecognizer` docstring to updated sample, `14.nlp-with-dispatch`.

### DIFF
--- a/libraries/botbuilder-ai/src/luisRecognizer.ts
+++ b/libraries/botbuilder-ai/src/luisRecognizer.ts
@@ -142,7 +142,7 @@ export interface LuisRecognizerTelemetryClient {
  *
  * @remarks
  * This class is used to recognize intents and extract entities from incoming messages.
- * See this class in action [in this sample application](https://github.com/Microsoft/BotBuilder-Samples/tree/master/samples/javascript_nodejs/12.nlp-with-luis).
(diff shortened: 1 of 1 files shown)


CoHealer:
Bump on this review; issue closing pending

PR #1114, opened 2019-08-19
Add Resource Response Id to cloned activities when logging

### LINKED ISSUE #1101: transcriptLogger should update activity.Id on the cloned activity
## Describe the bug

https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/transcriptLogger.ts#L52

```js
context.onSendActivities(async (ctx: TurnContext, activities: Partial<Activity>[], next2: () => Promise<ResourceResponse[]>) => {
// run full pipeline
const responses: ResourceResponse[] = await next2();
activities.forEach((a: ResourceResponse) => this.logActivity(transcript, this.cloneActivity(a)));

return responses;
});
```
The cloned activity in the code above should include the ResourceResponse.id retrieved from sending the activity to the server.  Currently, all outgoing activities logged will NOT have an activity.id

## Expected behavior
Activities logged should contain the actual id of the activity (which is determined by the server, when sending the message). 

[bug]

EricDahlvang:
Fixes #1101

## Description
When sending an activity, a resource response id is returned from the service.  This PR adds that id to the activity being logged (if one is not present).

## Testing
Added a unit test, and tested manually with a local bot.

### DIFF
--- a/libraries/botbuilder-core/src/transcriptLogger.ts
+++ b/libraries/botbuilder-core/src/transcriptLogger.ts
@@ -49,7 +49,21 @@ export class TranscriptLoggerMiddleware implements Middleware {
         context.onSendActivities(async (ctx: TurnContext, activities: Partial<Activity>[], next2: () => Promise<ResourceResponse[]>) => {
             // run full pipeline
             const responses: ResourceResponse[] = await next2();
-            activities.forEach((a: ResourceResponse) => this.logActivity(transcript, this.cloneActivity(a)));
+
+            let responseIndex = 0;
+            activities.forEach((a: Partial<Activity>) => {
+                let clonedActivity: Activity = this.cloneActivity(a);
+                if (responses) {
+                    // if there is no id on the activity, add the id returned from the service
+                    if (responseIndex < responses.length) {
+                        if (!clonedActivity.id) {
+                            clonedActivity.id = responses[responseIndex].id;
+                        }
+                    }
+                    responseIndex++;
+                }
+                this.logActivity(transcript, clonedActivity);
+            });
 
             return responses;
         });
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-core/src/transcriptLogger.ts
```diff
@@ -49,7 +49,21 @@ export class TranscriptLoggerMiddleware implements Middleware {
         context.onSendActivities(async (ctx: TurnContext, activities: Partial<Activity>[], next2: () => Promise<ResourceResponse[]>) => {
             // run full pipeline
             const responses: ResourceResponse[] = await next2();
-            activities.forEach((a: ResourceResponse) => this.logActivity(transcript, this.cloneActivity(a)));
+
+            let responseIndex = 0;
+            activities.forEach((a: Partial<Activity>) => {
+                let clonedActivity: Activity = this.cloneActivity(a);
+                if (responses) {
+                    // if there is no id on the activity, add the id returned from the service
```
jspruance: can you use map() here rather than incrementing a counter?

libraries/botbuilder-core/tests/transcriptMiddleware.test.js
```diff
@@ -196,6 +196,54 @@ describe(`TranscriptLoggerMiddleware`, function () {
             adapter.send('test');
         });
     });
+
+    it(`should add resource response id to activity when activity id is empty`, function (done) {
+        var conversationId = null;
```
jspruance: small thing but, should be 'const' in not reassigned, otherwise 'let'

EricDahlvang: fixed

stevengum:
@cleemullins @benbrown, @stevenic can one of you chime in on this PR? 

This PR a bug where the TranscriptLoggerMiddleware would blow up when trying to log the activity because of a missing `activity.id`.

The PR fixes this by logging cloned activities that have been updated with the `resourceResponse.id`.  

The original behavior is to log the original outgoing activities which may or may not have an `id` provided by the user. This would blow up in FileTranscriptStore, but does not in AzureBlobTranscriptStore, which then uses "undefined" or "null" as the `activity.id` when constructing the blob.

PR #1121, opened 2019-08-27
Use Recognizers Cultures in Choice & Confirm Prompt

mdrichardson:
## Description

Currently, Confirm and Choice Prompts use an object with all-lowercase keys for localization:

```js
{
    'es-es': { inlineSeparator: ', ', inlineOr: ' o ', inlineOrMore: ', o ', includeNumbers: true },
    [...]
};
```

The locale used for the prompt is determined with something like this:

```js
ChoicePrompt.defaultChoiceOptions[locale]
```

If the `locale` is not all-lowercase, it defaults to `en-us`. This can be problematic because:

1. [`activity.locale` uses ISO 3166](https://github.com/microsoft/botbuilder-js/blob/5b8b7efef63d31dc32fe99050aa15e42378e832d/libraries/botframework-schema/src/activityInterfaces.ts#L115), which I believe [is normally recognized as all-caps](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
2. [WebChat sends the capitalized country-code version](https://github.com/microsoft/BotFramework-WebChat/blob/a10096a5d769240da04d9c3fd4a57bcb384cbed4/packages/embed/src/locale.js#L26)

So, if an activity comes in with `locale: "fr-FR"` instead of `locale: "fr-fr"`, they're going to get an `en-us` prompt. Tested to confirm.

This allows for locales like:

* `fr-fr`
* `fr-FR`
* `FR`
* `fr`

to all get mapped to the appropriate `fr-fr`

Thanks to @stevkan for the assist!

## Specific Changes

I could have just done `.toLowerCase()` in each prompt, but instead, kept it more in line with [the DotNet versions](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.Dialogs/Prompts/ConfirmPrompt.cs) (*Note: [`recognizers-text` essentially calls `toLowerCase()` as well](https://github.com/microsoft/Recognizers-Text/blob/master/JavaScript/packages/recognizers-text/src/culture.ts#L41)*):

**C#**

```csharp
{ Spanish, (new Choice("Sí"), new Choice("No"), new ChoiceFactoryOptions(", ", " o ", ", o ", true)) },
```

by using:

**Node**

```js
import { Culture } from '@microsoft/recognizers-text';
[...]
[Culture.Spanish]: { choices: ['Sí', 'No'], options: { inlineSeparator: ', ', inlineOr: ' o ', inlineOrMore: ', o ', includeNumbers: true }},
```

## Testing

Current tests [DO test for locale](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-dialogs/tests/confirmPrompt.test.js#L245) and all current tests pass:

image

### DIFF
--- a/libraries/botbuilder-ai/src/luisRecognizer.ts
+++ b/libraries/botbuilder-ai/src/luisRecognizer.ts
@@ -142,7 +142,7 @@ export interface LuisRecognizerTelemetryClient {
  *
  * @remarks
  * This class is used to recognize intents and extract entities from incoming messages.
- * See this class in action [in this sample application](https://github.com/Microsoft/BotBuilder-Samples/tree/master/samples/javascript_nodejs/12.nlp-with-luis).
+ * See this class in action [in this sample application](https://github.com/microsoft/BotBuilder-Samples/tree/master/samples/javascript_nodejs/14.nlp-with-dispatch).
  *
  * This component can be used within your bots logic by calling [recognize()](#recognize).
  */
--- a/libraries/botbuilder-dialogs/src/prompts/choicePrompt.ts
+++ b/libraries/botbuilder-dialogs/src/prompts/choicePrompt.ts
@@ -71,7 +71,7 @@ export class ChoicePrompt extends Prompt<FoundChoice> {
 
     protected async onPrompt(context: TurnContext, state: any, options: PromptOptions, isRetry: boolean): Promise<void> {
         // Determine locale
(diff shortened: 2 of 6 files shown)


mdrichardson:
@stevengum I big time added those tests. Not sure how a couple of those unrelated changes snuck in there, though. Let me know if I need to remove them somehow.

PR #1138, opened 2019-09-03
use timestamp or processing time as fallback for activity.id

### LINKED ISSUE #1122: Populate Activity.Id in Logged Activities when undefined or no ResourceResponse.Id is returned
## Versions
`master`

## Describe the bug
Certain scenarios occur when sending activities to Teams where there is no ResourceResponse.Id in the ResourceResponse sent back to the bot. This means that the TranscriptLoggerMiddleware (and more precisely the stores it uses) runs into issues when there is no Activity.Id to be logged. (see microsoft/botbuilder-js#1114)

 - `FileTranscriptStore` blows up because it uses the activity.id as the filename
 - `BlobTranscriptStore` instead uses `undefined` instead of the activity.id when calculating the blob's id.

[bug]

stevengum:
Fixes #1122

## Description
Use outgoing `activity.timestamp` or current processing time in `TranscriptLoggerMiddleware` to populate `id` field in activities being logged by the middleware. In some cases, ResourceResponses from Teams does not have a `id` value.

## Testing
Unit tests added.


libraries/botbuilder-core/src/transcriptLogger.ts
```diff
@@ -55,6 +55,18 @@ export class TranscriptLoggerMiddleware implements Middleware {
...
+                        // This fix uses the timestamp on the activity to populate its id for logging the transcript.
+                        // If there is no outgoing timestamp, the current time for the bot is used for the activity.id.
+                        // See https://github.com/microsoft/botbuilder-js/issues/1122
+                        if (!clonedActivity.id) {
+                            if (clonedActivity.timestamp) {
+                                clonedActivity.id = new Date(clonedActivity.timestamp).getTime().toString();
+                            } else {
+                                clonedActivity.id = Date.now().toString();
+                            }
+                        }
```
cleemullins: This isn't unique. Can you use the random+datetime to get a unique value?

stevengum:
I'm considering generating a fake activity.id BotFrameworkAdapter.sendActivities for InvokeResponses. This change would need to be picked up in C# and would solve the original problem this PR was attempting to address.

PR #1164, opened 2019-09-11
Fix 1 test

BruceHaley:
For build BotBuilder-JS-4.future-daily.
Fix dialogContext.test.js, DialogContext, "should bind input & output state to a dialog."
     TypeError: Cannot set property 'value' of undefined.


cleemullins:
I don't think we (you and I) should be on the hook for fixing these builds. The owners of this codebase need to step up and own these issues.

PR #1190, opened 2019-09-20
Add support for structured LG template

Danieladu:
DCR: https://github.com/microsoft/botbuilder-dotnet/issues/2325


Danieladu:
re send a draft pr

Issue #1218, opened 2019-09-24
[Teams] Find a home for the interfaces that extend msrest interfaces


stevengum:
Make sure the `TeamsConnectorClientOptions` and its `ServiceClientOptions` dependency from `@azure/ms-rest-js` get placed into the proper location.

The [`TeamsConnectorClientOptions`](https://github.com/OfficeDev/BotBuilder-MicrosoftTeams-node/blob/master/botbuilder-teams-js/src/schema/models/index.ts#L1220-L1230) was originally found in the schema folder in `botbuilder-teams`, but we don't want to add the ms-rest-js dependency to `botframework-schema`. The `HttpResponse` interface from msrest is also used.

As such I've removed the interfaces from botframework-schema and not included the dependency.

**We may move the these interfaces into the `botbuilder` library.**

```ts
import { HttpResponse, ServiceClientOptions } from '@azure/ms-rest-js';

// ..

/**
 * @interface
 * An interface representing TeamsConnectorClientOptions.
 * @extends ServiceClientOptions
 */
export interface TeamsConnectorClientOptions extends ServiceClientOptions {
  /**
   * @member {string} [baseUri]
   */
  baseUri?: string;
}

// ...

/**
 * Contains response data for the fetchChannelList operation.
 */
export type TeamsFetchChannelListResponse = ConversationList & {
  /**
   * The underlying HTTP response.
   */
  _response: HttpResponse & {
      /**
       * The response body as text (string format)
       */
      bodyAsText: string;
      /**
       * The response body as parsed JSON or XML
       */
      parsedBody: ConversationList;
    };
};

/**
 * Contains response data for the fetchTeamDetails operation.
 */
export type TeamsFetchTeamDetailsResponse = TeamDetails & {
  /**
   * The underlying HTTP response.
   */
  _response: HttpResponse & {
      /**
       * The response body as text (string format)
       */
      bodyAsText: string;
      /**
       * The response body as parsed JSON or XML
       */
      parsedBody: TeamDetails;
    };
};
```

cleemullins:
This isn't a deliverable. Closing.

stevengum:
Done: This was a deliverable and was completed in #1249

PR #1220, opened 2019-09-25
[Teams] Add Teams schemas and initial TeamsActivityHandler

stevengum:
_Partial progress for Phase 1 in #1214_

In the interest of preserving everyone's sanity I'm doing this initial PR so the reviews are not grossly tedious, just partly tedious.

### Start of Teams work in main JS repo around schemas and TeamsActivityHandler. 

Some important differences to note about using the TeamsActivityHandler; for some `onTeams*` methods, they **expect** a return value. This affects the handlers for type "invoke" activities. These return values are sent on the user's behalf to the Teams channel. 

Since a return value is expected, only **one** of each type of `onTeams*` invoke processing handlers can be registered. Attempts to register more than one type of a handler will result in a thrown Error.

Additionally, Teams must receive an OK ([HTTP status code 200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)) InvokeResponse from the bot within 5 seconds for these Teams invoke activities (another reason to not register more than one handler for each type).

If a bot receives a Teams Invoke activity and does not handle it, the bot will send a Not Implemented ([HTTP status code 501](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501)) InvokeResponse. This design is because Teams cannot send these invoke activities to the bot unless the behavior is defined in the Teams App Manifest, which is created by the bot developer.

## Specific Changes
  - Adds the Teams schemas to the `botframework-schemas` library.
    - No new dependencies (see #1218)
  - Adds initial TeamsActivityHandler and unit tests
    - Mirror C# in naming pattern (except for `*async`)
    - Mirror C# in Tightening up return values

## Testing
Initial unit tests added for TeamsActivityHandler.


libraries/botbuilder/src/teamsActivityHandler.ts
```diff
@@ -0,0 +1,498 @@
...
+                            await runDialogs();
+                        }
+                    });
+                    break;
+                case ActivityTypes.Invoke:
+                    // The onInvokeActivity handler should not call `runDialogs`, it needs to let the
+                    // typed-Invoke Activity handler call `runDialogs`.
+                    await this.handle(context, 'Invoke', async () => { });
+                    let invokeResponse: InvokeResponse;
+                    switch (context.activity.name) {
```
johnataylor: possibly a new function

libraries/botbuilder/src/teamsActivityHandler.ts
```diff
@@ -0,0 +1,498 @@
...
+                        case 'fileConsent/invoke':
+                            await this.handle(context, 'TeamsFileConsent', async () => { });
```
johnataylor: can we make the default behavior run the child events... i.e. make the "on" an assignment rather than an add to a list

PR #1223, opened 2019-09-27
Create ActivityHandlerBase, rewrite ActivityHandler for increased extensibility points

stevengum:
## Description
This PR adds the ActivityHandlerBase with new `on*Activity()` methods. These methods mirror the corresponding C# activities. 

It also refactors the core logic in `ActivityHandler` and `ActivityHandler.run()` out into separate overrideable methods. This increases extensibility as developers can change the processing logic and reuse the base logic.

The original eventing logic is still included in `ActivityHandler`. 

None of the already public APIs in `ActivityHandler` have been modified, and `ActivityHandler.run()` is still the only point of ingesting activities.

## Specific Changes
 - New `ActivityHandlerBase` class used for subclassing. (Closely follows the C# impl of `ActivityHandler`)
   - New tests for `ActivityHandlerBase`
  - `ActivityHandler` extends `ActivityHandlerBase` and overrides the inherited `on*Activity()` methods from `ActivityHandlerBase`.
    -  The overrides still emit the appropriate events.
  - `ActivityHandler` now has `protected` methods that dispatch specific activity types by additional properties on the incoming Activity.
    - dispatchConversationUpdateActivity (called by overridden **onConversationUpdateActivity**)
    - dispatchMessageReactionActivity (called by overridden **onMessageReactionActivity**)
    - dispatchEventActivity (called by overridden **onEventActivity**)

## Testing
Current ActivityHandler unit tests still work, will add new unit tests (and maybe a TypeScript test collection 🚀) 

## Todo:
- [ ] Unit tests for refactored logic
- [ ] Unit tests for subclassing ActivtityHandler (Possibly also include TypeScript tests)
- [ ] Complete doc strings for new methods.

@kaiqb as FYI


libraries/botbuilder-core/src/activityHandler.ts
```diff
@@ -231,50 +231,191 @@ export class ActivityHandler {
...
+                break;
+        }
+    }
+
+    /**
+     * 
+     * @param context TurnContext A TurnContext representing an incoming Activity from an Adapter
+     * @param next () => Promise<void>
+     */
+    protected async handleMessageActivity(context: TurnContext, next: () => Promise<void>): Promise<void> {
```
johnataylor: this is now _very_ close to the C# onMessageActivity ... perhaps in the next iteration we could unify further
(FWIW the introduction of functions here is massively better factoring - much clear to read)

stevengum:
FYI @daveta @tsuwandy

PR #1225, opened 2019-09-27
add actionBasedMessagingExtension from samples repo, update .gitignore

stevengum:
Fixes: microsoft/botbuilder-js#1225

## Description
Adding `messaging-extension-action/` into `botbuilder/tests/teams`.

(Note this is not the final form, as the TeamsActivityHandler and ActivityHandler classes are updating, see microsoft/botbuilder-js#1223)


libraries/botbuilder/tests/teams/messaging-extension-action/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import { config } from 'dotenv';
+import * as path from 'path';
+import * as restify from 'restify';
+
+// Import required bot services.
+// See https://aka.ms/bot-services to learn more about the different parts of a bot.
+import { BotFrameworkAdapter } from 'botbuilder';
+
+// This bot's main dialog.
+import { TeamsActionExtensionBot } from './teamsActionExtensionBot';
```
tsuwandy: Bot is called ActionBasedMessagingExtensionBot on C# side, shall we keep them consistent naming wise?  Perhaps renamed the C# one?

stevengum: I'll rename this one to ActionBasedMessagingBot, to keep it consistent with the already checked-in scenario in C#. I'm going to add `*.zip` files to the .gitignore in this PR so we don't check in any zipped up manifests with credentials

stevengum:
@daveta and @tsuwandy I applied the feedback and the build is ongoing right now. This scenario isn't functioning right now as the signatures differ in the TeamsActivityHandler it expects versus the signatures I wrote it with.

PR #1229, opened 2019-09-30
Update choice prompt unit test for hero card prompt with attachment

jspruance:
Update to unit test for sending a choice prompt styled as a hero card and an attachment.

- Prior to this change, hero card and attachment were being sent back as two separate responses, one after another

- After adding type: ActivityTypes.Message to the attachment, both the attachment and hero card are contained in the same response, as expected.


cleemullins:
MErged as part of the group PR review

PR #1232, opened 2019-09-30
Change TS output to ES6 for all libraries

jspruance:
Relates to issue:
https://github.com/microsoft/botbuilder-js/issues/1076

'botbuilder-core' TS compiler modified to output 'ES5' instead of 'ESNext':

"target": "ES5",
"lib": [ "es2015", "dom" ]

Unit tests pass and browser bot runs with no errors in Chrome (JS errors in IE11 before and after this change).

Note that this change does not guarantee 100% ES5 compatibility since TypeScript compilation only targets syntax and does not do any polyfilling (in contrast, Babel does both). However this should not introduce any new errors - if anything ESNext --> ES5 should be safer.

We could introduce a Babel layer to achieve complete ES5 compatibility, which would look like:
TS > ES6 > Babel > ES5


stevengum:
@jspruance can you add this change to your PR? 

https://github.com/microsoft/botbuilder-js/pull/1102/files#diff-dc3dd8e5393908b2139ed0572ce6c0dfR9

jspruance:
> @jspruance can you add this change to your PR?
> 
> https://github.com/microsoft/botbuilder-js/pull/1102/files#diff-dc3dd8e5393908b2139ed0572ce6c0dfR9

@stevengum I've added your changes to the PR

cleemullins:
Could this be ES6 instead? We have ES6 features (such as in the RegEx). The browsers we target all (likley) support ES6.

cleemullins:
Conerns are (specifically) that runtime features such as the regex look-ahead, which is used in the recognizers-text, need ES6 features. These features cannot be polyfilled.

jspruance:
This has been updated to es6. Tested in browser echo bot and unit tests passed.

jspruance:
> Can you apply the output changes to the other libraries?

This is complete - ready for review.

PR #1233, opened 2019-10-01
Add TelemetryInitializerMiddleware

### LINKED ISSUE #1221: Telemetry: Support telemetry initializer middleware for C# parity
We have had to make a change to the telemetry implementation due to custom adapters not correctly having correlation data, such as user / channel / conversation ID, attached to App Insights logs.  This was due to the activity data used to populate these fields being captured from the incoming request data, which may not always be a Bot Framework Activity if a custom adapter is being used (e.g. Alexa / Twilio).

This change adds TelemetryInitializerMiddleware which can be added to the bot middleware pipeline and will store the activity **after** the adapter has transformed the incoming request into an activity.

This change was made in PR [2580 in the .NET repo](https://github.com/microsoft/botbuilder-dotnet/pull/2580).

JS appears to have the same issue as C# with the incoming request being stored [here](https://github.com/microsoft/botbuilder-js/blob/2115012394877abcfc8234ebad43d09b9ac2d9d0/libraries/botbuilder-applicationinsights/src/applicationInsightsTelemetryClient.ts#L37)

garypretty:
Added new middleware for storing activity on the Application Insights correlation context.

Fixes #1221

## Description
Add TelemetryInitializerMiddleware to JS for parity for C#

## Specific Changes
Add TelemetryInitializerMiddleware to JS for parity for C#


libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
```diff
@@ -0,0 +1,69 @@
...
+    // tslint:disable:variable-name
+    private readonly _logActivityTelemetry: boolean;
+    private readonly _logPersonalInformation: boolean;
+    // tslint:enable:variable-name
+
+    /**
+     * Initializes a new instance of the TelemetryInitializerMiddleware class.
+     * @param _telemetryClient The TelemetryClient used by the TelemetryLoggerMiddleware for logging activity telemetry.
+     * * @param _logActivityTelemetry (Optional) Enable/Disable logging of activity telemetry.
+     * * @param _logPersonalInformation (Optional) Enable/Disable logging original message name within Application Insights.
```
benbrown: these should not have _ in front of them

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
```diff
@@ -0,0 +1,69 @@
...
+     */
+    constructor(telemetryClient: TelemetryClient, logActivityTelemetry: boolean = false, logPersonalInformation: boolean = false) {
+        this._telemetryClient = telemetryClient || new NullTelemetryClient();
+        this._logActivityTelemetry = logActivityTelemetry;
```
benbrown: you used logActivityTelemetry, but did not use logPersonalInformation here.  you need to set that value into the class member as well.

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        if (context.activity && context.activity.id) {
+            const correlationContext: CorrelationContext = appInsights.getCorrelationContext();
+            correlationContext['activity'] = context.activity;
+        }
+
+        if(this._logActivityTelemetry)
+        {
+            let activityLogger = new TelemetryLoggerMiddleware(this._telemetryClient, this._logPersonalInformation);
```
benbrown: why are you constructing this here instead of passing it in, as is done with the c# version?

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
garypretty: @benbrown You're right. Made me realize that we shouldn't have this param here anyway, otherwise you hide the ability for someone to set it on the underlying logging middleware.  Ill remove the logPersonalInformation from here and raise a PR in C# too to amend.

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
garypretty: @benbrown I'll pass it in - I'll make the change tomorrow.

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
garypretty: @benbrown This param shouldn't be here and is a result of not passing in the logger middleware. I'll sort both tomorrow.

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
garypretty: @benbrown updated to pass in the logging middleware

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
garypretty: @benbrown Removed the redundant param now I am passing in the logging middleware

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
garypretty: Removed

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
```diff
@@ -0,0 +1,65 @@
...
+     * @param telemetryLoggerMiddleware The TelemetryLoggerMiddleware used for logging activity telemetry.
+     * * @param logActivityTelemetry (Optional) Enable/Disable logging of activity telemetry.
+     */
+    constructor(telemetryLoggerMiddleware: TelemetryLoggerMiddleware, logActivityTelemetry: boolean = false) {
+        this._telemetryLoggerMiddleware = telemetryLoggerMiddleware;
+        this._logActivityTelemetry = logActivityTelemetry;
+    }
+
+    /**
+     * Gets a value indicating whether determines whether to log personal information that came from the user.
```
daveta: >whether to log personal information [](start = 50, length = 35)

Hmm - might want to update..

garypretty: @daveta Thanks.  Will update.  FYI, the currently checked in unit test is not passing right now.  I have just got it running and will be checking in tests shortly.

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
```diff
@@ -0,0 +1,65 @@
...
+     * @param context The context object for this turn.
+     * @param next The delegate to call to continue the bot middleware pipeline
+     */
+    public async onTurn(context: TurnContext, next: () => Promise<void>): Promise<void> {
+        if (context === null) {
+            throw new Error('context is null');
+        }
+
+        if (context.activity && context.activity.id) {
+            const correlationContext: CorrelationContext = appInsights.getCorrelationContext();
```
daveta: >const correlationContext: CorrelationContext = appInsights.getCorrelationContext(); [](start = 12, length = 83)

You may want to add a check if dependency correlation is enabled (at least according to app insights)..
[https://github.com/microsoft/ApplicationInsights-node.js/blob/148052ebaef115cecc0bd8b3f92056ee111431fb/applicationinsights.ts#L136](https://github.com/microsoft/ApplicationInsights-node.js/blob/148052ebaef115cecc0bd8b3f92056ee111431fb/applicationinsights.ts#L136)

libraries/botbuilder-applicationinsights/tests/telemetryInitializer.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+const assert = require('assert');
+const appInsights = require("applicationinsights");
+const { TestAdapter, ActivityTypes, TelemetryLoggerMiddleware } = require('botbuilder-core');
+const { TelemetryInitializerMiddleware } = require('../');
+
+describe(`TelemetryMiddleware`, function () {
+    this.timeout(5000);
+
+    it(`telemetry initializer stores activity`, function (done) {
```
daveta: >telemetry initializer stores activity [](start = 8, length = 37)

May be valuable to do e2e with logging to make sure bot's session/userid actually land in the customEvent/dependency/trace/etc.

garypretty: @daveta I was thinking that the responsibility of this component is to 1. store the activity on the context and 2. optionally run the logging middleware.  Therefore that is what the tests for this particular class were going to cover.  I agree though that it might be worth adding tests to the class that actually grabs the activity and add the correlation properties.

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
garypretty: @daveta Comment updated.  Also additional unit tests now in place and passing.

libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts
garypretty: @daveta Added null check and unit test to ensure the middleware still runs through and call telemetry logger even if correlation is disabled.

libraries/botbuilder-applicationinsights/tests/telemetryInitializer.test.js
daveta: Seems like adding some additional asserts in the `.assertReply` might be the easiest way to validate.

Issue #1201, opened 2019-09-23
Port BFCard scenario from C# to JS


Virtual-Josh:
https://github.com/microsoft/botbuilder-dotnet/tree/master/tests/Teams/CardsBotFramework

daveta:
@EricDahlvang : Doh - I actually was working on this one..  Just put a PR.

daveta:
https://github.com/microsoft/botbuilder-js/commit/28407d398c9434562e39abadcdc3a50344296b5f

PR #1243, opened 2019-10-03
Add Cards scenario

daveta:
Fixes #https://github.com/microsoft/botbuilder-js/issues/1201

## Description
Ports over Bot Framework cards scenario.


libraries/botbuilder/tests/teams/cardBotFramework/src/cardsBot.ts
```diff
@@ -0,0 +1,128 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import {
+    TeamsActivityHandler,
+} from 'botbuilder';
+import {
```
tsuwandy: You could import them all from botbuilder…

Issue #1206, opened 2019-09-23
Port Mentions scenario from C# to JS


Virtual-Josh:
https://github.com/microsoft/botbuilder-dotnet/tree/master/tests/Teams/Mentions

PR #1245, opened 2019-10-03
Mentions scenario

daveta:
Fixes #https://github.com/microsoft/botbuilder-js/issues/1206

## Description
Mentions scenario.


libraries/botbuilder/tests/teams/mentionsBot/src/mentionsBot.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import {
+    Mention,
+    MessageFactory,
+    TeamsActivityHandler,
+} from 'botbuilder';
+
+export class MentionsBot  extends TeamsActivityHandler {
+    /*
+     * You can @mention the bot from any scope and it will reply with the mention.
```
tsuwandy: yay, I have started adding this too..

PR #1249, opened 2019-10-03
Add TeamsConnectorClient and fix BotFrameworkAdapter bug

stevengum:
- Adds TeamsConnectorClient to botframework-connector
- Fix bug in BotFrameworkAdapter introduced in #1246


libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -721,18 +722,20 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
-            body = err;
+            // Catch the error to try and throw the stacktrace out of processActivity()
+            processError = err;
+            body = err.toString();
         }
 
         // Return status 
         res.status(status);
-        if (body) { res.send(body.toString()); }
+        if (body) { res.send(body); }
```
EricDahlvang: We should have some tests around this (the .toString() breaking change should have been caught by tests).

Issue #1207, opened 2019-09-23
Port Message Reaction from C# to JS


Virtual-Josh:
https://github.com/microsoft/botbuilder-dotnet/tree/master/tests/Teams/MessageReaction

PR #1252, opened 2019-10-04
[Teams] Add message reaction bot

EricDahlvang:
Fixes https://github.com/microsoft/botbuilder-js/issues/1207


libraries/botbuilder-core/src/activityHandler.ts
```diff
@@ -351,7 +351,7 @@ export class ActivityHandler extends ActivityHandlerBase {
      */
     protected async dispatchMessageReactionActivity(context: TurnContext): Promise<void> {
         if (context.activity.reactionsAdded || context.activity.reactionsRemoved) {
-            super.onMessageReactionActivity(context);
+            await super.onMessageReactionActivity(context);
```
stevengum: thanks for catching this!

Issue #1214, opened 2019-09-24
[Teams] Add TeamsActivityHandler to botbuilder


stevengum:
### Invoke Activities:
- [x] onTeamsCardActionInvoke
- [x] onTeamsSigninVerifyState
- [x] onTeamsFileConsent
- [x] onTeamsFileConsentAccept
- [x] onTeamsFileConsentDecline
- [x] onTeamsO365ConnectorCardAction
- [x] onTeamsMessagingExtensionQuery
- [x] onTeamsAppBasedLinkQuery
- [x] onTeamsMessagingExtensionSelectItem
- [x] onTeamsMessagingExtensionFetchTask
- [x] onTeamsMessagingExtensionSubmitActionDispatch
- [x] onTeamsMessagingExtensionSubmitAction
- [x] onTeamsMessagingExtensionBotMessagePreviewEdit
- [x] onTeamsMessagingExtensionBotMessagePreviewSend
- [x] onTeamsMessagingExtensionConfigurationQuerySettingsUrl
- [x] onTeamsMessagingExtensionConfigurationSettings
- [x] onTeamsMessagingExtensionCardButtonClicked
- [x] onTeamsTaskModuleFetch
- [x] onTeamsTaskModuleSubmit

### ConversationUpdate Activities:
- [x] onConversationUpdateActivity override in TeamsActivityHandler
- [x] onTeamsMembersAdded
- [x] onTeamsMembersRemoved
- [x] onTeamsChannelCreated
- [x] onTeamsChannelDeleted
- [x] onTeamsChannelRenamed
- [x] onTeamsTeamRenamed

## Phase 2: TeamsClient property implementation

- [x] TeamsRosterClient impl (custom class)
- [x] getTeamDetails
- [x] getChannels
- [x] getMembers

## Phase 3: Beautiful Unit and Integration Tests 🚀

stevengum:
Closing this as the TeamsActivityHandler is complete. The TeamsInfo requirement/helper methods will be addressed by #1253

PR #1253, opened 2019-10-04
[Teams] Add TeamsInfo helper class, minor fix in ConversationUpdateBot scenario

stevengum:
Fixes: #1214

## Description
Adds the TeamsInfo helper class.

## Testing
Tested manually, unit tests are coming next


libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -852,7 +852,7 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
      * @remarks
      * Override this in a derived class to create a mock connector client for unit testing.
      */
-    protected createConnectorClient(serviceUrl: string): ConnectorClient {
+    public createConnectorClient(serviceUrl: string): ConnectorClient {
```
stevengum: I switched this to public because the `BotFrameworkAdapter.credentials` is protected, and `ConnectorClient.credentials` is not.

I also wanted the new ConnectorClient to include the `USER_AGENT`, not sure if this is a problem in C#...

johnataylor: could you not make the Credentials on ConnectorClient public - that would also bring the two implementations closer together.

stevengum: The credentials in the connectorClient are [public](https://github.com/microsoft/botbuilder-js/pull/1253/files#diff-8703bbaca7e46c1a78c0da79519b20d9R79), I use them in `TeamsInfo.getTeamsConnectorClient()`

```ts
return new TeamsConnectorClient(connectorClient.credentials, { baseUri: context.activity.serviceUrl });
```

johnataylor: Regarding USER_AGENT header - the C# uses the same HttpClient as the regular connector. So that should be good.
Its really just a couple of extra REST calls on top of the current connector

johnataylor:
and we need some additional unit tests?

Issue #1197, opened 2019-09-23
Port Action based Messaging Extensions (fetch task) to JS


Virtual-Josh:
Copy the C# scenario to JS

https://github.com/microsoft/botbuilder-dotnet/tree/master/tests/Teams/ActionBasedMessagingExtension-FetchTask

daveta:
I'm doing this in preview

johnataylor:
This is related to a change we want to make in the SendToChannel the code should look more like this:

https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/bots/bot-conversations/bots-conv-proactive#creating-a-channel-conversation

daveta:
PR in, waiting for feedback..

PR #1258, opened 2019-10-04
[Teams] Port of ActionBasedMessagingExtension Fetch Task

daveta:
Fixes https://github.com/microsoft/botbuilder-js/issues/1197 

## Description
This matches current C#, if we want to change behavior of how Preview stuff works, we'll change in both..


libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/cardResponseHelpers.ts
```diff
@@ -0,0 +1,50 @@
...
+
+            }
+        }
+    }
+
+    public static toMessagingExtensionBotMessagePreviewResponse(cardAttachment: Attachment) : MessagingExtensionActionResponse {
+        return <MessagingExtensionActionResponse> {
+            composeExtension: <MessagingExtensionResult> {
+                type: 'botMessagePreview',
+                activityPreview: MessageFactory.attachment(cardAttachment, null, null, InputHints.ExpectingInput) as Activity,
```
stevengum: Does this work with just `{ attachment: [cardAttachment] } as Activity`? The MessageFactory doesn't add any real value here, and the InputHint I believe is unneeded...

libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/actionBasedMessagingExtensionFetchBot.ts
```diff
@@ -0,0 +1,66 @@
...
+    protected async onTeamsMessagingExtensionBotMessagePreviewSend(context: TurnContext, action: MessagingExtensionAction) : Promise<MessagingExtensionActionResponse> {
+        var submitData = AdaptiveCardHelper.toSubmitExampleData(action);
+        var adaptiveCard = AdaptiveCardHelper.toAdaptiveCardAttachment(submitData);
+        var response = CardResponseHelpers.toComposeExtensionResultResponse(adaptiveCard);
+        return response;
+    }
+
+    protected async onTeamsMessagingExtensionCardButtonClicked(context: TurnContext, obj) {
+        const reply = MessageFactory.text("onTeamsMessagingExtensionCardButtonClicked Value: " + JSON.stringify(context.activity.value));
+        await context.sendActivity(reply);
```
stevengum: This could just be:

```ts
await context.sendActivity(`onTeamsMessagingExtensionCardButtonClicked Value: ${JSON.stringify(context.activity.value)}`);
```

The MessageFactory doesn't add much here...

libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/actionBasedMessagingExtensionFetchBot.ts
```diff
@@ -0,0 +1,66 @@
...
+                                                    submitData.Option1,
+                                                    submitData.Option2,
+                                                    submitData.Option3);
+        return response;
+    }
+
+    protected async onTeamsMessagingExtensionBotMessagePreviewSend(context: TurnContext, action: MessagingExtensionAction) : Promise<MessagingExtensionActionResponse> {
+        var submitData = AdaptiveCardHelper.toSubmitExampleData(action);
+        var adaptiveCard = AdaptiveCardHelper.toAdaptiveCardAttachment(submitData);
+        var response = CardResponseHelpers.toComposeExtensionResultResponse(adaptiveCard);
```
stevengum: Can you change these `var` to `const`? And the other `var` in this scenario?

libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/adaptiveCardHelper.ts
```diff
@@ -0,0 +1,99 @@
...
+                { type: 'TextBlock', text: 'Adaptive Card from Task Module', weight: 'bolder' },
+                { type: 'TextBlock', text: `${data.Question}`, id:"Question" },
+                { type: 'Input.Text', id: 'Answer', placeholder: 'Answer here...' },
+                { type: 'Input.ChoiceSet', id: 'Choices', isMultiSelect: Boolean(data.MultiSelect), style: 'expanded',
+                    choices: [{title: data.Option1, value: data.Option1}, 
+                            {title: data.Option2, value: data.Option2},
+                            {title: data.Option3, value: data.Option3}] },
+            ],
+            actions: [
+                { type: 'Action.Submit', title: 'Submit', data: { submitLocation: 'messagingExtensionSubmit'} },
```
stevengum: What's this `submitLocation` bit?

libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/cardResponseHelpers.ts
daveta: Just keeping consistency with C# (re: `inputhints`).  And I believe other sample in nodejs is taking same approach.

libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/actionBasedMessagingExtensionFetchBot.ts
daveta: Yep, again consistency with C#..

libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/actionBasedMessagingExtensionFetchBot.ts
daveta: Nice catch!  I'll update

libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/adaptiveCardHelper.ts
daveta: Good question.  @EricDahlvang : Do you know what this is?

stevengum:
@daveta, can you try to incorporate the fallback scenario that we discussed with Andrew? Checking if a bot is a member of the current channel/conversation and either sending a proactive message or sending a card to the compose box?

libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/adaptiveCardHelper.ts
Virtual-Josh: The submit location, I believe, is related to where the button is clicked from. This matters because the invoke location does come through in the inbound activity. Meaning you'll know if this was kicked off from the search bar, or compose area.

stevengum: @daveta I think this information factors in to the scenario outlined [here](https://github.com/microsoft/botbuilder-js/pull/1258#issuecomment-538590166). I'm not sure what the conversation data looks like when a bot receives an invoke from the command bar

EricDahlvang: submitLocation is not relevant... it could be removed, but is just helping demonstrate where the adaptive card submit occurs from.

EricDahlvang:
How does one check "if a bot is a member of the current channel/conversation"?

stevengum:
`TeamsInfo.GetRoster()` :)

libraries/botbuilder/tests/teams/actionBasedMessagingExtension-fetchTask/src/adaptiveCardHelper.ts
stevengum: On second thought ultimately it does not, if it’s not used we should consider removing it.

daveta:
> @daveta, can you try to incorporate the fallback scenario that we discussed with Andrew? Checking if a bot is a member of the current channel/conversation and either sending a proactive message or sending a card to the compose box?


I've opened https://github.com/microsoft/botbuilder-js/issues/1263 to track this (which should really be done in C# first, if that's the exemplar implementation).

libraries/botbuilder/src/teamsTurnContextHelpers.ts
```diff
@@ -2,28 +2,45 @@
...
+        channelData: <TeamsChannelData>{
+            channel: <ChannelInfo>{
+                id: teamsChannelId
+            }
+        },
+        activity: <Activity>message,
+    };
+    const adapter = <BotFrameworkAdapter>turnContext.adapter;
+    const connectorClient = adapter.createConnectorClient(turnContext.activity.serviceUrl);
+    // This call does NOT send the outbound Activity is not being sent through the middleware stack.
```
tsuwandy: Not sure if I understand the comment... :-)

stevengum:
To update others who may come to this thread, the getConversationMembers method does not retrieve the bot as it is an app, not a member. Previously provided information was incorrect.

libraries/botbuilder/src/teamsTurnContextHelpers.ts
daveta: Yeah, not super obvious - it's basically saying when you send a message it's bypassing any middleware you have set up.  I believe this is going to be addressed in a later PR.

stevengum: @johnataylor and I are going to think about it more to see if there's a sane way to expose the middleware to be called at will, or another solution is viable.

PR #1276, opened 2019-10-09
Fix casing

henryjenkins:
This was causing a problem on Linux (case-sensitive filesystem) but worked fine on MacOS.

Fixes #

## Description
Correct case (uppercase) of import declaration

## Specific Changes
Change import declaration case to match filename (first letter is uppercase)

## Testing
Tested in client application on bot MacOS and Linux.


cleemullins:
This file is in-process of being deleted by changing coming this week. @carlosscastro is driving those changes. For now, I'm going to close this PR out, as it's not needed.

cleemullins:
I have confirmed this file no longer exists.

PR #1278, opened 2019-10-09
Isomorphic Web Socket

compulim:
## Description

Moving from [`watershed`](https://npmjs.com/package/watershed) to [`ws`](https://npmjs.com/package/ws). Rework with an isomorphic Web Socket approach.

## Specific Changes

Moving to `ws` which closely resembles W3C Web Socket API.

We are now on an isomorphic approach. The only difference between `ws` and Web Socket is:

- `ws.onmessage` returns a `Buffer` (`Uint8Array`)
- `WebSocket.onmessage` returns a `Blob`, which need to use `FileReader.readAsArrayBuffer` to convert it
   - We added a new `createAsyncSequencer` to resolve promise sequentially

## Next steps

We are still on `ISocket` interface for now, almost ready to eliminate it, need some investigations:

- Can `connect()` be eliminated? I.e. it will connect immediately when the object is constructed
   - Look at `WebSocketClient.ts` line 64 and 71

## Testing

- [x] Browser Web Socket tested
- [ ] Node.js Web Socket tested


libraries/botframework-streaming-extensions/src/WebSocket/IsomorphicWebSocket.ts
```diff
@@ -0,0 +1,175 @@
...
+    /// True if the socket is currently connected.
+    /// </summary>
+    public isConnected(): boolean {
+        return !!this.webSocket && this.webSocket.readyState === 1;
+    }
+
+    /// <summary>
+    /// Writes a buffer to the socket and sends it.
+    /// </summary>
+    /// <param name="buffer">The buffer of data to send across the connection.</param>
```
DDEfromOR: buffer is message now

libraries/botframework-streaming-extensions/src/WebSocket/IsomorphicWebSocket.ts
```diff
@@ -0,0 +1,175 @@
...
+    public write(message: any) {
+        this.webSocket.send(message);
+    }
+
+    /// <summary>
+    /// Connects to the supporting socket using WebSocket protocol.
+    /// </summary>
+    /// <param name="url">The address the server is listening on.</param>
+    /// <param name="port">The port the server is listening on, defaults to 8082.</param>
```
DDEfromOR: Port is no longer needed

libraries/botframework-streaming-extensions/package.json
```diff
@@ -22,7 +22,7 @@
   "dependencies": {
     "promise.prototype.finally": "^3.1.0",
     "uuid": "^3.3.2",
-    "watershed": "^0.4.0"
```
DDEfromOR: This will probably require changes to BotFrameworkStreamingAdapter to no longer build and pass in a Watershed socket.

libraries/botframework-streaming-extensions/src/WebSocket/NodeWebSocket.ts
```diff
@@ -5,98 +5,15 @@
...
-            shed.connect(res, socket, head, wskey);
-        });
-
-        this.connected = true;
-
-        return new Promise<void>((resolve, reject): void => {
-            req.on('close', resolve);
-            req.on('error', reject);
+export class NodeWebSocket extends IsomorphicWebSocket {
+    constructor(webSocket?: WebSocket) {
```
DDEfromOR: The adapter will need to be updated to pass in a different object when setting up a WebSocket connection.

libraries/botframework-streaming-extensions/package.json
compulim: Good call. Will look into it.

libraries/botframework-streaming-extensions/src/WebSocket/NodeWebSocket.ts
compulim: Do you mean `constructor` could take up object other than `WebSocket` (from `ws`)?

DDEfromOR: Yeah, the WebSocket connection setup here https://github.com/microsoft/botbuilder-js/blob/1fe68b7472081df968d691c4b9bb69db607e419f/libraries/botbuilder-streaming-extensions/src/BotFrameworkStreamingAdapter.ts#L91 will need to be changed to pass in an object compatible with the WebSocket, instead of the WaterShed specific socket it's currently sending.

DDEfromOR: Oh that's cute, I didn't know it did mini-previews now. It's the three line connection setup dance that begins on line 91.

carlosscastro:
Closing since we wrote this differently.

Issue #1281, opened 2019-10-09
[DCR] Create a StorageBaseTests Class


mdrichardson:
The Dotnet repo has a [StorageBaseTests class](https://github.com/microsoft/botbuilder-dotnet/blob/master/tests/Microsoft.Bot.Builder.Tests/StorageBaseTests.cs) that all of the Storage classes borrow from to reduce duplicate code and ensure that the Storage classes all pass a base set of tests.

JS SDK doesn't have this. Instead, there's a ton of duplicate code.

It's a big task to make the class and then port everything over, but I'll put this on my backburner.

Progress:

- [x] Create StorageBaseTests
- [x] Convert CosmosDbPartitionedStorage
- [x] Convert CosmosDbStorage
- [x] Convert BlobStorage
- [X] Convert MemoryStorage

PR #1282, opened 2019-10-09
Cosmos partitioning

### LINKED ISSUE #1234: Port new CosmosDB Partitioned Storage from C#
A new Cosmos DB storage implementation which utilizes the v3 API and handles the partition key internally has been implemented in C#, which needs adding to the JS library.

C# implementation is here https://github.com/microsoft/botbuilder-dotnet/pull/2613

mdrichardson:
Fixes #1234 

## Description
Parity with [Dotnet](https://github.com/microsoft/botbuilder-dotnet/pull/2613), adding a separate class for supporting Partitioned Containers on Cosmos.

## Specific Changes
- The new implementation no longer handles database creation. We only handle container creation.
- Uses the new `@azure/cosmos` SDK
- Implements partitioning for the user, with the option to pass in a key now gone. We utilize the ID of the item as the partition key.

Also implements [these additional changes from the Dotnet PR](https://github.com/microsoft/botbuilder-dotnet/pull/2684)

## Testing

Added new, separate tests.

image

On a side note: I plan to eventually [create a StorageBaseTests class](https://github.com/microsoft/botbuilder-js/issues/1281) for the JS repo, so we remove a lot of the duplicate code. Wanted to get this out for 4.6, though.


mdrichardson:
Note: Currently figuring out how to make the tests record.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
```diff
@@ -0,0 +1,255 @@
...
+            // The ETag information is updated as an _etag attribute in the document metadata.
+            delete changesCopy.eTag;
+            const documentChange = new DocumentStoreItem({
+                id: CosmosDbKeyEscape.escapeKey(k),
+                realId: k,
+                document: changesCopy
+            });
+
+            const eTag: string = changes[k].eTag;
+            if (!eTag || eTag === '*') {
```
christopheranderson: I don't think you need to differentiate upsert vs replace on etag here. Upsert IfMatch will always succeed if it doesn't exist.

I wrote up a simple proof of concept showing this off.
https://gist.github.com/christopheranderson/57081bce1236a6941a553e9b9333c22d

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
```diff
@@ -0,0 +1,255 @@
...
+                // If we have an etag, do opt. concurrency replace
+                try {
+                    await this.container
+                        .item(CosmosDbKeyEscape.escapeKey(k), documentChange.partitionKey)
+                        .replace(documentChange, { accessCondition: { type: 'IfMatch', condition: eTag } });
+                } catch (err) {
+                    this.throwInformativeError('Error replacing document', err);
+                }
+            } else {
+                throw new Error(`etag empty`);
```
christopheranderson: `!etag` will catch "empty" etags (it will catch all falsey etags aka `"" == false` in JS). If you want to catch empty etags, you can change your if condition to check for `undefined` explicitly instead of just checking for falsey to only handle cases where etag is not defined on the object (because it is new).

christopheranderson: this currently just catches cases where it's not falsey and the object doesn't have a length property.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    public async delete(keys: string[]): Promise<void> {
+
+        await this.initialize();
+
+        await Promise.all(keys.map(async (k: string): Promise<void> => {
+            try {
+                await this.container
+                    .item(CosmosDbKeyEscape.escapeKey(k), k)
```
christopheranderson: should the pk be `escapeKey(k)` here as well?

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
```diff
@@ -0,0 +1,255 @@
...
+    public document: object
+    /**
+     * Gets or sets the ETag information for handling optimistic concurrency updates.
+     */
+    public eTag: string
+    /**
+     * Gets the PartitionKey value for the document.
+     */
+    public get partitionKey(): string {
+        return this.id;
```
christopheranderson: would this need to be `escapeKey(this.id)`?

mdrichardson:
@christopheranderson Great feedback! I should be able to fix those tomorrow.

cleemullins:
Please merge when ready.

mdrichardson:
@cleemullins Will do. Implementing @christopheranderson's recommendations now and will merge once build tests passes.

However, coveralls test coverage won't allow merge. I believe the issue is that most of the tests don't actually run since there's no Storage Emulator. Is that something we can bypass? ...I don't believe any of the JS storage tests actually run the read/write/deletes in the pipeline tests.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
```diff
@@ -0,0 +1,255 @@
...
+                    storeItems[documentStoreItem.realId].eTag = documentStoreItem._etag;
+                }
+            } catch (err) {
+                // When an item is not found a CosmosException is thrown, but we want to
+                // return an empty collection so in this instance we catch and do not rethrow.
+                // Throw for any other exception.
+                if (err.code === 404) { }
+                // Throw unique error for 400s
+                else if (err.code === 400) {
+                    this.throwInformativeError(`Error initializing container. You might be using partitions in a non-partitioned DB or
```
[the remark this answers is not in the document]
mdrichardson: Good call. I've re-worded it to:

> Error reading from container. You might be attempting to read from a non-partitioned container or a container that does not use '/id' as the partitionKeyPath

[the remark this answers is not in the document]
christopheranderson: If you wanna handle NP containers, as well, it'd only be a little more work. Dunno if that's in scope or not, though.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
```diff
@@ -0,0 +1,255 @@
...
+                if (documentStoreItem) {
+                    storeItems[documentStoreItem.realId] = documentStoreItem.document;
+                    storeItems[documentStoreItem.realId].eTag = documentStoreItem._etag;
+                }
+            } catch (err) {
+                // When an item is not found a CosmosException is thrown, but we want to
+                // return an empty collection so in this instance we catch and do not rethrow.
+                // Throw for any other exception.
+                if (err.code === 404) { }
+                // Throw unique error for 400s
```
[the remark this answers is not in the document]
mdrichardson: To maintain consistency with Dotnet, I'll probably just leave this for now. If we see issues on the support side, I might make the change in both. With the `initialize()` call, they *shouldn't* make it this far with a bad container/db, so this *should* only throw for missing items.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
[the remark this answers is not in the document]
mdrichardson: We've got a separate class for that (`CosmosDbStorage` vs. this, `CosmosDbPartitionedStorage`). We considered rolling them under the same class, but it gets a little more funky forcing them to use `/id` while maintaining backwards-compat with `CosmosDbStorage`, if they already had custom partitionKeys defined.

That being said, if you think we can adjust `CosmosDbStorage` to cover everything, might be worth having a conversation with Chris about. See [here](https://github.com/microsoft/botframework-sdk/issues/5467) for more info on our discussion of which direction to take.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
mdrichardson: Can you clarify that for me? I think I get it but want to make sure before we merge...I should get rid of `replace` and just use `upsert` for both `if` branches, right?

garypretty:
@mdrichardson @christopheranderson Let's discuss handling non partitioned etc tomorrow when you get online. I have some specific thoughts here.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
mdrichardson: Oh gross. That's embarrassing.

I switched to `if (etag == null)` to catch both `undefined` and the unlikely `null`.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
mdrichardson: Yep. Fixed.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
mdrichardson: Good question. I don't *believe* it's necessary, since it looks like we escape it in all the necessary spots in the code. The dotnet side also doesn't have it. If you feel it's needed, I can add it and submit a PR for the dotnet side. Thoughts?

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
mdrichardson: @christopheranderson Hey Chris, do you have anything to add for these two unresolved conversations, or would you say we're good to merge?

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
christopheranderson: seems odd that we'd set the id to escapeKey(id) and partition key to just id. That would imply that escapeKey isn't doing anything that the SDK wouldn't already do for you. Don't need to block this PR for it, but it seems like we should try removing escapeKey and seeing if we have any tests that fail.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
christopheranderson: Yes, you can just use upsert. You can probably avoid the if condition and use a ternary statement instead, as well.

christopheranderson: Your code works fine, so feel free to resolve if you want, but you could shrink your LOC count down. :)

mdrichardson: good call. done.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
mdrichardson: The [crazy keys test[(https://github.com/microsoft/botbuilder-js/pull/1282/files#diff-e509e24b560c2c2464b96c7c9b2e119bR125) fails when `escapeKey` calls are removed. I'm going to leave the code as-is, for this part.

PR #1284, opened 2019-10-10
Adding OAuthCard support for Streaming Extensions

Jeffders:
## Description
This adds OAuthCard resolution to connections made with streaming extensions.

## Specific Changes
Adds the polling mechanism to retrieve OAuth tokens when an OAuthCard is sent via streaming extensions. This was the approved design, so porting it to the JS bot builder. 

Basically:
When an OAuthCard is sent on streaming connection, the BotAdapter begins to poll for that particular token until it is found. The service can override and terminate the polling if need be. Once the token is retrieved, it is sent back through the adapter.

## Testing
Added new tests that match the C# implementation.


carlosscastro:
This is a port from a c# PR that we iterated a lot on. Looks good!

cleemullins:
Please merge when ready.

PR #1294, opened 2019-10-12
Relax the rules for what counts as a well formed GUID.

DDEfromOR:
The previous regex pattern used to determine if a header's identifier GUID was well formed enforced the timestamp based UUID rules as laid out in RFC 4122. The dotnet standard library used in C# bots only requires a GUID to match the 8-4-4-4-12 pattern to be valid, making them less restrictive. 

This change relaxes the regex matching to match any GUID following the 8-4-4-4-12 pattern for parity with the C# library.


DDEfromOR:
Added a test to make sure nil GUIDs are accepted.

PR #1300, opened 2019-10-14
[botframework-config] Remove browser compat libs

christopheranderson:
## Description

We were depending on `process` and `url` from npm, which are designed for browser compat. Given that our other libraries do not use these libs for browser compat, I'm removing them in this PR for consistency.

If we do intend for them to be used in the browser, we can close this and I'll open an issue to have a browser test for this library.

## Specific Changes

  - remove `process` and `url` dependencies in `botframework-config`

## Testing

N/A - This does not modify functionality.


stevengum:
We've moved away from using `.bot` files and have therefore stopped using these files.

We do need browser tests, which might be a 4.7 item.

PR #1303, opened 2019-10-14
normalize azure blob timestamp properties to be iso1806 format

tomlm:
fix for  https://github.com/microsoft/botbuilder-dotnet/issues/2120

## Description
js code was storing as number of seconds since 1970, and C# was using iso 1806

## Specific Changes
changed to use Date().toJSON() which is iso 1806

## Testing
updated tests

### DIFF
--- a/libraries/botbuilder-azure/tests/azureBlobTranscriptStore.test.js
+++ b/libraries/botbuilder-azure/tests/azureBlobTranscriptStore.test.js
@@ -198,7 +198,6 @@ describe('The AzureBlobTranscriptStore', () => {
 	it('should log an activity', async () => {
 		const date = new Date(1546214400000);
 		const activity = createActivity('logActivityTest', date);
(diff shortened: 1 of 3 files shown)


libraries/botbuilder-azure/tests/azureBlobTranscriptStore.test.js
```diff
@@ -196,9 +196,9 @@ describe('The AzureBlobTranscriptStore', () => {
 	});
 
 	it('should log an activity', async () => {
-		const date = {getTime: () => 1546214400000};
+		const date = new Date(1546214400000);
 		const activity = createActivity('logActivityTest', date);
-
+console.log(date.toJSON());
```
carlosscastro: Indentation seems off here

tomlm: woops don't want that line :)

PR #1337, opened 2019-10-21
[R7] Authentication: Use ADAL to acquire tokens + support certificate auth

carlosscastro:
# Authentication: Use ADAL to acquire tokens + support certificate auth

Fixes #1181

Summary of changes:

- Use Adal for token AAD acquisition: Currently we craft the http request and send it to login.microsoftonline.com/etc. Moving to ADAL to provide support for all oauth client authentication flows and be more robust to AAD api changes
- Extract common base class for client auth, AppCredentials, containing most of the former MicrosoftAppCredentials
- MicrosoftAppCredentials externally looks the same, but internal now extends AppCredentials
- Certificate authentication: Adding support for certificate authentication through the new class CertificateAppCredentials
- Adapter integration: adding parameters in the adapter settings to configure authentication

Usage:

Create an AAD app in the azure portal, and under secrets and keys, upload a certificate for authentication

image

when you create a bot adapter in your bot startup code, you can use your library of choice to load your certificate thumbprint and private key:

```javascript
const pem = require('pem');
const fs = require('fs');

// In windows, configure openssl path or add it to path
pem.config({
    pathOpenSSL: '<full path to openssl>'
});

const pfx = fs.readFileSync(<path to pfx>);
pem.readPkcs12(pfx, { p12Password: '<cert password>' }, (err, cert) => {

    if(err) {
        console.log(err);
        return;
    }

    // Create adapter.
    // See https://aka.ms/about-bot-adapter to learn more about adapters.
    const adapter = new BotFrameworkAdapter({
        appId: '<app id>',
        useCertificateAuthentication: true,
        certificateThumbprint: '<thumbprint>',
        certificateKey: cert.key
    });
```

PR #1338, opened 2019-10-21
[Streaming] Auth and browser compatibility fixes, linting, refactoring

stevengum:
## Description

- Add the ServiceUrl value from the claims in `BotFrameworkAdapter.authenticateConnection()` to have the bot call getToken() to refresh (or cache) a token.
- Refactor some code in `BotFrameworkAdapter.processRequest()` (streaming entrypoint)
- Move `ws` and `Watershed` Server/"WaterShed" construction into the WsNodeWebSocket and NodeWebSocket methods so the botframework-streaming library is browser compat (thanks for the report @ckkashyap!)
- Linting

### DIFF
--- a/libraries/botbuilder/src/botFrameworkAdapter.ts
+++ b/libraries/botbuilder/src/botFrameworkAdapter.ts
@@ -1003,11 +1003,24 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
             response.statusCode = StatusCodes.BAD_REQUEST;
             response.setBody(`Request missing verb and/or path. Verb: ${ request.verb }. Path: ${ request.path }`);
             return response;
-        } 
+        }
 
         if (request.verb.toLocaleUpperCase() === GET && request.path.toLocaleLowerCase() === VERSION_PATH) {
             response.statusCode = StatusCodes.OK;
-            response.setBody({UserAgent: USER_AGENT});
+
+            if (!this.credentials.appId || !this.credentials.appPassword) {
+                response.setBody({UserAgent: USER_AGENT});
+                return response;
+            }
+
+            try {
+                response.setBody({UserAgent: USER_AGENT, Token: await this.credentials.getToken()});
+            } catch (err) {
+                // If the MicrosoftAppCredentials.getToken() fails, not sending the UserAgent in this
+                // request will result in the connection being destroyed by the channel.
+                response.statusCode = StatusCodes.INTERNAL_SERVER_ERROR;
+                response.setBody(err.message);
+            }
 
             return response;
         }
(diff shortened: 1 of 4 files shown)


libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1003,11 +1003,24 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
             response.setBody(`Request missing verb and/or path. Verb: ${ request.verb }. Path: ${ request.path }`);
             return response;
-        } 
+        }
 
         if (request.verb.toLocaleUpperCase() === GET && request.path.toLocaleLowerCase() === VERSION_PATH) {
             response.statusCode = StatusCodes.OK;
-            response.setBody({UserAgent: USER_AGENT});
+
+            if (!this.credentials.appId || !this.credentials.appPassword) {
```
carlosscastro: Look at my side PR: https://github.com/microsoft/botbuilder-js/pull/1337

I would love to not condition things on the existance of the password because we have to assume that not all auth methods use passwords. This won't even compile when I merge my PR. Can we rely on app id + the call to Gettoken() in line 1017?

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1156,27 +1169,28 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
         return serviceUrl && !serviceUrl.toLowerCase().startsWith('http');
      }
 
-    private async authenticateConnection(req: WebRequest, appId?: string, appPassword?: string, channelService?: string): Promise<boolean> {
-        if (!appId || !appPassword) {
+    private async authenticateConnection(req: WebRequest, channelService?: string): Promise<void> {
+        if (!this.credentials.appId || !this.credentials.appPassword) {
```
carlosscastro: same here on password

DDEfromOR: I like that pattern better, it's been bothering me that we make the GetToken() call here and don't do anything with it.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1156,27 +1169,28 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
+        if (!claims.isAuthenticated) { throw new Error('Unauthorized Access. Request is not authorized'); }
     }
 
     /**
      * Connects the handler to a Named Pipe server and begins listening for incoming requests.
      * @param pipeName The name of the named pipe to use when creating the server.
      * @param logic The logic that will handle incoming requests.
      */
-    private async useNamedPipe(pipeName: string = defaultPipeName, logic: (context: TurnContext) => Promise<any>): Promise<void>{
+    public async useNamedPipe(pipeName: string = defaultPipeName, logic: (context: TurnContext) => Promise<any>): Promise<void>{
```
DDEfromOR: I hate making this public, because then we're stuck with it and at that point we may as well be releasing all of the NamedPipe support instead of holding some of it for a later release. 

Let's talk this over as a group and see if can come up with something clever.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1003,11 +1003,24 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
             response.statusCode = StatusCodes.OK;
-            response.setBody({UserAgent: USER_AGENT});
+
+            if (!this.credentials.appId || !this.credentials.appPassword) {
+                response.setBody({UserAgent: USER_AGENT});
+                return response;
+            }
+
+            try {
+                response.setBody({UserAgent: USER_AGENT, Token: await this.credentials.getToken()});
```
DDEfromOR: Let's rename "Token" to "BotToken" just to make it clear it's not a user token, since the other half of this line is UserAgent

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1156,27 +1169,28 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
-        let channelIdHeader: string = req.headers.channelid || req.headers.ChannelId || req.headers.ChannelID || '';
-        let credentials = new MicrosoftAppCredentials(appId, appPassword);
-        let credentialProvider = new SimpleCredentialProvider(credentials.appId, credentials.appPassword);
-        let claims = await JwtTokenValidation.validateAuthHeader(authHeader, credentialProvider, channelService, channelIdHeader);
+        const authHeader: string = req.headers.authorization || req.headers.Authorization || '';
+        const channelIdHeader: string = req.headers.channelid || req.headers.ChannelId || req.headers.ChannelID || '';
+        // Validate the received Upgrade request from the channel.
+        const claims = await JwtTokenValidation.validateAuthHeader(authHeader, this.credentialsProvider, channelService, channelIdHeader);
+        // Confirm the bot is able to fetch a token which is required to send Activities.
+        await this.credentials.getToken();
```
DDEfromOR: As per our talk, lets remove this call to getToken, unless we find a use for the return value.

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: Done!

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: I've removed the falsey password check, and taken the `getToken()` logic out of this method.

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: That's fine with me!

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: useNamedPipe is private again!

Issue #1340, opened 2019-10-21
Ensure Activity.Id is empty before sending


EricDahlvang:
## Describe the bug
We are currently sending Activity.Ids IF the user has provided one in the Activity.  The spec states: 

> A2031: Clients and bots SHOULD NOT include an id field in activities they generate.

ref: https://github.com/microsoft/botframework-sdk/blob/master/specs/botframework-activity/botframework-activity.md#id

ideal fix also change .js to add ResourceResponse.Id to activity after sent (if present, of course)

[bug]

PR #1344, opened 2019-10-23
add random to transcript generated ids and set sent activity.id

EricDahlvang:
Fixes https://github.com/microsoft/botbuilder-js/issues/1340

Also add random with prefix to log generated ids.


libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -429,6 +429,13 @@ export class TurnContext {
                 .then((responses: ResourceResponse[]) => {
                     // Set responded flag
                     if (sentNonTraceActivity) { this.responded = true; }
+                    if(responses) {
+                        output.map((a: Partial<Activity>, index: number) => {
+                            if (index < responses.length) {
+                                a.id = responses[index].id;
+                            }
+                        });
+                    }
```
stevengum: This all looks good, but can you add a test for this?

libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -420,7 +420,7 @@ export class TurnContext {
             const o: Partial<Activity> = TurnContext.applyConversationReference({...a}, ref);
             if (!o.type) { o.type = ActivityTypes.Message; }
             if (o.type !== ActivityTypes.Trace) { sentNonTraceActivity = true; }
-
+            if (o.id) { delete o.id; }
```
stevengum: Can you add a test for this? The bot sends the same activity twice and the id should be different on both activities.

libraries/botbuilder-core/src/transcriptLogger.ts
```diff
@@ -52,21 +52,19 @@ export class TranscriptLoggerMiddleware implements Middleware {
...
+                // For certain channels, a ResourceResponse with an id is not always sent to the bot.
+                // This fix uses the timestamp on the activity to populate its id for logging the transcript.
+                // If there is no outgoing timestamp, the current time for the bot is used for the activity.id.
+                // See https://github.com/microsoft/botbuilder-js/issues/1122
+                if (!clonedActivity.id) {
+                    const prefix = `g_${Math.random().toString(36).slice(2,8)}`;
+                    if (clonedActivity.timestamp) {
+                        clonedActivity.id = `${prefix}${new Date(clonedActivity.timestamp).getTime().toString()}`;
+                    } else {
+                        clonedActivity.id = `${prefix}${Date.now().toString()}`;
```
Stevenic: did you mean to not call getTime() here?

libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -429,6 +429,13 @@ export class TurnContext {
                 .then((responses: ResourceResponse[]) => {
                     // Set responded flag
                     if (sentNonTraceActivity) { this.responded = true; }
+                    if(responses) {
+                        output.map((a: Partial<Activity>, index: number) => {
+                            if (index < responses.length) {
+                                a.id = responses[index].id;
+                            }
+                        });
+                    }
```
Stevenic: what's the point of this? The output has been sent and is about to be thrown away...

EricDahlvang: Yeah, this is actually supposed to be:
```js
                        activities.map((a: Partial<Activity>, index: number) => {
                            if (index < responses.length) {
                                a.id = responses[index].id;
                            }
                        });
```

EricDahlvang: removed

libraries/botbuilder-core/src/transcriptLogger.ts
```diff
@@ -52,23 +52,23 @@ export class TranscriptLoggerMiddleware implements Middleware {
...
-                        // If there is no outgoing timestamp, the current time for the bot is used for the activity.id.
-                        // See https://github.com/microsoft/botbuilder-js/issues/1122
-                        if (!clonedActivity.id) {
-                            if (clonedActivity.timestamp) {
-                                clonedActivity.id = new Date(clonedActivity.timestamp).getTime().toString();
-                            } else {
-                                clonedActivity.id = Date.now().toString();
-                            }
-                        }
+                if (index < responses.length && !clonedActivity.id) {
```
stevengum: We should always set the resourceResponse Id on the activity to be logged down here.

stevengum:
I've removed the label for 4.6 for now since it doesn't seem like we're going to add this to our next RCs. @cleemullins, @sgellock if that's not kosher please relabel this PR.

sgellock:
@stevengum sounds right.  Thank you

libraries/botbuilder-core/src/transcriptLogger.ts
EricDahlvang: Good point ... removed '!clonedActivity.id'

PR #1356, opened 2019-10-24
Update typeDoc comments in dialog.ts

JonathanFingold:
## Description
- Fixed links
- Updated descriptions, hopefully for clarity and completeness.

## Specific Changes
- Major edits to all typeDoc comments in dialog.ts.
- Minor copy edits in dialogContext.ts

## Testing
Will do a visual review of the released docs and submit updates after the 4.6 release, as necessary.


libraries/botbuilder-dialogs/src/dialog.ts
```diff
@@ -1,6 +1,3 @@
-/**
- * @module botbuilder-dialogs
- */
```
stevengum: @JonathanFingold do we no longer need this in our files? What's the rule that we on the SDK team should follow for this docstring/comment?

JonathanFingold: @stevengum Our pubs team uses typeDoc + type2docFx + docFx. Only one `@module` comment is needed per module, and then only if it includes a description. (typeDoc is smart enough to figure out the module name on its own.)

Without a description, it is effectively a no-op (for our publishing pipeline). If there are multiple such tags across multiple files within a module. I believe the "first" description encountered is used.

Stevenic: we needed this comment do to a TypeDoc plugin we were using. You're saying Docs.com doesn't use the same module plugin? It used to.

Stevenic: Here's the plugin: https://github.com/christopherthielen/typedoc-plugin-external-module-name

JonathanFingold: @stevenic - I've asked the publishing team to comment on this thread. (For what it's worth, I've removed the `@module` from other files in previous PRs, [TurnContext](https://docs.microsoft.com/en-us/javascript/api/botbuilder-core/turncontext?view=botbuilder-ts-latest), for instance.)

JonathanFingold: @stevengum @Stevenic - I got confirmation this morning, "We’re not using any plugin for typedoc."

PR #1371, opened 2019-10-28
Storage base tests

### LINKED ISSUE #1281: [DCR] Create a StorageBaseTests Class

mdrichardson:
Fixes #1281

Ignore the test recording, `*.json` file changes.

## Description

The Dotnet repo has a [StorageBaseTests class](https://github.com/microsoft/botbuilder-dotnet/blob/master/tests/Microsoft.Bot.Builder.Tests/StorageBaseTests.cs) that all of the Storage classes borrow from to reduce duplicate code and ensure that the Storage classes all pass a base set of tests.

JS SDK doesn't have this. Instead, there's a ton of duplicate code...until now!

## Specific Changes

This doesn't remove any tests from any of the storage providers. Instead, it replaces them with equivalent tests and in some cases, adds a few for tests that probably should have been conducted, but weren't.

  - Adds `StorageBaseTests` class to `botbuilder-core` for parity w/ Dotnet, which Memory, Blob, Cosmos, and CosmosPartitioned storage providers now implement in addition to their own tests (mostly constructor argument checks)
  - Structure/layout of Blob, Cosmos, CosmosPartitioned now matches more closely
  - Some NOOP/linting fixes and general code cleanup
  - Disregard all of the test re-recordings

## Changes Worth a Closer Look

- [Deprecation message for `CosmosDbStorageSettings.partitionKey`](https://github.com/microsoft/botbuilder-js/compare/master...mdrichardson:storageBaseTests?expand=1#diff-93460b58cad2ee3a9c820fc0e37a8db0R49). Figured I may as well add it. Should I instead just add it for the entire `CosmosDbStorage` class? Should I just remove it?
- All tests have a [`testRan` check](https://github.com/microsoft/botbuilder-js/compare/master...mdrichardson:storageBaseTests?expand=1#diff-e90851fc53a7a6993bd18c7a2e292367R65). Is this necessary? I'd seen it used in some dotnet tests and thought it might be worthwhile since there aren't `assert` checks in most of the individual provider's tests.
- I added a `null` keys check for both [read](https://github.com/microsoft/botbuilder-js/compare/master...mdrichardson:storageBaseTests?expand=1#diff-5722ee1e8df94705e277b92ae39b456eR41) and [write](https://github.com/microsoft/botbuilder-js/compare/master...mdrichardson:storageBaseTests?expand=1#diff-5722ee1e8df94705e277b92ae39b456eR55) `MemoryStorage`. This gives [parity with dotnet](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder/MemoryStorage.cs#L113). Since without the null check, it would throw anyway, this doesn't break backwards-compat, right???
- MemoryStorage was throwing an eTag conflict where other tests were not. [This](https://github.com/microsoft/botbuilder-js/compare/master...mdrichardson:storageBaseTests?expand=1#diff-5722ee1e8df94705e277b92ae39b456eR71) should fix it but I'd like confirmation that it won't break backwards-compat.
- Most of the providers [respond to null keys in read/write differently](https://github.com/microsoft/botbuilder-js/compare/master...mdrichardson:storageBaseTests?expand=1#diff-8955aeefb89616546b7492b8b87ae47cR32). Should I just make them output the same thing? I held off due to potential backwards-compat issues.
- This is just generally a fairly big change and has implications on testing validity of future PRs. Worth a pretty close look just for that reason.


libraries/botbuilder-core/src/memoryStorage.ts
```diff
@@ -38,6 +38,7 @@ export class MemoryStorage  implements Storage {
     }
 
     public read(keys: string[]): Promise<StoreItems> {
+        if (!keys) { throw new ReferenceError(`Keys are required when reading.`); }
```
Stevenic: >if (!keys) { throw new ReferenceError(`Keys are required when reading.`); } [](start = 8, length = 75)

move inside the promise

libraries/botbuilder-core/src/memoryStorage.ts
```diff
@@ -51,6 +52,7 @@ export class MemoryStorage  implements Storage {
     }
 
     public write(changes: StoreItems): Promise<void> {
+        if (!changes) { throw new ReferenceError(`Changes are required when writing.`); }
```
Stevenic: >if (!changes) { throw new ReferenceError(`Changes are required when writing.`); } [](start = 8, length = 81)

move inside the promise

libraries/botbuilder-core/src/memoryStorage.ts
```diff
@@ -66,7 +68,7 @@ export class MemoryStorage  implements Storage {
                     saveItem(key, newItem);
                 } else {
                     const oldItem: any = <any>JSON.parse(old);
-                    if (newItem.eTag === oldItem.eTag) {
+                    if (newItem.eTag === oldItem.eTag || !newItem.eTag) {
```
Stevenic: >|| !newItem.eTag [](start = 54, length = 16)

I would move this to the outer if() since you're basically just defaulting newItem.eTag to '*'

libraries/botbuilder-core/src/memoryStorage.ts
mdrichardson: This isn't me debating, so much as trying to learn. Is this a style thing or does it provide a functional benefit?

Stevenic: If these methods were coded using async/await then technically everything would be inside the promise but my comment was more targeted at the fact that if you're not using async/await to consume this then you now need to handle both an exception or the promise being rejected.  If its inside the promise you just need to handle the promise rejection.

Stevenic: In your example its the outer async function that's mapping the throw into a rejected promise.

mdrichardson: @stevenic Gotcha. Since the promise is part of a `map` loop, would it be better (or work) to `reject` instead of `throw`? I'm sure the performance impact of a looping `if (!keys)` check is negligible, I'd just like to avoid it, if possible.

mdrichardson: @Stevenic. Oh wow I feel dumb. Completely disregard all of this. I didn't see that these were on `MemoryStorage` and not on `CosmosDbPartitionedStorage`. Ugh.

libraries/botbuilder-core/src/memoryStorage.ts
mdrichardson: Done.

libraries/botbuilder-core/src/memoryStorage.ts
mdrichardson: Done.

mdrichardson:
@stevenic Before we merge, do you have any comments/thoughts on the "**Changes Worth a Closer Look**" section at the top post?

PR #1373, opened 2019-10-28
Fix ChoicePrompt ListStyle.none when set via PromptOptions

### LINKED ISSUE #1370: Impossible to set ListStyle.none to ChoicePrompt.
## Versions
Botbuilder version 4.5.3
NodeJS version 10.16.3

## Describe the bug
It is impossible to set ListStyle.none to ChoicePrompt.

## To Reproduce
1.  Set a promptOptions with style as ListStyle.none
`const promptOptions: PromptOptions = { prompt: 'Say a letter' ,  choices: ['A', 'B'], style: ListStyle.none }`
2.  Call to a prompt type ChoicePromt with this promptOptions
`return await stepContext.prompt('CHOICE_PROMPT', promptOptions);`
3. Options A and B are shown on the screen when they should not be shown.

## Expected behavior
Show only the text  'Say a letter'  in the screen. (client)

## Additional context
The problem is in ChoicePrompt.ts
The style is set to ListStyle.auto by default
Line 67: ` this.style = ListStyle.auto;`
In the onPrompt method if the options.style is false, null undefined or 0 , the assigned style will be the default one (ListStyle.auto).
Line 83: `const choiceStyle: ListStyle = options.style || this.style;`
The number none is assigned the value 0 . 
File prompt.ts Line 17 :
```
export enum ListStyle {
    /**
     * Don't include any choices for prompt.
     */
    none,
```
`

[bug]

mdrichardson:
Fixes #1370 

* Parity [not needed for C#](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.Dialogs/Prompts/ChoicePrompt.cs#L137), unless I misunderstand null-coalescing operators).

* [Parity PR for Python](https://github.com/microsoft/botbuilder-python/pull/373)

## Description

`ChoicePrompt` determines style with:

```ts
const choiceStyle: ListStyle = options.style || this.style;
```

Since `ListStyle.none` enumerates to `0`, which is falsy, `ChoicePrompt` will fall back to `this.style`, which is defaulted to `ListStyle.auto`. This prevents `ListStyle.none` from being set by passing in PromptOptions.

## Specific Changes

```diff
- const choiceStyle: ListStyle = options.style || this.style;
+ const choiceStyle: ListStyle = options.style === 0 ? 0 : options.style || this.style;
```

I don't love the "complex" ternary and am willing to change it to something more like this, if preferred:

```ts
const choiceStyle;
if (options.style === 0) {
    choiceStyle = 0;
} else {
    choiceStyle: ListStyle = options.style || this.style;
}
```

## Testing

Added one additional test to ensure that `ListStyle.none` can be applied via `PromptOptions`.


compulim:
Thanks @mdrichardson. I have added @johnataylor and @stevengum as reviewers and pinged them accordingly.

PR #1381, opened 2019-10-31
Add LG/Expression library

Danieladu:


libraries/botbuilder-lg/README.MD
```diff
@@ -0,0 +1,188 @@
...
+
+## Grammar check and correction
+
+The current library does not include any capabilities for grammar check or correction.
+
+## Packages
+
+Packages for C# are available under the [BotBuilder MyGet feed][12]
+
+## Change Log
```
vishwacsena: We should clean up this page. These should not go here.

libraries/botbuilder-lg/README.MD
```diff
@@ -0,0 +1,188 @@
...
+- [API changes][2]: 
+    - Dropped FromFile and FromText methods in favor of AddFile and AddFiles. 
+    - Added ability to provide a delegate to externally resolve import references found in content. 
+- \[**NEW**\] Translate functionality in [MSLG CLI][23]
+
+### 4.5 PREVIEW
+- Initial preview release
+
+[1]:https://github.com/Microsoft/BotBuilder/blob/master/specs/botframework-activity/botframework-activity.md
+[2]:./docs/api-reference.md
```
vishwacsena: These links are not going to work unless we are moving all the docs over which we should not do.

lerna.json
```diff
@@ -8,6 +8,8 @@
     "libraries/botbuilder-core",
     "libraries/botbuilder-dialogs",
     "libraries/botbuilder-testing",
+    "libraries/botbuilder-lg",
+    "libraries/botframework-expressions",
```
vishwacsena: do we not need a ActivityGenerator? Will that be in lg for now?

libraries/botbuilder-lg/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "type": "git",
+    "url": "https://github.com/Microsoft/botbuilder-js.git"
+  },
+  "main": "./lib/index.js",
+  "typings": "./lib/index.d.ts",
+  "dependencies": {
+    "antlr4ts": "0.5.0-alpha.1",
+    "botframework-expressions": "4.1.6",
+    "lodash": "^4.17.11",
+    "uuid": "^3.3.3"
```
Stevenic: since the dialog system will have a dependency on these files we should ensure that they can be loaded into a browser.

libraries/botframework-expressions/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "@microsoft/recognizers-text-data-types-timex-expression": "^1.1.4",
+    "antlr4ts": "0.5.0-alpha.1",
+    "jspath": "^0.4.0",
+    "lodash": "^4.17.15",
+    "lru-cache": "^5.1.1",
+    "moment": "2.24.0",
+    "moment-timezone": "^0.5.25",
+    "xml2js": "^0.4.19",
+    "xmldom": "^0.1.27",
+    "xpath": "0.0.27"
```
Stevenic: same as other comment. we should ensure these will work in a browser.

lerna.json
boydc2014: It will be in as a follow-up PR into lg.

libraries/botframework-expressions/package.json
Danieladu: remove xml and xpath builtin function

PR #1384, opened 2019-11-05
Stevenic/4.6 dialog parity

Stevenic:
This PR adds parity with C#'s 4.6 changes for dialog.  One delta is that I've went ahead and folded the DialogStateManager into the dialog system as  I believe I was able to cleanup the initialization issues we wanted to resolve as part of R7. 

## Testing
I've added a basic set of unit tests for the DialogStateManager but more are needed.  We can hold this check-in if we want until those additional tests are added but I need to get the "future" branch unblocked so we can get to adaptive to parity.


libraries/botbuilder-dialogs/src/dialogContainer.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+import { Dialog } from './dialog';
+import { DialogSet } from './dialogSet';
+import { DialogContext } from './dialogContext';
+
+export abstract class DialogContainer<O extends object = {}> extends Dialog<O> {
+    /**
+     * The containers dialog set.
+     */
+    protected readonly dialogs = new DialogSet(undefined);
```
tomlm: >protected [](start = 4, length = 9)

public

libraries/botbuilder-dialogs/src/memory/dialogStateManager.ts
```diff
@@ -0,0 +1,522 @@
...
+            }
+
+            return this._config;
+        }
+    }
+
+    public set configuration(value: DialogStateManagerConfiguration) {
+        if (this.dialogContext.parent) {
+            this.dialogContext.parent.state.configuration = value;
+        } else {
```
tomlm: remove this

PR #1390, opened 2019-11-06
[Teams] Call defaultNextEvent after .invoke activity is processed

EricDahlvang:
Without this fix, this.onDialog of the TeamsActivityHandler will not be called if present, as it should be at the end of the turn.


EricDahlvang:
I agree it makes much more sense to use the .run method override to save state.  However, I don't understand why we can't do both.  Add this line, so our previous advice still works AND modify our docs to use the newly recommended approach of the .run method.  

What are the disadvantages to "call-onDialog-after-all-activities" for Invoke type activities?

stevengum:
That sounds good to me

PR #1401, opened 2019-11-12
add Activityfactory

Danieladu:
Add ActivityFactory, transfer from LG string output to activity


libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+        [ 'signincard', CardFactory.contentTypes.signinCard ],
+        [ 'oauthcard', CardFactory.contentTypes.oauthCard ],
+    ]);
+
+    public static CreateActivity(lgStringResult: string): Activity {
+        let lgStructuredResult: any;
+        try {
+            lgStructuredResult = JSON.parse(lgStringResult);
+        } catch (error) {
+            return this.buildActivityFromText(lgStructuredResult.trim());
```
Stevenic: >lgStructuredResult [](start = 46, length = 18)

Shouldn't this be lgStringResult?

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+
+    private static buildActivityFromLGStructuredResult(lgJObj: any): Activity {
+        let activity: any = {};
+
+        const type: string = this.getStructureType(lgJObj);
+        if (ActivityFactory.GenericCardTypeMapping.has(type)) {
+            const attachment: Attachment = this.getAttachment(lgJObj);
+            if (attachment !== undefined) {
+                activity = MessageFactory.attachment(attachment);
+            }
```
Stevenic: >} [](start = 12, length = 1)

Do you really expect to have an empty activity returned if you can't build an attachment? Seems like this should be an error.

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+
+        return this.buildActivityFromLGStructuredResult(lgStructuredResult);
+    }
+
+    private static buildActivityFromText(text: string): Activity {
+        return MessageFactory.text(text, text) as Activity;
+    }
+
+    private static buildActivityFromLGStructuredResult(lgJObj: any): Activity {
+        let activity: any = {};
```
Stevenic: >let activity: any = {} [](start = 8, length = 22)

This should probably be "let activity: Activity;".  That will initialize it to be undefined but it will be properly typed.

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+        [ 'herocard', CardFactory.contentTypes.heroCard ],
+        [ 'thumbnailcard', CardFactory.contentTypes.thumbnailCard ],
+        [ 'audiocard', CardFactory.contentTypes.audioCard ],
+        [ 'videocard', CardFactory.contentTypes.videoCard ],
+        [ 'animationcard', CardFactory.contentTypes.animationCard ],
+        [ 'signincard', CardFactory.contentTypes.signinCard ],
+        [ 'oauthcard', CardFactory.contentTypes.oauthCard ],
+    ]);
+
+    public static CreateActivity(lgStringResult: string): Activity {
```
Stevenic: >Activity [](start = 58, length = 8)

These should all be "Partial<Activity>".

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+            lgStructuredResult = JSON.parse(lgStringResult);
+        } catch (error) {
+            return this.buildActivityFromText(lgStructuredResult.trim());
+        }
+
+        return this.buildActivityFromLGStructuredResult(lgStructuredResult);
+    }
+
+    private static buildActivityFromText(text: string): Activity {
+        return MessageFactory.text(text, text) as Activity;
```
Stevenic: >as Activity [](start = 47, length = 11)

If your return type is "Partial<Activity>" you won't need this cast

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+                case '$type':
+                    break;
+                case 'name':
+                    activity.name = value.toString();
+                    break;
+                case 'value':
+                    activity.value = value.toString();
+                    break;
+                default:
+                    console.log(`Skipping unknown activity property ${ property }`);
```
Stevenic: There are potentially other valid property types. Do we really want to restrict the properties they can add to the activity?

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+                    activity.inputHint = value.toString();
+                    break;
+                case 'attachments':
+                    activity.attachments = this.getAttachments(value);
+                    break;
+                case 'suggestedactions':
+                    activity.suggestedActions = this.getSuggestions(value);
+                    break;
+                case 'attachmentlayout':
+                    activity.AttachmentLayout = value.toString();
```
Stevenic: >AttachmentLayout [](start = 29, length = 16)

Case is wrong... That's another reason you want activity to be strongly typed and not an any

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+        return {
+            contentType: type,
+            content: card
+        };
+    }
+
+    private static normalizedToList(item: any): any[] {
+        if (item === undefined) {
+            return [];
+        } else if (item instanceof Array){
```
Stevenic: >item instanceof Array [](start = 19, length = 21)

use Array.isArray() instead

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+            to: []
+        };
+
+        let actions: any[] = this.normalizedToList(value);
+
+        for (const action of actions) {
+            if (typeof action === 'string') {
+                const actionStr: string = action.trim();
+                suggestedActions.actions.push({
+                    type: ActionTypes.MessageBack,
```
Stevenic: >MessageBack [](start = 38, length = 11)

Unless something has changed recently... MessageBack isn't supported across by all channels. IMBack is safer.

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+                    text: actionStr,
+                    value: actionStr
+                });
+            } else {
+                const cardAction: CardAction = this.getCardAction('', action);
+                if (cardAction !== undefined) {
+                    suggestedActions.actions.push(cardAction);
+                }
+            }
+        }
```
Stevenic: This whole block could be replaced with "suggestedActions.actions = CardFactory.actions(actions);"

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    private static getButtons(cardType: string, value: any): CardAction[] {
+        let buttons: any[] = [];
+        let actions: any[] = this.normalizedToList(value);
+
+        for (const action of actions) {
+            if (typeof action === 'string') {
+                const actionStr = action.trim();
+                if (cardType === CardFactory.contentTypes.signinCard ||cardType === CardFactory.contentTypes.oauthCard ) {
```
Stevenic: >cardType === CardFactory.contentTypes.signinCard ||cardType === CardFactory.contentTypes.oauthCard  [](start = 20, length = 99)

This logic would make every button for a sign-in card a sign-in button. That doesn't seem desirable.

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+        }
+
+        return buttons;
+    }
+
+    private static getCardAction(cardType: string, cardActionJObj: any): CardAction {
+        const type: string = this.getStructureType(cardActionJObj);
+        let cardAction: any = {};
+
+        if (cardType === CardFactory.contentTypes.signinCard || cardType === CardFactory.contentTypes.oauthCard) {
```
Stevenic: >cardType === CardFactory.contentTypes.signinCard || cardType === CardFactory.contentTypes.oauthCard [](start = 12, length = 99)

Again... This would make every button or suggested action for a signin-card a signin button. That doesn't seem desirable. You can't add a "cancel" or "help" suggested action for instance.

libraries/botbuilder-lg/src/activityFactory.ts
```diff
@@ -0,0 +1,361 @@
...
+
+        return attachments;
+    }
+
+    private static getAttachment(lgJObj: any): Attachment {
+        let attachment: any = {};
+        const type: string = this.getStructureType(lgJObj);
+        if (ActivityFactory.GenericCardTypeMapping.has(type)) {
+            attachment = this.getCardAttachment(ActivityFactory.GenericCardTypeMapping.get(type), lgJObj);
+        } else if(type === 'adaptivecard') {
```
Stevenic: >type [](start = 18, length = 4)

is this always lower-cased? I didn't see anything in getStructureType() that does a lower case but maybe it's always lower case coming back from LG.

Danieladu: yes, there is a 'toLowerCase()' in getStructuredType()

PR #1403, opened 2019-11-12
add more function in JS

Danieladu:
add the missing functions in c sharp

### DIFF
--- a/libraries/botframework-expressions/src/expression.ts
+++ b/libraries/botframework-expressions/src/expression.ts
@@ -8,7 +8,7 @@
  */
 import { BuiltInFunctions } from './builtInFunction';
 import { Constant } from './constant';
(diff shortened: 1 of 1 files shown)


libraries/botframework-expressions/src/expression.ts
```diff
@@ -88,6 +88,80 @@ export class Expression {
         return expr;
     }
 
+    /**
+     * Construct an expression from a EvaluateExpressionDelegate
+     * @param func Function to create an expression from.
+     */
+    public static lambaExpression(func: EvaluateExpressionDelegate): Expression {
```
Stevenic: >lambaExpression [](start = 18, length = 15)

These all need unit tests

Danieladu:
close it temporarily, because these functions are just used by trigger tree library.

PR #1409, opened 2019-11-13
[Fix #1404, Transcripts] Fix for TranscriptLoggerMiddleware not checking responses type

### LINKED ISSUE #1404: [Bug, Transcripts] TranscriptLoggerMiddleware should check for responses before checking responses.length
## Versions
package version of the SDK: 4.6.0
nodejs version are you using: 12.13.0 and 10.17.0

## Describe the bug
Below I shared the image, every message throw error. 
error in botbuilder-core\lib\transcriptLogger.js:56:43 line package. 

I'm testing inside the ms-team channel. 

## Screenshots
image

[bug]

stevengum:
Fixes #1404

## Specific Changes
  - Creates a pointer in TranscriptLoggerMiddleware's SendActivitiesHandler which ensures that the return value from the pipeline is continuously passed through
  - If the return value is not an array, then create an array with the same number of empty ResourceResponses as there were outgoing activities.
 - More tests

## Testing
Add new tests for:
- An Adapter not returning the ResourceResponses to the TranscriptLoggerMiddleware (via TurnContext)
- Another SendActivitiesHandler before the TranscriptLoggerMiddleware does not return the value from `next()`
- Another SendActivitiesHandler before the TranscriptLoggerMiddleware returns a value that is not an Array.
- The actual return value from a SendActivitiesHandler before TranscriptLoggerMiddleware is pushed through the pipeline to any SendActivitiesHandlers _after_ the TranscriptLoggerMiddleware

@cleemullins FYI

### DIFF
--- a/libraries/botbuilder-core/src/transcriptLogger.ts
+++ b/libraries/botbuilder-core/src/transcriptLogger.ts
@@ -47,48 +47,32 @@ export class TranscriptLoggerMiddleware implements Middleware {
 
         // hook up onSend pipeline
         context.onSendActivities(async (ctx: TurnContext, activities: Partial<Activity>[], next2: () => Promise<ResourceResponse[]>) => {
-            // Run full pipeline and create a pointer to the return value from the pipeline.
-            let nextValue: ResourceResponse[] = await next2();
-            let responses: ResourceResponse[] = nextValue;
-
-            // If the return value from the pipeline is not an Array, point `responses` to an array of empty ResourceResponses.
-            // This allows for the following:
-            //      1. Allows the transcript logging to progress and not error out.
-            //         See https://github.com/microsoft/botbuilder-js/issues/1404
-            //      1. This prevents TranscriptLoggers from throwing an error if no activity.id is set.
-            //         See https://github.com/microsoft/botbuilder-js/issues/1122
-            //      2. By mocking an array of ResourceResponses, we still return the value from `next2()`.
-            //         This prevents any pre-4.6.0 bots with multiple SendActivityHandlers that rely
-            //         on values that are not an array of ResourceResponses.
-            if (!Array.isArray(responses)) {
-                responses = activities.map(() => {
-                    return {} as ResourceResponse;
-                });
-            }
+            // Run full pipeline.
+            const responses: ResourceResponse[] = await next2();
 
             activities.map((a: Partial<Activity>, index: number) => {
                 const clonedActivity = this.cloneActivity(a);
-                if (index < responses.length) {
-                    if (!clonedActivity.id) {
-                        clonedActivity.id = responses[index].id;
+                if (!clonedActivity.id) {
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-core/src/transcriptLogger.ts
```diff
@@ -47,8 +47,24 @@ export class TranscriptLoggerMiddleware implements Middleware {
 
         // hook up onSend pipeline
         context.onSendActivities(async (ctx: TurnContext, activities: Partial<Activity>[], next2: () => Promise<ResourceResponse[]>) => {
-            // run full pipeline
-            const responses: ResourceResponse[] = await next2();
+            // Run full pipeline and create a pointer to the return value from the pipeline.
+            let nextValue: ResourceResponse[] = await next2();
```
christopheranderson: Shouldn't these be consts?

libraries/botbuilder-core/src/transcriptLogger.ts
```diff
@@ -47,8 +47,24 @@ export class TranscriptLoggerMiddleware implements Middleware {
...
+            // If the return value from the pipeline is not an Array, point `responses` to an array of empty ResourceResponses.
+            // This allows for the following:
+            //      1. Allows the transcript logging to progress and not error out.
+            //         See https://github.com/microsoft/botbuilder-js/issues/1404
+            //      1. This prevents TranscriptLoggers from throwing an error if no activity.id is set.
+            //         See https://github.com/microsoft/botbuilder-js/issues/1122
+            //      2. By mocking an array of ResourceResponses, we still return the value from `next2()`.
+            //         This prevents any pre-4.6.0 bots with multiple SendActivityHandlers that rely
+            //         on values that are not an array of ResourceResponses.
+            if (!Array.isArray(responses)) {
```
christopheranderson: why isn't this fix as simple as adding the isArray check on line 71?

libraries/botbuilder-core/src/transcriptLogger.ts
stevengum: nextValue should definitely be a const, that was a mistake on my part

libraries/botbuilder-core/src/transcriptLogger.ts
stevengum: The reason for the mocked ResourceResponse[] is to make sure that the TranscriptLoggerMiddleware is still able to successfully log the activities . Without an Activity ID the JS TransctipStores would blow up; by creating the mock, we enter the logic where a fake activity is can be generated for the logged activities if

1. The cloned activity does not already have an Id
2. The resourceresponse did not have an Id

christopheranderson: But if I follow this logic correctly, if someone returns a responses with a length of 1 and activity is length of 2, then you'll still have missing ids, right?

PR #1415, opened 2019-11-14
[Adaptive] actions refactor to C# version

chon219:
## Description

Refactored actions in botbuilder-dialogs-adaptive

## Specific Changes

  - 
  -
  -


libraries/botbuilder-dialogs/src/dialog.ts
```diff
@@ -417,4 +417,63 @@ export abstract class Dialog<O extends object = {}> extends Configurable {
...
+        if (dc.stack.length > 0 || !dc.parent) {
+            return await dc.cancelAllDialogs();
+        } else {
+           const turnResult = await dc.parent.cancelAllDialogs();
+           turnResult.parentEnded = true;
+           return turnResult;
+        }
+    }
+
+    private popCommands(dc: DialogContext): void {
```
Stevenic: >popCommands [](start = 12, length = 11)

This shouldn't be needed as we removed the concept of inherited memory.  This may be on the C# side of things but should be removed from there as well.

libraries/botbuilder-dialogs/src/dialog.ts
```diff
@@ -417,4 +417,63 @@ export abstract class Dialog<O extends object = {}> extends Configurable {
...
+        this.popCommands(dc);
+        if (dc.stack.length > 0 || !dc.parent) {
+            return await dc.cancelAllDialogs();
+        } else {
+           const turnResult = await dc.parent.cancelAllDialogs();
+           turnResult.parentEnded = true;
+           return turnResult;
+        }
+    }
+
```
Stevenic: These should be moved up to the actions where they're needed. Same goes on the C# side of things.

libraries/botbuilder-dialogs-adaptive/src/actions/case.ts
```diff
@@ -0,0 +1,29 @@
+import { Dialog } from 'botbuilder-dialogs';
+import { Expression, Constant } from 'botframework-expressions';
+export class Case
```
Stevenic: small nit... add a space and comments.

libraries/botbuilder-dialogs-adaptive/src/actions/httpRequest.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import { DialogTurnResult, DialogConfiguration, DialogContext, DialogCommand, Dialog } from 'botbuilder-dialogs';
+import { Activity } from 'botbuilder-core';
+import { ExpressionProperty, ExpressionPropertyValue } from '../expressionProperty';
+import * as rp from 'request-promise-native';
+
```
Stevenic: we should be using node-fetch everywhere as it can be easily polyfilled to run on a browser

Stevenic:
>  * @module botbuilder-planning

these should all change to "botbuilder-dialogs-adaptive"

---
Refers to: libraries/botbuilder-dialogs-adaptive/src/actions/logAction.ts:2 in 7e87733. [](commit_id = 7e87733941190fb2819a465a583ec57d671a2193, deletion_comment = False)

libraries/botbuilder-dialogs-adaptive/src/actions/deleteProperty.ts
```diff
@@ -37,7 +37,7 @@ export class DeleteProperty<O extends object = {}> extends DialogCommand<O> {
         return super.configure(config);
     }
 
-    public async onRunCommand(dc: DialogContext): Promise<DialogTurnResult> {
+    public async beginDialog(dc: DialogContext): Promise<DialogTurnResult> {
         dc.state.setValue(this.property, undefined);
```
Stevenic: > dc.state.setValue(this.property, undefined); [](start = 7, length = 45)

there is actually a dc.state.deleteValue() method you can call.

libraries/botbuilder-dialogs-adaptive/src/actions/httpRequest.ts
```diff
@@ -0,0 +1,191 @@
...
+        this.responseType = responseType;
+        this.resultProperty = resultProperty;
+    }
+
+    protected onComputeId(): string {
+        return `HttpRequest[${this.method} ${this.url}]`;
+    }
+
+    public configure(config: HttpRequestConfiguration): this {
+        return super.configure(config);
```
Stevenic: >return super.configure(config); [](start = 8, length = 31)

The base class doesn't know how to configure the expressions passed in. You'll need to add logic here to explicitly configure any expressions. Look at ifCondition.config() for an example.

PR #1433, opened 2019-11-20
[Streaming, Preview] rewrite useWebSocket for low-level usage, remove from processActivity

### LINKED ISSUE #1386: [Streaming] Low-level hook for initializing WebSocket connection
The current way to set up a WebSocket connection via the BotFrameworkAdapter is to use `processActivity(req, res, logic: async (context) => Promise<any>)` which may call `useWebSocket(req, res, logic: async (context) => Promise<any>)`.

The problem with the current implementation is that it ties useWebSocket() to the [restify](https://www.npmjs.com/package/restify) framework, which helpfully exposes the `socket` and `head` via [`ServerUpgradeResponse.claimUpgrade()`](https://github.com/restify/node-restify/blob/master/lib/upgrade.js#L272) method.

Other web service frameworks do not offer the same helper code.

The one shared feature most Node.js Web libraries offer is a way to listen for the [`"upgrade"`](https://nodejs.org/api/http.html#http_event_upgrade_1) event emitted from the [`http.Server`](https://nodejs.org/api/http.html#http_class_http_server).

___

A low level entry-point for `useWebSockets()`. The updated signature may look like the following:

```ts
public async useWebSockets(req: http.IncomingMessage, socket: net.Socket, head: Buffer, logic: (context: TurnContext) => Promise<void>): Promise<void>
```

The current preview high-level entry-point `BotFrameworkAdapter.processActivity()` would not be used for WebSockets, only for regular HTTP requests.

___

**Additional notes**
Upgrade requests for WebSockets from Direct Line Speech only come in with GET requests. In the code, we would begin the authentication process if it's a valid Upgrade request using the GET method.

Requests with non-GET methods would simply pass through `BotFrameworkAdapter.useWebSockets()`

___

@christopheranderson, @carlosscastro, @DDEfromOR

stevengum:
Fixes #1386

## Description
Change entry point for WebSocket Streaming connections from `processActivity()` to only `useWebSocket()`.

useWebSocket() also no longer relies on `restify` to expose the socket and head for WebSocket consumption. As such, it is now a method that should be called in a handler for the ['upgrade'](https://nodejs.org/api/http.html#http_event_upgrade_1) event. (See **Testing** for more information)

useWebSocket() continues to perform the authentication of the WebSocket Upgrade request from the client, and will send back the correct status code and close the socket connection if the `authenticateConnection()` call fails.

## Specific Changes
  - BotFrameworkAdapter.useWebSocket() signature change
    - **from** `public async useWebSocket(req: WebRequest, res: WebResponse, logic: (context: TurnContext) => Promise<any>): Promise<void>`
    - **to** `public async useWebSocket(req: IncomingMessage, socket: Socket, head: Buffer, logic: (context: TurnContext) => Promise<any>): Promise<void>`
  - Remove `restify` dependencies in BotFrameworkAdapter
  - BotFrameworkAdapter.useWebSocket() now sends the correct status code back to the channel when the Upgrade request fails authentication
___
  - Testing:
    - Added test for `BotFrameworkAdapter.useWebSocket()` sad-path via the `this.authenticateConnection()` call.
    - Moved Helper classes out of `botFrameworkAdapterStreaming.test.js` and into separate files
    - Cleaned up tests
    - Finished the removal of using one helper class to cover both HTTP and Streaming Requests

## Testing
Manually tested changes to BotFrameworkAdapter.useWebSocket() with 3 DLS clients concurrently. 

The following code snippet shows the relevant information (i.e. the handling of the `'upgrade'` event from the HTTP Server); but is otherwise derived from the Node.js 02.echo-bot sample.

```js
// Create HTTP server.
const server = restify.createServer();
server.listen(process.env.port || process.env.PORT || 3978, () => {
    console.log(`\n${ server.name } listening to ${ server.url }`);
});

// BotFrameworkAdapterSettings configuration object
// Must have a valid appId and appPassword to use Direct Line Speech
const adapterSettings = {
    appId: process.env.MicrosoftAppId,
    appPassword: process.env.MicrosoftAppPassword
};

const { EchoBot } = require('./bot');
const myBot = new EchoBot();

// Listen for Upgrade requests to the same route to accept Upgrade requests for Streaming.
server.on('upgrade', async (req, socket, head) => {
    // Create an adapter scoped to this WebSocket connection to allow storing session data.
    const streamingAdapter = new BotFrameworkAdapter({
        ...adapterSettings,
        enableWebSockets: true
    });

    try {
        await streamingAdapter.useWebSocket(req, socket, head, async (context) => {
            // After connecting via WebSocket, run this logic for every request sent over
            // the WebSocket connection.
            await myBot.run(context);
        });
    } catch (err) {
        console.error(err);
    }
});
```

### DIFF
--- a/libraries/botbuilder/src/botFrameworkAdapter.ts
+++ b/libraries/botbuilder/src/botFrameworkAdapter.ts
@@ -135,12 +135,7 @@ export interface BotFrameworkAdapterSettings {
     channelService?: string;
 
     /**
-     * Optional. The option to determine if this adapter accepts WebSocket connections
-     */
-    enableWebSockets?: boolean;
-
-    /**
-     * Optional. Used to pass in a NodeWebSocketFactoryBase instance. Allows bot to accept WebSocket connections.
+     * Optional. Used to pass in a NodeWebSocketFactoryBase instance.
      */
     webSocketFactory?: NodeWebSocketFactoryBase;  
 }
@@ -268,12 +263,7 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
             this.credentials.oAuthScope = GovernmentConstants.ToChannelFromBotOAuthScope;
         }
 
-        // If the developer wants to use WebSockets, but didn't provide a WebSocketFactory,
-        // create a NodeWebSocketFactory.
-        if (this.settings.enableWebSockets && !this.settings.webSocketFactory) {
-            this.webSocketFactory = new NodeWebSocketFactory();
-        }
-
+        // If a NodeWebSocketFactoryBase was passed in, set it on the BotFrameworkAdapter.
         if (this.settings.webSocketFactory) {
             this.webSocketFactory = this.settings.webSocketFactory;
         }
@@ -1145,13 +1135,14 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
     /**
      * Process the initial request to establish a long lived connection via a streaming server.
      * @param req The connection request.
-     * @param res The response sent on error or connection termination.
-     * @param logic The logic that will handle incoming requests.
+     * @param socket The raw socket connection between the bot (server) and channel/caller (client).
+     * @param head The first packet of the upgraded stream.
+     * @param logic The logic that handles incoming streaming requests for the lifetime of the WebSocket connection.
      */
     public async useWebSocket(req: WebRequest, socket: INodeSocket, head: INodeBuffer, logic: (context: TurnContext) => Promise<any>): Promise<void> {   
-        if (!this.webSocketFactory || !this.webSocketFactory.createWebSocket) {
-            throw new Error('BotFrameworkAdapter must have a WebSocketFactory in order to support streaming.');
-        }
+        // Use the provided NodeWebSocketFactoryBase on BotFrameworkAdapter construction,
+        // otherwise create a new NodeWebSocketFactory.
+        const webSocketFactory = this.webSocketFactory || new NodeWebSocketFactory();
 
         if (!logic) {
             throw new Error('Streaming logic needs to be provided to `useWebSocket`');
@@ -1176,7 +1167,7 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
             throw err;
         }
 
-        const nodeWebSocket = await this.webSocketFactory.createWebSocket(req, socket, head);
+        const nodeWebSocket = await webSocketFactory.createWebSocket(req, socket, head);
 
         await this.startWebSocket(nodeWebSocket);
     }
--- a/libraries/botbuilder/src/botFrameworkAdapter.ts
+++ b/libraries/botbuilder/src/botFrameworkAdapter.ts
@@ -6,7 +6,8 @@
  * Licensed under the MIT License.
  */
 
-import { IncomingMessage } from 'http';
+import { IncomingMessage, STATUS_CODES } from 'http';
+import { Socket } from 'net';
 import * as os from 'os';
 
 import { Activity, ActivityTypes, BotAdapter, BotCallbackHandlerKey, ChannelAccount, ConversationAccount, ConversationParameters, ConversationReference, ConversationsResult, IUserTokenProvider, ResourceResponse, TokenResponse, TurnContext } from 'botbuilder-core';
@@ -755,9 +756,6 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
      * ```
      */
     public async processActivity(req: WebRequest, res: WebResponse, logic: (context: TurnContext) => Promise<any>): Promise<void> {
-        if (this.settings.enableWebSockets && req.method === GET && (req.headers.Upgrade || req.headers.upgrade)) {
-            return this.useWebSocket(req, res, logic);
-        }
         
         let body: any;
         let status: number;
@@ -1151,36 +1149,37 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
      * @param res The response sent on error or connection termination.
      * @param logic The logic that will handle incoming requests.
      */
-    public async useWebSocket(req: WebRequest, res: WebResponse, logic: (context: TurnContext) => Promise<any>): Promise<void> {   
-        if (!logic) {
-            throw new Error('Streaming logic needs to be provided to `useWebSocket`');
-        }
-
+    public async useWebSocket(req: IncomingMessage, socket: Socket, head: Buffer, logic: (context: TurnContext) => Promise<any>): Promise<void> {   
         if (!this.webSocketFactory || !this.webSocketFactory.createWebSocket) {
             throw new Error('BotFrameworkAdapter must have a WebSocketFactory in order to support streaming.');
         }
 
-        this.logic = logic;
-
-        // Restify-specific check.
-        if (typeof((res as any).claimUpgrade) !== 'function') {
-            throw new Error('ClaimUpgrade is required for creating WebSocket connection.');
+        if (!logic) {
+            throw new Error('Streaming logic needs to be provided to `useWebSocket`');
         }
+
+        this.logic = logic;
 
         try {
             await this.authenticateConnection(req, this.settings.channelService);
         } catch (err) {
-            // Set the correct status code for the socket to send back to the channel.
-            res.status(StatusCodes.UNAUTHORIZED);
-            res.send(err.message);
+            // If the authenticateConnection call fails, send back the correct error code and close
+            // the connection.
+            if (typeof(err.message) === 'string' && err.message.toLowerCase().startsWith('unauthorized')) {
+                abortWebSocketUpgrade(socket, 401);
+            } else if (typeof(err.message) === 'string' && err.message.toLowerCase().startsWith(`'authheader'`)) {
+                abortWebSocketUpgrade(socket, 400);
+            } else {
+                abortWebSocketUpgrade(socket, 500);
+            }
+
             // Re-throw the error so the developer will know what occurred.
             throw err;
         }
 
-        const upgrade = (res as any).claimUpgrade();
-        const socket = await this.webSocketFactory.createWebSocket(req as IncomingMessage, upgrade.socket, upgrade.head);
+        const nodeWebSocket = await this.webSocketFactory.createWebSocket(req, socket, head);
 
-        await this.startWebSocket(socket);
+        await this.startWebSocket(nodeWebSocket);
     }
 
     private async authenticateConnection(req: WebRequest, channelService?: string): Promise<void> {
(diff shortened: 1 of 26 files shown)


libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1151,36 +1149,37 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
-            res.send(err.message);
+            // If the authenticateConnection call fails, send back the correct error code and close
+            // the connection.
+            if (typeof(err.message) === 'string' && err.message.toLowerCase().startsWith('unauthorized')) {
+                abortWebSocketUpgrade(socket, 401);
+            } else if (typeof(err.message) === 'string' && err.message.toLowerCase().startsWith(`'authheader'`)) {
+                abortWebSocketUpgrade(socket, 400);
+            } else {
+                abortWebSocketUpgrade(socket, 500);
+            }
```
stevengum: @carlosscastro I'm not happy with this code, but I'm not sure if there are better steps that can be taken here.

Node.js does not have HTTP Status Code-specific errors, so our underlying auth layers throw `new Error` with different message strings. Due to this, the code parses the message to attempt to determine the correct status code to send back before destroying the socket connection.

I have a test that covers the 401 Unauthorized route, and tested the other two routes manually.

libraries/botbuilder/tests/streaming/mockNetSocket.js
```diff
@@ -0,0 +1,18 @@
...
+
+class MockNetSocket {
+    constructor(readable = true, writable = true) {
+        this.readable = readable;
+        this.writable = writable;
+    }
+
+    write(response) { }
+
+    destroy(err) { }
```
stevengum: These two methods are for spying on during the authentication portion of `useWebSocket()`. The static method `MockNetSocket.createNonSuccessResponse()` is for testing the 401 scenario in useWebSocket().

libraries/botbuilder/tests/streaming/botFrameworkAdapterStreaming.test.js
```diff
@@ -298,53 +222,48 @@ describe('BotFrameworkAdapter Streaming tests', () => {
...
-            let mw = {
+        it('returns a 500 when BotFrameworkAdapter.logic is not callable', async () => {
+            const adapter = new BotFrameworkAdapter();
+            const request = new MockStreamingRequest();
+
+            const response = await adapter.processRequest(request);
+            expect(response.statusCode).to.equal(500);
+        });
+
+        it('returns a 500 and calls middleware when BotFrameworkAdapter.logic is not callable', async () => {
```
stevengum: BotAdapter.runMiddleware will run all middleware before attempting to call the application logic. This test could be stored in the previous test, but I separated the two to make the tests more explicit and preserve the "intent" of the previous test.

libraries/botbuilder/tests/streaming/botFrameworkAdapterStreaming.test.js
```diff
@@ -298,53 +222,48 @@ describe('BotFrameworkAdapter Streaming tests', () => {
             expect(response.statusCode).to.equal(501);
         });
 
-        it('returns a 500 when bot can not run', async () => {
-            const MiddleWare = require('botbuilder-core');
-            const bot = {};
-            let mw = {
```
stevengum: The logic in this test was out of date, as we no longer pass in a MiddlewareSet to the constructor. The 500 was probably being thrown due to `BotFrameworkAdapter.logic` being missing, so we should harden the rewritten test and the [test](https://github.com/microsoft/botbuilder-js/pull/1433/files#r348312596) that follows it.

stevengum:
I cherry picked the 4.6 branch changes pinning the TypeScript version and fixing the versioning script used on the CI machines.

The latest commit, b08c9fa, removes the unnecessary enableWebSockets flag. 

The flag was deemed unnecessary because now the developer explicitly opts in to using WebSockets (even more so than before).

Because each BotFrameworkAdapter is scoped to the lifetime of the WebSocket connection, we don’t need to create and bind a websocketfactory to every BotFrameworkAdapter instance.

libraries/botbuilder/src/botFrameworkAdapter.ts
carlosscastro: Ah, that hurts, but I understand. Can we improve the error reporting in the auth layers instead of here? Maybe we could create an abstraction that processes error messages and creates well defined errors + codes, and then the auth layer uses that in order to throw richer errors. Sounds like something that may not be possible without breaking backward compat though.

If it is not possible to move this logic to the auth layer, then I guess we are out of choices.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1148,39 +1136,41 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
-            res.status(StatusCodes.UNAUTHORIZED);
-            res.send(err.message);
+            // If the authenticateConnection call fails, send back the correct error code and close
+            // the connection.
+            if (typeof(err.message) === 'string' && err.message.toLowerCase().startsWith('unauthorized')) {
+                abortWebSocketUpgrade(socket, 401);
+            } else if (typeof(err.message) === 'string' && err.message.toLowerCase().startsWith(`'authheader'`)) {
+                abortWebSocketUpgrade(socket, 400);
+            } else {
+                abortWebSocketUpgrade(socket, 500);
```
carlosscastro: Why dont we send the error message in the response? Seems impossible to detect the error without some more info.

carlosscastro: Or are they usually cryptic?

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1148,39 +1136,41 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
             await this.authenticateConnection(req, this.settings.channelService);
         } catch (err) {
-            // Set the correct status code for the socket to send back to the channel.
-            res.status(StatusCodes.UNAUTHORIZED);
-            res.send(err.message);
+            // If the authenticateConnection call fails, send back the correct error code and close
+            // the connection.
+            if (typeof(err.message) === 'string' && err.message.toLowerCase().startsWith('unauthorized')) {
+                abortWebSocketUpgrade(socket, 401);
+            } else if (typeof(err.message) === 'string' && err.message.toLowerCase().startsWith(`'authheader'`)) {
```
carlosscastro: Just to confirm - here you used the quotes `'authheader'` on purpose right?

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1303,4 +1293,13 @@ function delay(timeout: number): Promise<void> {
     return new Promise((resolve) => {
         setTimeout(resolve, timeout);
     });
+}
+
+function abortWebSocketUpgrade(socket: Socket, code: number) {
+    if (socket.writable) {
+        const connectionHeader = `Connection: 'close'\r\n`;
+        socket.write(`HTTP/1.1 ${code} ${STATUS_CODES[code]}\r\n${connectionHeader}\r\n`);
```
carlosscastro: Just checking - could we have any cross platform issues with the `\r\n`?

stevengum: This should work on the different platforms, but I can check on a different one.

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: That's correct, but I _think_ we're going to try and change all of this. Maybe not in this PR though...

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: Filed https://github.com/microsoft/botbuilder-js/issues/1445 for this. I've tentatively labeled it as a P2 for R7. Since this is internal, I don't want to block on this specific code.

I'd like to get this in so that I can have the package available for the WebChat DLS bug bash tomorrow.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1148,39 +1136,41 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
      * Process the initial request to establish a long lived connection via a streaming server.
      * @param req The connection request.
-     * @param res The response sent on error or connection termination.
-     * @param logic The logic that will handle incoming requests.
+     * @param socket The raw socket connection between the bot (server) and channel/caller (client).
+     * @param head The first packet of the upgraded stream.
+     * @param logic The logic that handles incoming streaming requests for the lifetime of the WebSocket connection.
      */
-    public async useWebSocket(req: WebRequest, res: WebResponse, logic: (context: TurnContext) => Promise<any>): Promise<void> {   
+    public async useWebSocket(req: IncomingMessage, socket: Socket, head: Buffer, logic: (context: TurnContext) => Promise<any>): Promise<void> {
```
christopheranderson: This is adding a public type dependency on Socket and Buffer. This means if there is a breaking change to these APIs between Node.js versions, we may run into type issues.

If we want to isolate ourselves from this, we can create interfaces which are compatible with the Socket and Buffer types.

libraries/botbuilder/src/botFrameworkAdapter.ts
DDEfromOR: What about HTTP/2 calls to upgrade to WebSockets? Can we dynamically populate the HTTP version in this response? I'm not sure it matters, but it may lead to confusing debugging and AFAIK we don't specifically NOT support HTTP/2 anywhere. (The ABS channels are all HTTP/1.1, but there's nothing stopping an outside channel from using HTTP/2.) Not a ship stopper, but food for thought.

DDEfromOR: Also it looks like we're never sending the actual error message back now, is that right? That will make debugging problematic, especially if the code is a 500.

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: Our libraries actually don't use the `head`, we just pass it through to `ws`. 

Do we still need an interface for `head`/Buffer or should I just change the type to `any`?

christopheranderson: We can keep `Buffer` for now, and always swap to a compat interface or `any` later. We can't switch from `any` to anything else without it being breaking, so we should avoid that unless we're out of options.

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: - Requests over HTTP/2 to bootstrap a WebSocket connection need to use the `CONNECT` method which comes over the `'connect'` event in the `http.Server` and `http2.Http2Server`.
- We can dynamically create the HTTP version via the `httpVersion` off of the `IncomingMessage`.
- https://github.com/microsoft/botbuilder-js/issues/1460 was filed to address the shared concern about sending the error message back.

PR #1437, opened 2019-11-21
[4.6.x] Pin TypeScript devDependency in each library. (Fixes #1436)

### LINKED ISSUE #1436: [4.6.1 Breaking Change] 4.6.1 libraries were transpiled with TypeScript 3.7.2
## Versions
What package version of the SDK are you using. **4.6.1**

## Describe the bug
4.6.1 was transpiled with `typescript@3.7.2` which led to unintentional **breaking changes** for TypeScript consumers of the botbuilder-js libraries.

The breaking changes revolve around the inclusion of [getters and setters in the `.d.ts` files](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#class-field-mitigations) in TypeScript 3.7.0.

As such, we need to cut a new patch release (4.6.2) with a lower version of TypeScript.

Steps to resolution:
1. Remove un-pinned `typescript` devDependency from root package.json
2. Pin `typescript` devDependency (**pre-3.7.0**) in each library's package.json (`botbuilder`, etc.)
3. Release 4.6.2 libraries with pinned new devDependency

[bug]

FYI @juanar @Batta32 @enzocanoo this is the bug the skills team was encountering in the TypeScript BF-Solutions work.

stevengum:
Fixes #1436

## Description
Pins `typescript` as devDependency in each library to fix bug of transpiling 4.6.1 libraries with 3.7.2.

## Specific Changes
  - Remove `typescript@^3.5.2` devDependency from root package.json
  - Create `typescript@3.5.3` devDependency in each library
  - Change botbuilder, botbuilder-ai and botbuilder-dialogs dependencies in `transcripts/` to instead be `4.1.6`, not `^4.2.1` which led to `4.6.1` botbuilder-js libraries being hoisted at the root
  - Fix uncaught import bug in `botbuilder-core` `SkypeMentionNormalizeMiddleware`
    - The class was importing from `botbuilder-core`, the library it is in

### DIFF
--- a/libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
+++ b/libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
@@ -5,7 +5,9 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-import { Activity, Middleware, TurnContext } from 'botbuilder-core';
+import { Activity } from 'botframework-schema';
+import { Middleware } from './middlewareSet';
+import { TurnContext } from './turnContext';
 
 
 /**
--- a/transcripts/package.json
+++ b/transcripts/package.json
@@ -10,9 +10,9 @@
     "dependencies": {
         "@types/node": "^10.12.18",
         "@types/restify": "^7.2.1",
-        "botbuilder": "^4.2.1",
-        "botbuilder-ai": "^4.2.1",
-        "botbuilder-dialogs": "^4.2.1"
+        "botbuilder": "4.1.6",
+        "botbuilder-ai": "4.1.6",
+        "botbuilder-dialogs": "4.1.6"
     },
     "devDependencies": {
         "mocha": "^5.2.0",
--- a/libraries/botbuilder-ai/package.json
+++ b/libraries/botbuilder-ai/package.json
@@ -38,7 +38,8 @@
     "nock": "^10.0.3",
     "nyc": "^11.4.1",
     "source-map-support": "^0.5.3",
-    "ts-node": "^4.1.0"
+    "ts-node": "^4.1.0",
+    "typescript": "3.5.2"
   },
   "scripts": {
     "test": "tsc && nyc mocha tests/",
--- a/libraries/botbuilder-ai/package.json
+++ b/libraries/botbuilder-ai/package.json
@@ -39,7 +39,7 @@
     "nyc": "^11.4.1",
     "source-map-support": "^0.5.3",
     "ts-node": "^4.1.0",
-    "typescript": "3.5.2"
+    "typescript": "3.5.3"
   },
   "scripts": {
     "test": "tsc && nyc mocha tests/",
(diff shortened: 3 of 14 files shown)


libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
```diff
@@ -5,7 +5,9 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-import { Activity, Middleware, TurnContext } from 'botbuilder-core';
```
cleemullins: Not sure why this change is here...

transcripts/package.json
```diff
@@ -10,9 +10,9 @@
     "dependencies": {
         "@types/node": "^10.12.18",
         "@types/restify": "^7.2.1",
-        "botbuilder": "^4.2.1",
-        "botbuilder-ai": "^4.2.1",
-        "botbuilder-dialogs": "^4.2.1"
+        "botbuilder": "4.1.6",
```
cleemullins: I don't understand this set of changes.

libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
stevengum: This is necessary because it was telling a class inside of botbuilder-core to look inside the botbuilder-core's node_modules/ for **botbuilder-core**, which makes no logical sense.

It is part of the reason that we found the bug regarding the typescript 3.7. 

By looking for botbuilder-core, it tried to resolve this dependency through the normal Node.js fashion, and traversed the path upward where it found `botbuilder-core@4.6.1` installed at the root node_modules/.

transcripts/package.json
stevengum: By having a different dependency version that was unpinned, `lerna bootstrap --hoist` was fetching this version and placing it at the root of the repository.

All `4.1.6` versions (e.g. the libraries and test bot) were being correctly symlinked, while this package.json was telling lerna to fetch `^4.2.1` from npm.

Without this unpinned version, the `SkypeMentionNormalizeMiddleware` class was not successfully building because `botbuilder-core` is not a dependency of `botbuilder-core`.

[See my previous comment.](https://github.com/microsoft/botbuilder-js/pull/1437#discussion_r349251015)

stevengum: > _Without this unpinned version, the SkypeMentionNormalizeMiddleware class was not successfully building because botbuilder-core is not a dependency of botbuilder-core._

To clarify, by setting the version here to 4.1.6, the dependencies in the transcripts/package.json were being symlinked.

This meant that there were no `botbuilder-core` dependencies at the root for the SkypeMentionNormalizeMiddleware class to resolve against, which caused the `botbuilder-core` library to not build.

libraries/botbuilder-core/src/skypeMentionNormalizeMiddleware.ts
christopheranderson: I also had to fix this in my rush branch. This is a good fix.

PR #1446, opened 2019-11-26
[Adaptive] implemented trigger conditions

chon219:
## Description

Implemented the rest of trigger conditions in botbuilder-dialogs-adaptive.

## Specific Changes

  - Implemented trigger conditions
  - Retired *addRule()* method in AdaptiveDialog (use *triggers.push()* instead)
  - Retired *actions* property in AdaptiveDialog (use *dialogs* instead)
  - Removed outdated samples
  - Implemented *getDependencies()* method in OnCondition to implement *DialogDependencies* interface


libraries/botbuilder-dialogs-adaptive/src/adaptiveDialog.ts
```diff
@@ -84,24 +72,19 @@ export class AdaptiveDialog<O extends object = {}> extends DialogContainer<O> {
...
-
         // Install each trigger actions
-        this.triggers.forEach((trigger) => {
-            trigger.actions.forEach((action) => this.dialogs.add(action));
-        });
+        for (let i = 0; i < this.triggers.length; i++) {
+            const trigger = this.triggers[i];
+            if (typeof ((trigger as any) as DialogDependencies).getDependencies == 'function') {
+                ((trigger as any) as DialogDependencies).getDependencies().forEach((child) => this.dialogs.add(child));
+            }
```
Stevenic: You don't need to do this. When a dialog is added it will recursively have it's dependencies added. The C# implementation is wrong and should be changed. The old code was correct.

libraries/botbuilder-dialogs-adaptive/src/actions/editActions.ts
```diff
@@ -20,7 +20,7 @@ export interface EditActionsConfiguration extends DialogConfiguration {
     actions?: Dialog[];
 }
 
-export class EditActions<O extends object = {}> extends Dialog<O> {
+export class EditActions<O extends object = {}> extends Dialog<O> implements DialogDependencies {
```
Stevenic: >implements DialogDependencies [](start = 66, length = 29)

good catch!

libraries/botbuilder-dialogs-adaptive/src/conditions/index.ts
```diff
@@ -5,8 +5,23 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-export * from './onDialogEvent';
 export * from './onCondition';
+export * from './onDialogEvent';
```
Stevenic: I try to list all of the exports in alphabetical order. That way you can more easily see if you're missing an export by just comparing the exports and ensure they're in the same order as VSCode has.

libraries/botbuilder-dialogs-adaptive/src/conditions/onCondition.ts
```diff
@@ -100,6 +100,25 @@ export class OnCondition {
...
+            dependencies.push(action);
+
+            // Automatically add any child dependencies the dialog might have
+            if (typeof ((action as any) as DialogDependencies).getDependencies == 'function') {
+                ((action as any) as DialogDependencies).getDependencies().forEach((child) => dependencies.push(child));
+            }
+        }
+
+        return dependencies;
+    }
```
Stevenic: All you need to do here is "return this.actions;"  When a Dialog is added to a DialogSet it will call getDependencies() recursively to get any of the actions dependencies and any of the dependencies dependencies.

PR #1451, opened 2019-11-27
Inputs Refactor, Fix HttpRequest, Add More Samples

zidaneymar:
Inputs Refactor, Fix HttpRequest, Add More Samples

## Description

1. InputsDialog two properties aligned with C#
2. Fix HttpRequest
3. Add More Samples

## Specific Changes

  - 
  -
  -


libraries/botbuilder-dialogs-adaptive/src/actions/httpRequest.ts
```diff
@@ -137,20 +144,40 @@ export class HttpRequest<O extends object = {}> extends Dialog<O> {
...
-        const url = this.url.evaluate(this.id, dc.state);
-        const body = this.body.evaluate(this.id, dc.state);
-        const headers = this.headers.evaluate(this.id, dc.state);
 
-        const response = await fetch(url, {
-            method: this.method.toString(),
-            headers: headers,
-            body: body
-        });
+        const url = stringTemplate.format(this.url, dc);
```
Stevenic: >const url = stringTemplate.format(this.url, dc); [](start = 8, length = 48)

This is likely a bug in the C# version of the code as well, but we need to make sure we URL encode any query string params.  We should decide on a solution for both JS and C# here and fix in both.

Stevenic:
>         return `AttachmentInput[]`;

We should include the name of the property we're bound to just to help make sure ID's are unique.

---
Refers to: libraries/botbuilder-dialogs-adaptive/src/input/attachmentInput.ts:47 in e9fabc0. [](commit_id = e9fabc0c75f4d8e51c49867f3b186b70a9a97a89, deletion_comment = False)

Stevenic:
>         return `AttachmentInput[]`;

This should be done for all inputs.

---
In reply to: [562376699](https://github.com/microsoft/botbuilder-js/pull/1451#issuecomment-562376699) [](ancestors = 562376699)

---
Refers to: libraries/botbuilder-dialogs-adaptive/src/input/attachmentInput.ts:47 in e9fabc0. [](commit_id = e9fabc0c75f4d8e51c49867f3b186b70a9a97a89, deletion_comment = False)

libraries/botbuilder-dialogs-adaptive/src/actions/httpRequest.ts
zidaneymar: > > const url = stringTemplate.format(this.url, dc); [](start = 8, length = 48)
> 
> This is likely a bug in the C# version of the code as well, but we need to make sure we URL encode any query string params. We should decide on a solution for both JS and C# here and fix in both.

Do we need to replace all string templates with expression property?

PR #1468, opened 2019-12-05
Add KeySuffix and CompatibilityMode to CosmosDbPartitionedStorage

EricDahlvang:
https://github.com/microsoft/botbuilder-dotnet/issues/2766

and

https://github.com/microsoft/botbuilder-dotnet/issues/2765


EricDahlvang:
Build is failing on this draft because of a test. Container existence is determined with a different call url now, so i just need to fix that test (and add some new ones).

libraries/botbuilder-azure/src/cosmosDbKeyEscape.ts
```diff
@@ -48,10 +53,14 @@ export namespace CosmosDbKeyEscape {
...
         );
 
-        return truncateKey(sanitizedKey);
+        return truncateKey(`${sanitizedKey}${keySuffix || ''}`, compatibilityMode);
     }
 
-    function truncateKey(key: string): string {
+    function truncateKey(key: string, truncateKeysForCompatibility?: boolean): string {
+        if (truncateKeysForCompatibility === false) {
+            return key;
```
mdrichardson: I'd be surprised if it ever exceeded it, but should we also do a check to ensure `key.length < 1024`? For BotState storage, this will probably never occur, but I could see somebody maybe doing something weird like stringifying some JSON and using that as a key for custom storage.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
```diff
@@ -106,6 +127,22 @@ export class CosmosDbPartitionedStorage implements Storage {
...
+        if (typeof cosmosDbStorageOptions.compatibilityMode === "undefined") {
+            cosmosDbStorageOptions.compatibilityMode = true;
+        }
+        if (cosmosDbStorageOptions.keySuffix) {
+            if (cosmosDbStorageOptions.compatibilityMode){
+                throw new ReferenceError('compatibilityMode cannot be true while using a keySuffix.');
+            }
+            // In order to reduce key complexity, we do not allow invalid characters in a KeySuffix
+            // If the keySuffix has invalid characters, the escaped key will not match
+            var suffixEscaped = CosmosDbKeyEscape.escapeKey(cosmosDbStorageOptions.keySuffix);
```
mdrichardson:
```suggestion
            const suffixEscaped = CosmosDbKeyEscape.escapeKey(cosmosDbStorageOptions.keySuffix);
```

nitpicky, but might prevent somebody doing something weird with it in the future

libraries/botbuilder-azure/src/cosmosDbKeyEscape.ts
EricDahlvang: That was considered, but a decision was made to to let cosmosdb throw an exception under this condition.

libraries/botbuilder-azure/src/cosmosDbPartitionedStorage.ts
EricDahlvang: committed your suggestion ... thanks!

PR #1473, opened 2019-12-06
[Skills] add skill support in BotFrameworkAdapter

stevengum:
## Description
Refactoring around BotFrameworkAdapter to add Skill support

## Specific Changes
 - ClaimsIdentity and a ConnectorClient are now cached in the `context.turnState` in `processActivity()`
 - **BotFrameworkAdapter.createConnectorClient()**
    - returns result from createConnectorClientInternal()
 - **private createConnectorClientInternal(serviceUrl: string, credentials: AppCredentials)**
    - Contains logic formerly in createConnectorClient()
    - If the serviceUrl is a streaming URL, it uses the StreamingHttpClient instead of the default HttpClient from ms-rest. (#1464)
 - **public async createConnectorClientWithIdentity(serviceUrl: string, identity: ClaimsIdentity)**
    - Creates a ConnectorClient and uses the ClaimsIdentity to determine if the request was a Skills request
 - **private async getAppCredentials(appId: string, oAuthScope: string = '')**
    - Gets a new AppCredentials for use with skills
    - There is no instance-level cache of AppCredentials per BotFrameworkAdapter, so it currently makes a new AppCredentials instance which is expensive (due to an `adal-node`-level token cache which is confined to the instance of the AppCredentials)

### DIFF
--- a/libraries/botbuilder/src/botFrameworkAdapter.ts
+++ b/libraries/botbuilder/src/botFrameworkAdapter.ts
@@ -157,7 +157,7 @@ export interface BotFrameworkAdapterSettings {
     certificatePrivateKey?: string;
     
     /**
-     * Optional. Recommended for Skills
+     * Optional. Used to require specific endorsements and verify claims. Recommended for Skills.
      */
     authConfig?: AuthenticationConfiguration;
 }
@@ -250,6 +250,8 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
     private streamingServer: IStreamingTransportServer;
     private webSocketFactory: NodeWebSocketFactoryBase;
 
+    private authConfiguration: AuthenticationConfiguration;
+
     /**
      * Creates a new instance of the [BotFrameworkAdapter](xref:botbuilder.BotFrameworkAdapter) class.
      *
@@ -286,6 +288,8 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
         // These values may be set when a bot is provisioned on Azure and if so are required for the bot to properly work in Public Azure or a National Cloud.
         this.settings.channelService = this.settings.channelService || process.env[AuthenticationConstants.ChannelService];
         this.settings.openIdMetadata = this.settings.openIdMetadata || process.env[AuthenticationConstants.BotOpenIdMetadataKey];
+
+        this.authConfiguration = this.settings.authConfig || new AuthenticationConfiguration();
 
         if (this.settings.openIdMetadata) {
             ChannelValidation.OpenIdMetadataEndpoint = this.settings.openIdMetadata;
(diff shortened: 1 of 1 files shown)


libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -968,25 +979,87 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
+            const scope = JwtTokenValidation.getAppIdFromClaims(identity.claims);
+            if (this.credentials.oAuthScope === scope) {
+                // Do nothing, the current credentials and its scope are valid for the skill.
+                // i.e. the adatper instance is pre-configured to talk with one skill.
+            } else {
+                // Since the scope is different, we will create a new instance of the AppCredentials
+                // so this.credentials.oAuthScope isn't overridden.
+                credentials = await this.getAppCredentials(botAppId, scope);
+
+                if (JwtTokenValidation.isGovernment(this.settings.channelService)) {
```
carlosscastro: In R8 we should discuss Skills in gov, and if yes test Skills in gov.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -968,25 +979,87 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
...
+        return new ConnectorClient(credentials, { baseUri: serviceUrl, userAgent: USER_AGENT });
+    }
+
+    /**
+     * @private
+     * @remarks
+     * @param appId 
+     * @param oAuthScope 
+     */
+    private async getAppCredentials(appId: string, oAuthScope: string = ''): Promise<AppCredentials> {
```
carlosscastro: I made this protected in C# so people can override.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -155,6 +155,11 @@ export interface BotFrameworkAdapterSettings {
      * Optional. Certificate key to authenticate the appId against AAD.
      */
     certificatePrivateKey?: string;
+    
+    /**
+     * Optional. Recommended for Skills
```
carlosscastro: If recommended for skills, should we say why we recommend them in skills?

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: The equivalent of this code is in [C#](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder/BotFrameworkAdapter.cs#L1168) and [Python](https://github.com/microsoft/botbuilder-python/blob/master/libraries/botbuilder-core/botbuilder/core/bot_framework_adapter.py#L648)... Are you saying we should remove it from all three languages? (I'm not opposed to this, as we haven't tested Skills in Gov)

Technically speaking, the type of Cloud shouldn't matter for parent-skill pairings as long as they're in the SAME cloud. Inter-cloud skills on the other-hand, it's not clear if there is any support for this.

There is _some_ interest in 3rd party channels on Gov Cloud, so we should factor code in a way that accounts for this possibly being a required scenario in the event that these 3rd party channels are NOT supported in Gov.

stevengum: Spoke with @gabog, leaving code in.

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: Renamed to `buildCredentials` to match the [BuildCredentialsAsync](https://github.com/microsoft/botbuilder-dotnet/blob/09409674cb3ca643b0d29b9f38154a55adca5d53/libraries/Microsoft.Bot.Builder/BotFrameworkAdapter.cs#L1027) method in C# and made it `protected`.

PR #1482, opened 2019-12-08
[Skills] add SkillHandler & SkillConversationIdFactoryBase classes

stevengum:
Fixes #

## Description
Port SkillHandler and SkillConversationIdFactoryBase classes from C#.

## Specific Changes
  - Add `SkillHandler`
  - Add `SkillConversationIdFactoryBase`
  - Parity for microsoft/botbuilder-dotnet#3111

## Testing
Added some unittests for `SkillHandler` and tests for `OAuthPrompt`.


stevengum:
@gabog that's correct, I think I can get away without adding a new overload/optional parameter for the ClaimsIdentity to continueConversation since we create the callback and pass it to the method. This way I don't have to touch the Adapter and this PR just focuses on the Skills layering

libraries/botbuilder/src/skills/skillHttpClient.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * A BotFrameworkHttpClient specialized for Skills that encapsulates Conversation ID generation.
+ */
+export class SkillHttpClient extends BotFrameworkHttpClient {
+    constructor(credentialProvider: ICredentialProvider, private readonly conversationIdFactory: SkillConversationIdFactoryBase, channelService?: string) {
+        super(credentialProvider, channelService);
+    }
+
+    public async postToSkill(fromBotId: string, toSkillId: string, toSkillUrl: string, serviceUrl: string, activity: Activity): Promise<InvokeResponse> {
```
gabog: It doesn't seem to be using the BotFrameworkSkill class, the goal of this class is to simplify the number of attributes needed to call a skill and also use the factory in a consistent way like the SkillHandler does. 
Are we not porting this 1:1 to match c#?

stevengum: I've changed it to match C# 👍

libraries/botbuilder/src/skills/botFrameworkSkill.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+/**
+ * Registration for a BotFrameworkHttpProtocol based Skill endpoint.
+ */
+export interface BotFrameworkSkill { 
+    /**
+     * Id of the skill.
+     */
+    id?: string;
```
gabog: I think it shouldn't be nullable (is that what the ? mean in js?)

libraries/botbuilder/src/skills/skillHttpClient.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export class SkillHttpClient extends BotFrameworkHttpClient {
+    constructor(credentialProvider: ICredentialProvider, private readonly conversationIdFactory: SkillConversationIdFactoryBase, channelService?: string) {
+        super(credentialProvider, channelService);
+        if (!this.conversationIdFactory) {
+            throw new Error('conversationIdFactory missing');
+        }
+    }
+
+    public async postToSkill(fromBotId: string, toSkill: BotFrameworkSkill, serviceUrl: string, activity: Activity): Promise<InvokeResponse> {
```
gabog: >postToSkill [](start = 17, length = 11)

In C#, this method is called PostActivity instead of PostToSkill, consider matching the name.

stevengum: That's not possible in TypeScript 😞

libraries/botbuilder/src/skills/skillHandler.ts
```diff
@@ -0,0 +1,185 @@
...
+     * @param bot The ActivityHandlerBase instance.
+     * @param conversationIdFactory A SkillConversationIdFactoryBase to unpack the conversation ID and map it to the calling bot.
+     * @param credentialProvider The credential provider.
+     * @param authConfig The authentication configuration.
+     * @param channelService The string indicating if the bot is working in Public Azure or in Azure Government (https://aka.ms/AzureGovDocs).
+     */
+    constructor(
+        private readonly adapter: BotAdapter,
+        private readonly bot: ActivityHandlerBase,
+        private readonly conversationIdFactory: SkillConversationIdFactoryBase,
```
gabog: I think this parameter is required, should we add a null check?

libraries/botbuilder/src/skills/botFrameworkSkill.ts
stevengum: Made property required 👍

libraries/botbuilder/src/skills/skillHandler.ts
stevengum: There's a falsy check on line 49

PR #1488, opened 2019-12-09
[Streaming] change isConnected() to a getter

### LINKED ISSUE #1427: [Streaming] Remove unnecessary methods and properties from botframework-streaming
## Versions
`botframework-streaming@4.6.0-preview3` via MyGet

## Describe the bug
There are classes implementing interfaces where the methods implemented could be getters. Some of the methods on the interfaces may also be unneeded and could be removed. 

`ITransport` which is impelmented by the `NamedPipeTransport` and `WebSocketTransport` denotes that  `isConnected(): boolean` be implemented. 

Searching the repository for references to `ITransport.isConnected()` beyond those three structs yields nothing, so the method should be removed.

## Expected behavior
Necessary methods and properties should indicate what they're used for via their documentation, implementation and usage.

[bug]

stevengum:
Fixes #1427

## Description
Convert `isConnected` function into getter.

## Testing
Touched up unit tests to not call isConnected, tested with multiple DLS connections to a bot.

### DIFF
--- a/libraries/botframework-streaming/src/interfaces/ITransport.ts
+++ b/libraries/botframework-streaming/src/interfaces/ITransport.ts
@@ -10,6 +10,6 @@
  * Abstraction for a generic transport definition.
  */
 export interface ITransport {
-    isConnected(): boolean;
+    isConnected: boolean;
     close();
 }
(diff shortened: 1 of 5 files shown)


ckkashyap:
We've validated that isConnected is not used in DLJS so we are good.

PR #1500, opened 2019-12-12
Add Italian Support for Confirm/Choice Prompts

mdrichardson:
Related: https://github.com/microsoft/botbuilder-dotnet/pull/3129

Recognizers-Text [added support for a few additional languages in C#](https://github.com/microsoft/Recognizers-Text/blob/7b20c1bbbfdebc3a801fbc61b094706ceebd7acf/.NET/Microsoft.Recognizers.Text/Culture.cs#L25) and [Italian in JS](https://github.com/microsoft/Recognizers-Text/blob/7b20c1bbbfdebc3a801fbc61b094706ceebd7acf/JavaScript/packages/recognizers-text/src/culture.ts#L12), which we don't currently support in our Confirm and Choice prompts:

* Bulgarian (C# only)
* Hindi (C# only)
* Italian
* Korean (C# only)
* Swedish (C# only)
* Turkish (C# only)

This PR adds support for those, ensures they get test coverage, and also adds test coverage for `PromptCultureModels`.

*Notes:* 

* I skipped adding Arabic support because it is read Right to Left, which makes prompt support tricker.
* Recognizers-Text JS SDK currently has only added Italian, but I don't believe there's a reason our C#/Node SDKs need to have matching language support
* I'm also far from a polyglot, so most of the translations were provided by [cough]Google Translate[cough]*

# Need to Decide Before Merge

English is really the only language that uses the Oxford/Serial comma. I.e, in this sentence:

> Yes, no, or maybe

...the second comma is the Oxford/Serial comma. Most other languages would write it as:

> Yes, no or maybe

In our SDK, this is the `inlineOrMore` property:

```csharp
public static Spanish: PromptCultureModel = {
    locale: Culture.Spanish,
    separator: ', ',
    inlineOr: ' o ',
    inlineOrMore: ', o ',
    yesInLanguage: 'Sí',
    noInLanguage: 'No',
}
```

..without the Oxford comma, `inlineOrMore` would look the same as `inlineOr`.

The languages that I added **do not** have the Oxford comma (as appropriate). However, many of the current languages we have support for **do** use the comma in our SDK, although would likely not in real life. 

- [ ] **Should those commas be removed in the SDK, or would that possibly create backwards-compat issues?**


mdrichardson:
@cleemullins ^s removed.

mdrichardson:
^s removed, I mean (can't edit)

PR #1513, opened 2019-12-18
Create new TurnContext using function so that derived classes can ove…

shmuelgutman:
…rride the default TurnContext

Like happens in the regular adapter, use a helper function to create a new instance of the TurnContext so that derived classes can override the default TurnContext


cleemullins:
This only applies to the Test Adapter. Handy.

PR #1531, opened 2019-12-24
fix comments and add get fullName method in fileResource

cosmicshuai:


libraries/botbuilder-dialogs-declarative/src/pathUtil.ts
```diff
@@ -1,19 +1,24 @@
...
-export class PathUtil {
-    public static IsDirectory = source => lstatSync(source).isDirectory()
-    public static GetDirectories = source => readdirSync(source).map(name => join(source, name)).filter(PathUtil.IsDirectory)
+/**
+ * @module botbuilder-dialogs-declarative
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
```
Stevenic: Should be top of file

libraries/botbuilder-dialogs-declarative/src/resources/resourceExplorer.ts
```diff
@@ -7,14 +7,19 @@ import { FolderResourceProvider } from "./folderResoureProvider";
...
 import { EventEmitter } from "events";
 
+/**
+ * @module botbuilder-dialogs-declarative
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
```
Stevenic: should be top of file

PR #1540, opened 2020-01-02
[Webpack][ISSUE#818][BotBuilder-AI][BF-Connector] Fix webpack dependency warning

### LINKED ISSUE #818: Webpack bundling broken
## Versions
What package version of the SDK are you using. Latest
What nodejs version are you using. 8.10.0
What os are you using Linux

## Describe the bug
After 4.1.7 - polyfill fixes broke webpack bundling
Commit here: https://github.com/Microsoft/botbuilder-js/commit/a20a44c638332eea1666cde4907fe66d552f5224#diff-a4cb641e4ac7086ddd5b5af0968c47f4
Webpack cannot 'see' the require therefore leaves it in tries to look in node_modules

## To Reproduce
1. Bundle app using bot framework with webpack
2. Run bundle (tries to load form-data from node_modules), throws module error

## Expected behavior
Not to throw module error and form-data, node-fetch to be included in the bundle

## Additional context
Webpack used for bundling code on lambdas due to size limit

[bug]

Bill7zz:
Fixes #818

## Description
This pull request fixes the critical dependency warning reported in BotBuilder-AI and BotFramework-Connector.

We followed the same approach used in BotFramework-Emulator that uses an external file to set Fetch as part of the Global object [here](https://github.com/microsoft/BotFramework-Emulator/blob/master/packages/app/main/src/main.ts#L45).

#### BotBuilder-AI warning solution

we updated the way that the Fetch API is required in the [httpRequestUtils ](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/qnamaker-utils/httpRequestUtils.ts#L11) class to use an external file.

#### Testing

First, we added tests for the logic added to the library. 
image

Second, we modified a browser-bot to test the package fixed. Then, we tested it using Edge, Chrome and Firefox browsers.
image

Finally, we tested the case that fetch is not part of the Global object. To do that, we modified a console-echo-bot to use the package fixed. We decided to use this bot since it is the one that doesn’t use the BotBuilder package. This package has a sub-dependency that adds the fetch function to the Global object in this [line ](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botframework-connector/src/index.ts#L1)before running the QnA code. 

The next image shows the qna-console-bot running as we expected.
image

#### BotFramework-Connector warning solution

we updated the way that the Form-Data and Fetch APIs are set as part of the global object in the [index](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botframework-connector/src/index.ts#L1) file of the library.

#### Testing

First, we added tests for the logic added to the library.
image

Second, we tested the package fixed using a browser-bot. The next image shows the result of the Webpack build process before and after updating the BF-Connector package. In this case, the APIs are already part of the Window object in the browsers. 
image

Finally, we configure an echo-bot to use the package with the new code and it works as we expected. 
image

## Testing
The next image shows the result of the build process running on the branch with the fixes for the Webpack dependency warnings on BotBuilder-AI and BF-Connector.
image


libraries/botbuilder-ai/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es6",
-    "lib": ["es2015"],
+    "lib": ["es2015", "dom"],
```
stevengum: We don't want to add `"dom"` to our tsconfig.json as this can force changes on downstream TypeScript consumers of our libraries. See Issue #1565 and PR #1575

cleemullins:
@Bill7zz please address @stevengum  comment and re-submit.

denscollo:
Hi @cleemullins, we already submitted these changes that @stevengum asked for. What's left only is the re-review of the changes

PR #1575, opened 2020-01-14
remove "dom" from bf-streaming tsconfig and refactor code

### LINKED ISSUE #1565: [Streaming] bf-streaming 4.7.0 requires TS projects to use "dom" lib in tsconfig.json
## Versions
What package version of the SDK are you using. **4.7.0**

## Describe the bug
`BrowserWebSocket` includes a dependency on a Browser type (WebSocket) which is satisfied through our package including `"dom"` in our [tsconfig.json](https://github.com/microsoft/botbuilder-js/blob/688c9bf1a969c95167e981f5b1407fcfa9808eb7/libraries/botframework-streaming/tsconfig.json#L4) for bf-streaming.

https://github.com/microsoft/botbuilder-js/blob/a0db67794dabec4d377a3facef3e2cbb21ec5342/libraries/botframework-streaming/src/webSocket/browserWebSocket.ts#L10-L11

https://github.com/microsoft/botbuilder-js/blob/688c9bf1a969c95167e981f5b1407fcfa9808eb7/libraries/botframework-streaming/tsconfig.json#L4

Projects that use `botbuilder` and do not have `"dom"` in the libs of their tsconfig.json will be unable to build due to the missing type. 

The immediate workaround is for projects with a dependency on `botframework-streaming` or `botbuilder` to add `"dom"` to the project's tsconfig.json, but we should remove this from our package and replace any `"dom"`-specific types with interfaces.

## Expected behavior
Ability to build TS libraries that have a dependency on `botbuilder` without changing their `tsconfig.json`

## Additional context
First reported via https://github.com/howdyai/botkit/issues/1894

[bug]

stevengum:
Fixes #1565 in `4.7` branch.

## Description
For the 4.7.0 release, botbuilder and bf-streaming shipped with an inadvertent requirement for TypeScript developers to include the `"dom"` lib in the developer's own `tsconfig.json`.

This problem arises for users even when they don't take a direct dependency on botframework-streaming. Building a library that uses only botbuilder fails without including `"dom"`. (https://github.com/howdyai/botkit/issues/1894)

Additionally, another side effect of relying on WebSocket and FileReader types and definitions via TypeScript leads us to potential breaking changes in the future. The `"dom"` types included are also tightly coupled with the version of TypeScript, so relying on these types could allow the code to successfully transpile, but not actually work at runtime.

## Specific Changes
  - Create IBrowserWebSocket and IBrowserFileReader interfaces for WebSocket and FileReader
  - Add helper methods to create [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) and [FileReader](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) without relying on `"dom"` being in the tsconfig.json.

## Testing
Unit tests run locally pass.

### DIFF
--- a/libraries/botframework-streaming/src/interfaces/IBrowserFileReader.ts
+++ b/libraries/botframework-streaming/src/interfaces/IBrowserFileReader.ts
@@ -1,0 +1,19 @@
+/**
+ * @module botframework-streaming
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Partially represents a FileReader from the W3C FileAPI Working Draft.
+ * For more information, see https://w3c.github.io/FileAPI/#APIASynch.
+ * 
+ * This interface supports the framework and is not intended to be called directly for your code.
+ */
+export interface IBrowserFileReader {
+    result: any;
+    onload: (event: any) => void;
+    readAsArrayBuffer: (blobOrFile: any) => void;
+}
--- a/libraries/botframework-streaming/src/interfaces/IBrowserWebSocket.ts
+++ b/libraries/botframework-streaming/src/interfaces/IBrowserWebSocket.ts
@@ -1,0 +1,24 @@
+/**
+ * @module botframework-streaming
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * Partially represents a WebSocket from the HTML Living Standard.
+ * For more information, see https://html.spec.whatwg.org/multipage/web-sockets.html.
+ * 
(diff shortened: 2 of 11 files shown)


stevengum:
https://github.com/microsoft/botbuilder-js/pull/1575/commits/f6434298f12d6a67a58575698ceb3740b02b8c92 removes the `isBrowser()` utility and instead adds helpers that check for the presence of global `WebSocket` and `FileReader` to better align with the shipped implementation.

With @ckkashyap's help I tested the code against his branch ([`ckk/protocoljs`](https://github.com/microsoft/BotFramework-DirectLineJS/tree/ckk/protocoljs)) in [microsoft/BotFramework-DirectLineJS](https://github.com/microsoft/BotFramework-DirectLineJS) to verify no breaking changes downstream. Live and recorded tests were passing against this branch.

PR #1578, opened 2020-01-14
Declare return types in waterfallDialog.ts

v-kydela:
See files for small fixes

Edit: My second commit fixes an actual bug rather than just cleaning up code. If you cancelled a waterfall during a named step (a step not defined by an anonymous function), the telemetry logger would treat it as an unnamed step. I noticed that by reading the code. The bug is not in the other repos though it may have originated by copying code from Python and not adapting it correctly.

### DIFF
--- a/libraries/botbuilder-dialogs/src/waterfallDialog.ts
+++ b/libraries/botbuilder-dialogs/src/waterfallDialog.ts
@@ -245,7 +245,7 @@ export class WaterfallDialog<O extends object = {}> extends Dialog<O> {
                 'InstanceId': instanceId,
             }});
         } else if (reason === DialogReason.cancelCalled) {
-            var index = state.stepIndex;
+            var index = instance.state[state.stepIndex];
             var stepName = this.waterfallStepName(index);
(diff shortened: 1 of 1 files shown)


libraries/botbuilder-dialogs/src/waterfallDialog.ts
```diff
@@ -246,7 +245,7 @@ export class WaterfallDialog<O extends object = {}> extends Dialog<O> {
                 'InstanceId': instanceId,
             }});
         } else if (reason === DialogReason.cancelCalled) {
-            var index = instance.state[state.stepIndex];
+            var index = state.stepIndex;
```
johnataylor: is this correct?

stevengum: @v-kydela can you demonstrate that there is a bug and write a test that shows there is a bug in the code? After we determine there is a bug, we should port the test to the other languages as well.

v-kydela: I've rolled back this bug fix so I can put it in a separate PR

PR #1593, opened 2020-01-19
Fix Expression mishandles day formats and extends the ability of datetime formating

### LINKED ISSUE #1584: LG formatDateTime expression mishandles day formats
Using `formatDateTime(date,'ddd')` should result in "Mon" or "Tues" but results in the numeric day of the month -- same as using `d`.

Similarly, `dddd` should result in the full day name, "Monday" but results in the day of the month prepended with a 0, that is "016" on Jan 16. instead of "Thursday"

cosmicshuai:
Fixes #1584 

## Description
align formatDateTime funtion with C#.
Now, the expression in JS will convert C# style datetime format string to a Moment.js capable datetime format string. For example:
formatDateTime('2018-03-15T13:00:00Z', 'ddd') will return Thu.


libraries/botframework-expressions/src/formatConverter.ts
```diff
@@ -0,0 +1,497 @@
+enum State {
```
Danieladu: cpoyright

libraries/botframework-expressions/src/formatConverter.ts
```diff
@@ -0,0 +1,497 @@
...
+    LowerZ3,
+    InSingleQuoteLiteral,
+    InDoubleQuoteLiteral,
+    EscapeSequence,
+}
+
+export function convertCSharpDateTimeToMomentJS(fmtString: string): string {
+    let fmtResult = '';
+    let fmtState: any = State.None;
+    const fmtTolerant = true;
```
Danieladu: `fmtTolerant`  is always true, and can not be modified. So, should we remove the logic that when `fmtTolerant` is false?

libraries/botframework-expressions/src/formatConverter.ts
```diff
@@ -0,0 +1,497 @@
...
+    LowerY5,
+    LowerZ1,
+    LowerZ2,
+    LowerZ3,
+    InSingleQuoteLiteral,
+    InDoubleQuoteLiteral,
+    EscapeSequence,
+}
+
+export function convertCSharpDateTimeToMomentJS(fmtString: string): string {
```
Danieladu: add the remarks and comments for this function.[params, summary, return. etc.]

PR #1627, opened 2020-01-29
[Streaming] add generic socket closure handling

stevengum:
## Specific Changes
* add `IStreamingTransportServer.isConnected?: boolean`
   * add isConnected to WebSocketServer
   * add isConnected to NamedPipeServer
* expose `isConnected` via `BotFrameworkAdapter.isStreamingConnectionOpen`
    * this should be used used in `onTurnError` to avoid throwing an error due to socket closures
* add check to verify socket connection is open in sendActivities
* add falsy check tests

## Testing
To repro, setup an JS Echo bot with the following code in the onMessage handler:

```js
constructor() {
  super();
  this.onMessage(async (context, next) => {
    const replyText = `Echo: ${ context.activity.text }`;
    const promisifyTimeout = (delay) => {
      return new Promise(resolve => setTimeout(resolve, delay));
    };

    await promisifyTimeout(5000);
    await context.sendActivity(MessageFactory.text(replyText, replyText));
    // By calling next() you ensure that the next BotHandler is run
    await next();
  });
}
```

And in the onTurnError handler for the adapter, use include the following if-block:
```js
if (!context.adapter.isStreamingConnectionOpen) {
    console.error(' Streaming connection is closed and not able to message the user. End of onTurnError handling logic.');
    return;
}
```

After sending an activity to the bot via a Direct Line Speech client, close the client to close the socket connection to the bot.

When the bot attempts to use the socket and send an activity to the bot after the timeout, it will check the socket connection and throw an error if the connection is not open.

### DIFF
--- a/libraries/botbuilder/src/botFrameworkAdapter.ts
+++ b/libraries/botbuilder/src/botFrameworkAdapter.ts
@@ -332,6 +332,13 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
     }
 
     /**
+     * Used in streaming contexts to check if the streaming connection is still open for the bot to send activities.
+     */
+    public get isStreamingConnectionOpen(): boolean {
+        return (this.streamingServer as WebSocketServer).isConnected;
+    }
+ 
+    /**
      * Asynchronously resumes a conversation with a user, possibly after some time has gone by.
      *
      * @param reference A reference to the conversation to continue.
@@ -942,6 +949,9 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
                         throw new Error(`BotFrameworkAdapter.sendActivity(): missing conversation id.`);
                     }
                     if (activity && BotFrameworkAdapter.isStreamingServiceUrl(activity.serviceUrl)) {
+                        if (!this.isStreamingConnectionOpen) {
+                            throw new Error('BotFrameworkAdapter.sendActivities(): Unable to send activity as Streaming connection is closed.');
+                        }
                         TokenResolver.checkForOAuthCards(this, context, activity as Activity);
                     }
                     const client = this.getOrCreateConnectorClient(context, activity.serviceUrl, this.credentials);
--- a/libraries/botbuilder/src/botFrameworkAdapter.ts
+++ b/libraries/botbuilder/src/botFrameworkAdapter.ts
@@ -335,7 +335,7 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
      * Used in streaming contexts to check if the streaming connection is still open for the bot to send activities.
      */
     public get isStreamingConnectionOpen(): boolean {
-        return (this.streamingServer as WebSocketServer).isConnected;
+        return this.streamingServer.isConnected;
     }
  
     /**
--- a/libraries/botbuilder/src/streaming/streamingHttpClient.ts
+++ b/libraries/botbuilder/src/streaming/streamingHttpClient.ts
@@ -34,8 +34,12 @@ export class StreamingHttpClient implements HttpClient {
      */
     public async sendRequest(httpRequest: WebResource): Promise<HttpOperationResponse> {
         if (!httpRequest) {
-            throw new Error('SendRequest invalid parameter: httpRequest should be provided');
+            throw new Error('StreamingHttpClient.sendRequest(): missing "httpRequest" parameter');
         }
+        if (!this.server.isConnected) {
+            throw new Error('StreamingHttpClient.sendRequest(): Streaming connection is disconnected');
+        }
+
         const request = this.mapHttpRequestToProtocolRequest(httpRequest);
         request.path = request.path.substring(request.path.indexOf('/v3'));
         const res = await this.server.send(request);
--- a/libraries/botbuilder/src/streaming/streamingHttpClient.ts
+++ b/libraries/botbuilder/src/streaming/streamingHttpClient.ts
@@ -37,7 +37,8 @@ export class StreamingHttpClient implements HttpClient {
             throw new Error('StreamingHttpClient.sendRequest(): missing "httpRequest" parameter');
         }
         if (!this.server.isConnected) {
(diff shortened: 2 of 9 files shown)


libraries/botframework-streaming/src/webSocket/webSocketServer.ts
```diff
@@ -38,6 +39,7 @@ export class WebSocketServer implements IStreamingTransportServer {
      * @param requestHandler Optional [RequestHandler](xref:botframework-streaming.RequestHandler) to process incoming messages received by this server.
      */
     public constructor(socket: ISocket, requestHandler?: RequestHandler) {
+        this._socket = socket;
```
stevengum: This constructor is missing a `null` check, I'll add it in the next commit.

stevengum: Done!

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -942,6 +949,9 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
                         throw new Error(`BotFrameworkAdapter.sendActivity(): missing conversation id.`);
                     }
                     if (activity && BotFrameworkAdapter.isStreamingServiceUrl(activity.serviceUrl)) {
+                        if (!this.isStreamingConnectionOpen) {
+                            throw new Error('BotFrameworkAdapter.sendActivities(): Unable to send activity as Streaming connection is closed.');
+                        }
```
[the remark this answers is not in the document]
stevengum: I'm leaving this check here, but I've also added a lower-level check in the StreamingHttpClient implmementation. 

Direct Line Speech only supports sending Activities, but not updates and deletes.

libraries/botbuilder/src/streaming/streamingHttpClient.ts
```diff
@@ -34,8 +34,12 @@ export class StreamingHttpClient implements HttpClient {
      */
     public async sendRequest(httpRequest: WebResource): Promise<HttpOperationResponse> {
         if (!httpRequest) {
-            throw new Error('SendRequest invalid parameter: httpRequest should be provided');
+            throw new Error('StreamingHttpClient.sendRequest(): missing "httpRequest" parameter');
```
christopheranderson: Changing error message strings in JS can have the effect of being a breaking change due to JS devs doing regex matches for handling error cases. This is probably a minuscule amount of risk to change in this case, though.

libraries/botbuilder/src/streaming/streamingHttpClient.ts
```diff
@@ -34,8 +34,12 @@ export class StreamingHttpClient implements HttpClient {
      */
     public async sendRequest(httpRequest: WebResource): Promise<HttpOperationResponse> {
         if (!httpRequest) {
-            throw new Error('SendRequest invalid parameter: httpRequest should be provided');
+            throw new Error('StreamingHttpClient.sendRequest(): missing "httpRequest" parameter');
         }
+        if (!this.server.isConnected) {
+            throw new Error('StreamingHttpClient.sendRequest(): Streaming connection is disconnected');
```
christopheranderson:
```suggestion
            throw new Error('StreamingHttpClient.sendRequest(): Streaming connection is disconnected, and the request could not be sent.');

```

Would be good to explain why it's a problem in the error message

libraries/botframework-streaming/src/namedPipe/namedPipeServer.ts
```diff
@@ -40,6 +40,10 @@ export class NamedPipeServer implements IStreamingTransportServer {
      * @param autoReconnect Optional setting to determine if the client sould attempt to reconnect automatically on disconnection events. Defaults to true.
      */
     public constructor(baseName: string, requestHandler?: RequestHandler, autoReconnect: boolean = true) {
+        if (!baseName) {
+            throw new TypeError('NamedPipeServer: Missing baseName parameter');
```
christopheranderson: Throwing `TypeError` here for missing parameter, but you throw an `Error` here for the same thing: https://github.com/microsoft/botbuilder-js/pull/1627/files#diff-f90ca2f557a7f5ac0cad4a842c08d731R37

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -331,6 +331,13 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
 
     }
 
+    /**
+     * Used in streaming contexts to check if the streaming connection is still open for the bot to send activities.
+     */
+    public get isStreamingConnectionOpen(): boolean {
```
DDEfromOR: would `isStreamingConnectionReady` be more clear? I'm thinking of the four states; Connected, Disconnected, Connecting, Error, specifically the case around Connecting where the connection isn't ready to send messages but it is open and waiting for a connection to come in.

libraries/botbuilder/src/streaming/streamingHttpClient.ts
stevengum: That's true, I think in this case I think we can get away with it. But if there is major push-back I can   take this out...

libraries/botframework-streaming/src/namedPipe/namedPipeServer.ts
stevengum: Yes, the difference here is because the line you highlighted is already throwing an `Error`.

On this [line](https://github.com/microsoft/botbuilder-js/pull/1627/files#diff-f90ca2f557a7f5ac0cad4a842c08d731R37) I would rather throw a TypeError, but I've avoided changing the type here and plan on doing it in master for the R8 release.

libraries/botbuilder/src/botFrameworkAdapter.ts
stevengum: Valid point, I'm partial to the name because `"Ready"` to me implies a "just-begun" state, and I think the level of information conveyed from this `getter` is simple enough to either be "open" or "closed".

If we decide to actually dig into these four states, we can then surface this status from the server.

Issue #1670, opened 2020-02-05
botframework-streaming should publish ES5 code to NPM


compulim:
## Versions

`master` as of right now.

## Describe the bug

`botframework-streaming` is not working in IE11.

Probably, it is related to `tsconfig.json` not emitting ES5 code.

https://github.com/microsoft/botbuilder-js/blob/master/libraries/botframework-streaming/tsconfig.json#L3

If non-ES5 is highly desirable for efficiency reason, please publish dual entrypoints to NPM, with `browser` pointing to ES5 code and `main` point to ES6+ code.

## To Reproduce

1. Create a web project (e.g. Webpack)
1. Include `botframework-streaming` in the dependencies and the bundle
1. Open the web project in IE11

## Expected behavior

`botframework-streaming` should work in IE11.

Instead, it throw syntax error when loading in IE11.

## Additional context

[bug]

stevengum:
@compulim can you confirm if the dependency chain below is correct? 

> botframework-streaming -> botframework-directlinejs -> botframework-webchat

stevengum:
https://github.com/microsoft/botbuilder-js/issues/1669#issuecomment-589383079

PR #1677, opened 2020-02-06
Change the target to es5 to support IE11

ckkashyap:
This change is needed so that the library works on IE11. A bug was reported for this - https://github.com/microsoft/botbuilder-js/issues/1670

### DIFF
--- a/libraries/botframework-streaming/tsconfig.json
+++ b/libraries/botframework-streaming/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es6",
+    "target": "es5",
     "lib": ["es2015"],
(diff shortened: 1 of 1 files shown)


libraries/botframework-streaming/tsconfig.json
```diff
@@ -10,4 +10,4 @@
     "rootDir": "./src",
     "types" : ["node"]
   },
-}
\ No newline at end of file
+}
```
stevengum: Approved

stevengum:
Small update, we're currently planning on doing a multi-target approach where we support browser@es6, node.js, and es5.

stevengum:
Closing in favor of #1764

Issue #1510, opened 2019-12-17
[Parity] Support providing additional credentials for OAuth flow (JS)


gabog:
Tacking issue for implementing https://github.com/microsoft/botbuilder-dotnet/issues/3152

stevengum:
Closed via #1697

PR #1697, opened 2020-02-12
migrate c# change in for adding oauth app credentials as a parameter

lzc850612:
Fixes #
#1510 

## Description

This is the JS change for adding OAuthAppCredentials in for OAuthPrompt

## Specific Changes

Update signatures of methods in IUserTokenProvider interface
Update functions in OAuthPrompt to pass in the OAuthAppCredentials
Update BotFrameworkAdapter functions to pass in OAuthAppCredentials and use it to create TokenApiClient
Update TokenResolver to use the OAuthAppCredentials from turn state that's saved in BotFrameworkAdapter


carlosscastro:
@lzc850612 we dont want this change right? If so, please close PR

lzc850612:
> There appear to be breaking changes in this code. Please sync with @carlosscastro Carlos Castro FTE to review and get this code approve.d

Breaking changes fixed. Steven signed off.

libraries/botbuilder-core/src/appCredentialsProvider.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * @module botbuilder-core
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export class WebResource { }
```
stevengum: This should be an interface

Issue #1499, opened 2019-12-12
[Core] Add onEndOfConversation to ActivityHandler and ActivityHandlerBase 


stevengum:
[Parity with C# PR 3110](https://github.com/microsoft/botbuilder-dotnet/pull/3110)

[Specific changes](https://github.com/microsoft/botbuilder-dotnet/commit/a9b745240df82441e23c8e134784eb80e5e5661b#diff-d7f7837896d965756e71991ab1b0d404R79)

Issue #1140, opened 2019-09-04
Missing onTyping on ActivityHandler


compulim:
## Versions

`master` and Node 12.

## Describe the bug

In `ActivityHandler.ts`, it supports various type of activities, including "Message", "Event", "MessageReaction", etc.

One notable and common activity type is "Typing". But it is missing. That means, the following code is not working.

```js
this.onTyping(async (context, next) => {
   // Process "typing" activity here

   await next();
});
```

To workaround, one have to listen to "UnrecognizedActivityType".

```js
this.onUnrecognizedActivityType(async (context, next) => {
  if (context.activity.type === 'typing') {
    // Process "typing" activity here
  }

  await next();
});
```     

## To Reproduce

1. Create a new bot using `ActivityHandler` pattern
1. In the constructor of the `Bot` class, which inherits `ActivityHandler`, call `this.onTyping(handler)`

## Expected behavior

`ActivityHandler` should have `onTyping` handler. "Typing" is a common activity IMO.

## Additional context

> Although I am working on streaming extensions, this is not related.

[bug]

PR #1706, opened 2020-02-12
add EndOfConversation and Typing activities to ActivityHandler

johnataylor:


johnataylor:
This addresses #1499 and #1140

PR #1756, opened 2020-02-20
Adding our own partial versions of node's 'net' Server and Socket classes

### LINKED ISSUE #1663: [Streaming] Stop using @types/node in botframework-streaming
## Versions
What package version of the SDK are you using. **4.7.1**

## Describe the bug
In the NamedPipeServer we have direct type dependencies on classes from the `"net"` module. We should use the library abstractions to avoid unclear type dependencies on Node.js builtins.

## Additional context
Related to https://github.com/microsoft/botbuilder-js/pull/1575

[bug]

Zerryth:
Fixes #1663

## Description
**Issue**: In the NamedPipeServer we have direct type dependencies on classes from the "net" module. We should use the library abstractions to avoid unclear type dependencies on Node.js builtins.

## Specific Changes
Similar to work done to [remove "dom" from bf-streaming tsconfig and code refactoring](https://github.com/microsoft/botbuilder-js/pull/1575), added the following:

- Create `INodeServer` interface for usage of what was node's `Sever` class from 'net' module in our framework's `NamedPipeServer`
- Add helper methods to create node net `Server` without relying on "net" being in the tsconfig.json.


Zerryth:
Still work in progress -- tests not passing...

Zerryth:
It breaks right now, because the only way for `NamedPipeServer` to create a `Server` right now is if Server is in the global scope.

In order to make this work, might need to build a `NodeNetServer` class similar to [`NodeWebSocket`](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botframework-streaming/src/webSocket/nodeWebSocket.ts#L17).

That way, as in [`WebSocketClient.connect()`](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botframework-streaming/src/webSocket/webSocketClient.ts#L62), there is a fallback--if there's a global one, create a Server using `NamedPipeServer.createServer()`, otherwise initialize our `NodeNetServer` .

PR #1741, opened 2020-02-19
Added error messages to abortWebSocketUpgrade()

Zerryth:
Fixes #1460

## Description
- `abortWebSocketsUpgrade()` originally only returned the error's status code to the client.
- Now it can return the error message, for a better debugging experience

## Specific Changes
  -  Added optional `message` parameter to `abortWebSocketsUpgrade()` method

PR #1757, opened 2020-02-20
Added Channels enum in botbuilder-schema for ChannelIds constants

### LINKED ISSUE #1100: Define constants for Channel IDs
In dotnet we have centralized constants for channel ids [here](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Connector/Channels.cs). However, in Javascript there are channel id constants defined within the choice prompt classes, but they are used in nearby classes and I found occurrences of the string literals for channel ids.

We have new adapters for new channels coming and each one has a const for the channelId, it may be better to have this in BotBuilder.Connector to avoid future collisions but I'm not 100% sure. 

The alternative would be to have each new channel adapter expose a const with the channelId they handle so we don't have to modify the product when we add new channels, but we probably need to agree on where a dev would expect to find this definition so we are consistent across the board.

If we go with the alternative, devs would have two places for ChannelId (Connector and the adapter), which may be confusing I am more inclined to have it in a single place for consistency.

Zerryth:
Fixes #1100 

## Description
* Currently Channel Ids were defined tucked away in prompt choices, but were being used by other dialogs (such as OAuthPrompt)
* Additionally we have new adapters for new channels coming, and each one has a const for the channelId
* It's better to have the channelIds in `botframework-schema`, to help avoid future collisions
    * Can't put the channelIds in bb-connector as in C#, because the dialogs library is supposed to be browser-friendly, and connector has node dependencies

## Specific Changes
  - Created readonly string enum in botbuilder-schema to define channelId constants
  - Changed references in the various classes and tests to point to new channels enum


Zerryth:
(Breaking on the postinstalls step again, like in [abortWebSocketUpgrade() PR](https://github.com/microsoft/botbuilder-js/pull/1741) )

Note: package.json of `botbuilder-dialogs` currently is pointing to relative path to `botframework-schema`, as the current published version doesn't have the new Channels enum bits. 
Consider changing to version number once it is published?

libraries/botbuilder-dialogs/package.json
```diff
@@ -26,6 +26,7 @@
     "@microsoft/recognizers-text-suite": "1.1.4",
     "@types/node": "^10.12.18",
     "botbuilder-core": "4.1.6",
+    "botframework-schema": "../botframework-schema",
```
stevengum: This isn't necessary as botbuilder-core surfaces the contents of `botframework-schema`.

libraries/botbuilder-dialogs/src/choices/channel.ts
```diff
@@ -6,28 +6,7 @@
...
-    line: 'line',
-    msteams: 'msteams',
-    skype: 'skype',
-    skypeforbusiness: 'skypeforbusiness',
-    slack: 'slack',
-    sms: 'sms',
-    telegram: 'telegram',
-    webchat: 'webchat'
-};
+import { Channels } from 'botframework-schema';
```
stevengum: `import { Channels, TurnContext } from 'botbuilder-core';`

Zerryth: @stevengum coolio, deleted schema dependency as instructed, as core already requires and exports it

libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts
```diff
@@ -9,7 +9,7 @@ import { Activity, ActivityTypes, Attachment, CardFactory, InputHints, MessageFa
 import { Dialog, DialogTurnResult } from '../dialog';
 import { DialogContext } from '../dialogContext';
 import { PromptOptions, PromptRecognizerResult,  PromptValidator } from './prompt';
-import { channels } from '../choices/channel';
+import { Channels } from 'botframework-schema';
```
stevengum: These imports still need to be updated

cleemullins:
@Zerryth CAn we get this resolved & closed out today?

Zerryth:
@stevengum @cleemullins fixed bb-core references in OauthPrompt and OAuthPrompt's tests

Right now it looks like check that is failing is unrelated to this PR? (It's that non-deterministic TokenResolver test:
image
)

stevengum:
@Zerryth can you file a bug for that failing test, and label it as R9, engineering, P1?

Zerryth:
Done! [Michael beat me to it](https://github.com/microsoft/botbuilder-js/issues/1787) :satisfied:.

PR #1761, opened 2020-02-21
Updated ShowTypingMiddleware to re throw errors

### LINKED ISSUE #726: ShowTypingMiddleware suppresses errors and does not allow adapter.onTurnError to handle them
## Versions
4.2.0
8.11.1
N/A
What os are you using
MacOS

## Describe the bug
I have a onTurnError handler and ShowTypingMiddleware attached to the botframework adapter.
```
    adapter = new BotFrameworkAdapter(adapterSettings)
        .use(new ShowTypingMiddleware());

    // Catch-all for errors.
    adapter.onTurnError = onTurnError;
```
I've done some "digging" in the source code and found out this line 
```
return await next().then(stopInterval, stopInterval);
```
which just stops the interval in error/success scenario and does not allows the error to propagate to the adapter which normally "fires" onTurnError handler function.

The fix: 
```
function stopInterval(): void {
    finished = true;
    if (hTimeout) { clearTimeout(hTimeout); }
}

function stopIntervalWithError(err: Error): void {
    stopInterval();
    throw err; //  here the error is re-thrown
}
return await next().then(stopInterval, stopIntervalWithError);
```

## To Reproduce
N/A

## Expected behavior
BotFrameworkAdapter to "fire" the provided onTurnError function.

I would gladly contribute with a pull request, but following the guidelines https://github.com/Microsoft/botbuilder-js/wiki/Contribution-Guidelines

when I try to push the new created branch, I encounter the following

git push -u origin fix/typing-middleware-suppresses-errors
ERROR: Permission to Microsoft/botbuilder-js.git denied to jovanmilenkoski.

[bug]

garypretty:
Fixes #726 

Typing indicator works as normal, unless an error is thrown, in which case it is now re thrown to the adapter.

### DIFF
--- a/libraries/botbuilder-core/src/showTypingMiddleware.ts
+++ b/libraries/botbuilder-core/src/showTypingMiddleware.ts
@@ -89,6 +89,11 @@ export class ShowTypingMiddleware implements Middleware {
             if (hTimeout) { clearTimeout(hTimeout); }
         }
 
+        function stopIntervalWithError(err: Error): void {
+            stopInterval();
+            throw err;
+        }
+
         if (context.activity.type === ActivityTypes.Message) {
             // Set a property to track whether or not the turn is finished.
             // When it flips to true, we won't send anymore typing indicators.
@@ -98,7 +103,7 @@ export class ShowTypingMiddleware implements Middleware {
 
         // Let the rest of the process run.
         // After everything has run, stop the indicator!
-        return await next().then(stopInterval, stopInterval);
+        return await next().then(stopInterval, stopIntervalWithError);
 
     }
     private async sendTypingActivity(context: TurnContext): Promise<void> {
--- a/libraries/botbuilder-core/src/showTypingMiddleware.ts
+++ b/libraries/botbuilder-core/src/showTypingMiddleware.ts
@@ -84,16 +84,6 @@ export class ShowTypingMiddleware implements Middleware {
             );
         }
 
-        function stopInterval(): void {
-            finished = true;
-            if (hTimeout) { clearTimeout(hTimeout); }
-        }
-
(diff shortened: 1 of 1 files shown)


johnataylor:
@garypretty we had another fix in mind.  rather than mixing the await and the then() why not wrap the await next call in a try... finally and call the stopInterval from there.

libraries/botbuilder-core/src/showTypingMiddleware.ts
```diff
@@ -98,7 +103,7 @@ export class ShowTypingMiddleware implements Middleware {
 
         // Let the rest of the process run.
         // After everything has run, stop the indicator!
-        return await next().then(stopInterval, stopInterval);
+        return await next().then(stopInterval, stopIntervalWithError);
```
johnataylor: discussed with @Stevenic and his suggested fix was more like this:

```ts
        try
        {
              return await next();
         }
        finally
        {
               stopInterval();
        }
```

i.e. similar to the C#

EricDahlvang: @johnataylor  Wouldn't a try/finally still swallow all exceptions that occur within 'next()'?  The goal of the fix is to propagate those exceptions up the stack, not catch and swallow them.

johnataylor: @ERicDahlvang that's not my understanding. this should behave exactly like the finally in C#. in other words adding a catch will stop the exception but a finally just gets run as the exception is propagated

EricDahlvang: Oh, there's no catch.  Sorry, I missed that... carry on. :)

Issue #1523, opened 2019-12-20
[PORT] Update adapters for Skills compatibility


tomlm:
> Port this change from botbuilder-dotnet/master branch:
https://github.com/microsoft/botbuilder-dotnet/pull/3148

microsoft/botbuilder-dotnet#3147  - Added additional ContinueConversationAsync overload for custom adapters for skills compatibility.
~microsoft/botbuilder-dotnet#3146 - Move 'BotIdentityKey' const to BotAdapter and made it public.~ Done via #1763

garypretty:
Dependent on #1623

Stevenic:
These changes need to be applied to all JS adapters for skills to work.

benbrown:
Part 1 is here -> #1763, this includes moving the BotIdentityKey to BotAdapter and changing the BotFrameworkAdapter's implementation of continueConversation to include the overload.

The other change will need to be applied to the adapters hosted in the Botkit repo... which will have to happen _after_ the changes to BotAdapter are published.

stevengum:
We moved the BotIdentityKey to the BotAdapter in #1736, so porting microsoft/botbuilder-dotnet#3146 has been completed.

cleemullins:
Closing per Ben and Steven.

antishok:
Hi, what about porting https://github.com/microsoft/botbuilder-dotnet/issues/3147 ?
It seems adapters are still not working with skills.
Thanks

PR #1763, opened 2020-02-21
Move BotIdentityKey to BotAdapter from BotFrameworkAdapter

benbrown:
Related to #1523

## Description
Move BotIdentityKey from  BotFrameworkADapter to BotAdapter.

This is a port of the changes made here:
https://github.com/microsoft/botbuilder-dotnet/issues/3146


libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -143,7 +143,6 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
     // These keys are public to permit access to the keys from the adapter when it's a being
     // from library that does not have access to static properties off of BotFrameworkAdapter.
     // E.g. botbuilder-dialogs
-    public readonly BotIdentityKey: Symbol = Symbol('BotIdentity');
```
stevengum: Ugh who wrote this code? :tired_face:

I've been looking at the C# BotFrameworkAdapter this week and saw that I put this key in the wrong location, thanks for fixing this!

stevengum:
@benbrown, @stevenic I have a PR out in dotnet that adds an overload for taking the OAuthScope/Audience value: https://github.com/microsoft/botbuilder-dotnet/pull/3409, maybe I should schedule a sync-up to discuss the necessary changes in dotnet and what needs to be ported.

@johnataylor, @carlosscastro

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -276,15 +275,28 @@ export class BotFrameworkAdapter extends BotAdapter implements IUserTokenProvide
      * });
      * ```
      */
-    public async continueConversation(reference: Partial<ConversationReference>, logic: (context: TurnContext) => Promise<void>): Promise<void> {
-        const request: Partial<Activity> = TurnContext.applyConversationReference(
+    public async continueConversation(claimsIdentity: ClaimsIdentity, reference: Partial<ConversationReference>, logic: (context: TurnContext) => Promise<void>): Promise<void>;
+    public async continueConversation(reference: Partial<ConversationReference>, logic: (context: TurnContext) => Promise<void>): Promise<void>;
+    public async continueConversation(...args): Promise<void> {
+        let claimsIdentity, reference, logic;
+        if (args.length === 3) {
```
stevengum: It would be great to generalize this logic for the other overloads, but that could be done in a different milestone.

If we take a laissez faire (more JS-like?) approach for this logic, should we instead do `args.length <== 3`? Otherwise if we continue with the current code I feel we should probably throw an Error.

benbrown: Not sure what you are trying to achieve.  if args.length is < 3 we want the else to fire...

stevengum: D'oh, `args.length >== 3`

benbrown: will do.

benbrown:
NOTE: This should not be merged until it has been adequately tested. I don't have the information required to test skills myself.

PR #1764, opened 2020-02-21
create browser (es5) output in botframework-streaming

### LINKED ISSUE #1669: botframework-streaming should separate main/browser exports
## Versions

`master` as of right now.

## Describe the bug

> I briefed @stevengum and @ckkashyap offline on this one.

In https://github.com/microsoft/botbuilder-js/blob/master/libraries/botframework-streaming/src/index.ts, it is exporting `namedPipe.js`, which depends on Node.js `net` package.

Instead, we should have 2 entrypoints, one for Node.js and one for browser. For the browser one, it should not export `namedPipe.js`.

## To Reproduce

1. Create a Webpack project
1. Include `botframework-streaming`
1. Run Webpack to bundle

## Expected behavior

Webpack should not throw error.

Instead, Webpack is complaining it cannot bundle `net` package. This is because `botframework-streaming/namedPipe.js` depends on `botframework-streaming`.

## Additional context

Also same for Node.js, which should not include dependencies that is targeting only browsers.

[bug]

### LINKED ISSUE #1670: botframework-streaming should publish ES5 code to NPM

stevengum:
Fixes #1669 
Fixes #1670

## Description
Creates a browser/es5 output of botframework-streaming
- Stop exporting Node.js-specific bits in the browser packages.

## Testing
@ckkashyap?

### DIFF
--- a/libraries/botframework-streaming/src/webSocket/index-browser.ts
+++ b/libraries/botframework-streaming/src/webSocket/index-browser.ts
@@ -6,7 +6,7 @@
  * Licensed under the MIT License.
  */
 
-export * from './browserWebSocket';
-export * from './webSocketClient';
-export * from './webSocketServer';
-export * from './webSocketTransport';
+export { BrowserWebSocket } from './browserWebSocket';
+export { WebSocketClient } from './webSocketClient';
+export { WebSocketServer } from './webSocketServer';
+export { WebSocketTransport } from './webSocketTransport';
--- a/libraries/botframework-streaming/src/webSocket/index-browser.ts
+++ b/libraries/botframework-streaming/src/webSocket/index-browser.ts
@@ -1,0 +1,12 @@
+/**
+ * @module botframework-streaming
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export * from './browserWebSocket';
+export * from './webSocketClient';
+export * from './webSocketServer';
+export * from './webSocketTransport';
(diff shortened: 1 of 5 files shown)


stevengum:
@benbrown, @stevenic FYI

libraries/botframework-streaming/src/webSocket/index-browser.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * @module botframework-streaming
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+export * from './browserWebSocket';
```
christopheranderson: nit: `export *` is potentially dangerous and I'd prefer being specific on which objects are being exported, especially given the fragility of our browser support without CI based testing.

stevengum:
Merging as we're not "officially officially" supporting Node.js 12 yet.

PR #1769, opened 2020-02-23
Port QnAMakerDialog from C#

### LINKED ISSUE #1744: Create the QnAMakerDialog class
In C# there is a QnAMakerDialog class. 

The PR in which that was created [is here](https://github.com/microsoft/botbuilder-dotnet/pull/3048). 

This class needs to be brought to Javascript and Python.

garypretty:
Port of QnAMakerDialog from C# https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.AI.QnA/Dialogs/QnAMakerDialog.cs

Fixes #1744


garypretty:
@stevengum I have one outstanding issue with this PR - as I mentioned the other day, this check -> https://github.com/microsoft/botbuilder-js/blob/80ebf56a22478cdfabe88796294aaa938ceb6cb0/libraries/botbuilder-dialogs/src/dialogSet.ts#L95 is failing.  

Adding a WaterfallDialog is fine, but trying to add a QnAMakerDialog (which extends WaterfallDialog), the instanceof check fails.  When debugging you can see that QnAMakerDialog has got Dialog in its prototype chain, but the check still fails.   

Any ideas?

garypretty:
@darrenj This PR should unblock Southworks on the JS work.

garypretty:
@Stevenic See my comments a couple further up this thread to see details of the issue I am facing.  Ultimately, if I new up an instance of the QnAMakerDialog and try and add it to a DialogSet I get an error.  Any help / advice appreciated here. Thanks!

libraries/botbuilder-ai/src/qnaMakerDialog.ts
```diff
@@ -0,0 +1,359 @@
...
+     * @param threshold (Optional) The threshold above which to treat answers found from the knowledgebase as a match.
+     * @param activeLearningCardTitle (Optional) Title of the card displayed showing active learning options if active learning is enabled.
+     * @param cardNoMatchText (Optional) Text to be show on a button alongside active learning options, allowing a user to indicate none of the options are applicable.
+     * @param top (Optional) Maximum number of answers to return from the knowledgebase.
+     * @param cardNoMatchResponse (Optional) Activity to be sent if the user selects the no match option on an active learning card.
+     * @param strictFilters (Optional) QnAMakerMetadata collection used to filter / boost queries to the knowledgebase.
+     * @param dialogId (Optional) Id of the created dialog. Default is 'QnAMakerDialog'.
+     */
+    constructor(knowledgeBaseId, endpointKey, hostName, noAnswer = null, threshold = 0.3, activeLearningCardTitle = 'Did you mean:', cardNoMatchText = 'None of the above.', top = 3, cardNoMatchResponse: Activity = null, strictFilters: QnAMakerMetadata[] = null, dialogId = 'QnAMakerDialog') {
```
Stevenic: >null [](start = 67, length = 4)

Never use "null" in JS. Use either "undefined" or the "?" operator.

Stevenic: All of these parameters need type annotations

libraries/botbuilder-ai/src/qnaMakerDialog.ts
```diff
@@ -0,0 +1,359 @@
...
+    
+    // state and step value key constants
+    private qnAContextData: string = 'previousContextData';
+    private previousQnAId: string = 'previousQnAId';
+    private options: string = 'options';
+    private qnAData: string = 'qnaData';
+    private currentQuery: string = 'currentQuery';
+
+    // Dialog options parameters
+    private defaultCardNoMatchResponse: string = "Thanks for the feedback.";
```
Stevenic: >"Thank [](start = 49, length = 6)

A bit of a nit but we try to use single quote ' or back tick ` for all strings. I'd probably go with back tick for these longer strings.

libraries/botbuilder-ai/src/qnaMakerDialog.ts
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    public async beginDialog(dc: DialogContext, options?: object): Promise<DialogTurnResult> {
+        if (!dc) { throw new Error('Missing DialogContext'); }
+        
+        if (dc.context.activity.type != ActivityTypes.Message) 
+        { 
+            return Dialog.EndOfTurn; 
+        }
+
```
Stevenic: The problem with this is that the waterfalls state won't be properly initialized since you don't call super.beginDialog().

If you're doing this because you could be a root dialog then I would call endDialog() instead. If you're not a root dialog, it's very unlikely you'd get started by a non-message activity.

libraries/botbuilder-ai/src/qnaMakerDialog.ts
```diff
@@ -0,0 +1,359 @@
...
+            Object.assign(dialogOptions, options);
+        }
+
+        return await super.beginDialog(dc, dialogOptions);
+    }
+
+    /**
+     * Returns a new instance of QnAMakerOptions.
+    **/
+    private async getQnAMakerOptions(): Promise<QnAMakerOptions> {
```
Stevenic: >async [](start = 12, length = 5)

why is this async?

libraries/botbuilder-ai/src/qnaMakerDialog.ts
```diff
@@ -0,0 +1,359 @@
...
+            qnaId: 0,
+            rankerType: RankerTypes.default,
+            isTest: false
+        };
+    };
+    
+    /**
+     * Returns a new instance of QnAMakerResponseOptions.
+    **/
+    private async getQnAResponseOptions(): Promise<QnAMakerDialogResponseOptions> {
```
Stevenic: >async [](start = 12, length = 5)

Again why is this async?

libraries/botbuilder-ai/src/qnaMakerDialog.ts
```diff
@@ -0,0 +1,359 @@
...
+     * if active learning is enabled and multiple score close answers are returned.
+    **/
+    private async callGenerateAnswer(step: WaterfallStepContext): Promise<DialogTurnResult> {
+        const dialogOptions: QnAMakerDialogOptions = step.activeDialog.state[this.options];
+        dialogOptions.qnaMakerOptions.qnaId = 0;
+        dialogOptions.qnaMakerOptions.context = { previousQnAId: 0, previousUserQuery: '' };
+
+        step.values[this.currentQuery] = step.context.activity.text;
+        const previousContextData: { [key: string]: number } = step.activeDialog.state[this.qnAContextData] ? step.activeDialog.state[this.qnAContextData] : {};
+        var previousQnAId = step.activeDialog.state[this.previousQnAId] ? step.activeDialog.state[this.previousQnAId] : 0;
```
Stevenic: >step.activeDialog.state[this.previousQnAId] ? step.activeDialog.state[this.previousQnAId] : 0 [](start = 28, length = 93)

You can simplify this to step.activeDialog.state[this.previousQnAId] || 0;

libraries/botbuilder-ai/src/qnaMakerDialog.ts
```diff
@@ -0,0 +1,359 @@
...
+                qnaResponse.answers.forEach(answer => {
+                    suggestedQuestions.push(answer.questions[0]);
+                });
+
+                var message = new QnACardBuilder().GetSuggestionsCard(suggestedQuestions, dialogOptions.qnaDialogResponseOptions.activeLearningCardTitle, dialogOptions.qnaDialogResponseOptions.cardNoMatchText); 
+                await step.context.sendActivity(message);
+
+                step.activeDialog.state[this.options] = dialogOptions;
+            
+                return { status: DialogTurnStatus.waiting, result: undefined };
```
Stevenic: >{ status: DialogTurnStatus.waiting, result: undefined } [](start = 23, length = 55)

You can just return Dialog.EndOfTurn

Stevenic: This is going to cause the waterfall to skip to the next step when it continues.

---
In reply to: [383549009](https://github.com/microsoft/botbuilder-js/pull/1769#discussion_r383549009) [](ancestors = 383549009)

libraries/botbuilder-ai/src/qnaMakerDialog.ts
```diff
@@ -0,0 +1,359 @@
...
+    /**
+    * QnAMakerDialog response options.
+    */
+    qnaDialogResponseOptions: QnAMakerDialogResponseOptions;
+}
+
+/**
+ * Query QnAMaker knowledgebase using user utterance. The dialog will also present user with appropriate multi-turn prompt or active learning options.
+ */
+export class QnAMakerDialog extends WaterfallDialog {
```
Stevenic: >WaterfallDialog [](start = 36, length = 15)

This doesn't feel like it should derive from waterfall... The fact you need to loop, conditionally run steps, etc. means that you need your own state machine. The waterfall is just complicating your life here.

libraries/botbuilder-ai/src/qnaCardBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+export class QnACardBuilder {
+
+    /**
+     * Returns an activity with a hero card attachment, containing buttons for active learning suggestions.
+     * @param suggestionsList List of suggestions to be displayed on hero card.
+     * @param cardTitle Title of the hero card.
+     * @param cardNoMatchText Text for button to be added to card to allow user to select 'no match'.
+     */
+    public GetSuggestionsCard(suggestionsList: string[], cardTitle: string, cardNoMatchText: string): Partial<Activity> {
```
Stevenic: >GetSuggestionsCard [](start = 11, length = 18)

All JS methods should be cammelCased.

Stevenic: >public [](start = 4, length = 6)

This should be a static method.

libraries/botbuilder-ai/src/qnaCardBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import { Activity, MessageFactory, CardAction, CardFactory } from 'botbuilder-core';
+import { QnAMakerResult } from './';
+
+/**
+ * Provides methods to create activities containing hero cards for showing active learning or multi-turn prompt options for the QnAMakerDialog.
+ */
+export class QnACardBuilder {
```
Stevenic: >QnACardBuilder [](start = 13, length = 14)

This should probably be named QnACardFactory to be consistent with our other factory classes.

libraries/botbuilder-ai/src/qnaCardBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Returns an activity with a hero card attachment, containing buttons for active learning suggestions.
+     * @param suggestionsList List of suggestions to be displayed on hero card.
+     * @param cardTitle Title of the hero card.
+     * @param cardNoMatchText Text for button to be added to card to allow user to select 'no match'.
+     */
+    public GetSuggestionsCard(suggestionsList: string[], cardTitle: string, cardNoMatchText: string): Partial<Activity> {
+        if (!suggestionsList) { throw new Error('Missing suggestionsList'); }
```
Stevenic: >!suggestionsList [](start = 12, length = 16)

I'd make this check: !Array.isArray(suggestionsList)

libraries/botbuilder-ai/src/qnaCardBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+        const message = MessageFactory.attachment(promptsCard, cardTitle);
+
+        return message;
+    }
+
+    /**
+     * Returns an activity with answer text and a hero card attachment, containing buttons for multi turn prompts.
+     * @param result QnAMaker result containing the answer text and multi turn prompts to be displayed.
+     */
+    public getQnAPromptsCard(result: QnAMakerResult): Partial<Activity> {
```
Stevenic: >public [](start = 4, length = 6)

Should be static

libraries/botbuilder-ai/src/qnaCardBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+
+        result.context.prompts.forEach(prompt => {
+            buttonList.push({
+                value: prompt.displayText,
+                type: 'imBack',
+                title: prompt.displayText
+            });
+        });
+
+        const promptsCard = CardFactory.heroCard('', null, buttonList);
```
Stevenic: >null [](start = 53, length = 4)

don't use null. use undefined

libraries/botbuilder-ai/src/qnaCardBuilder.ts
```diff
@@ -0,0 +1,73 @@
...
+            });
+        });
+
+        buttonList.push({
+            value: cardNoMatchText,
+            type: 'imBack',
+            title: cardNoMatchText
+        });
+
+        const promptsCard = CardFactory.heroCard('', null, buttonList);
```
Stevenic: >null [](start = 53, length = 4)

Don't use null. use undefined

libraries/botbuilder-ai/src/qnaCardBuilder.ts
garypretty: @Stevenic I am inclined to agree - issue is we already have a class in C# with this name.  Where do we stand in terms of amending that one to match?  Arguably the only place it is likely ot be used right now is QnAMakerDialog, but it technically would be a breaking change. https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.AI.QnA/Utils/QnACardBuilder.cs
cc @cleemullins

libraries/botbuilder-ai/src/qnaMakerDialog.ts
garypretty: @Stevenic I agree - I think right now though, if we want to maintain parity we need to live with this as a waterfall given the C# version has already been released https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.AI.QnA/Dialogs/QnAMakerDialog.cs

libraries/botbuilder-ai/src/qnaMakerDialog.ts
garypretty: Done.

libraries/botbuilder-ai/src/qnaMakerDialog.ts
garypretty: Done - thanks for the tip.

libraries/botbuilder-ai/src/qnaMakerDialog.ts
garypretty: Done.

libraries/botbuilder-ai/src/qnaMakerDialog.ts
garypretty: I have switched to use endDialog()

libraries/botbuilder-ai/src/qnaMakerDialog.ts
garypretty: Agreed - it shouldn't be. Altered. This was a hangover from the C# version which should arguably be changed, but not sure if that breaks binary compat?

libraries/botbuilder-ai/src/qnaMakerDialog.ts
garypretty: Again, agreed - it shouldn't be. Altered. This was a hangover from the C# version which should arguably be changed, but not sure if that breaks binary compat?

libraries/botbuilder-ai/src/qnaMakerDialog.ts
garypretty: Thanks - altered this in a few places where I was using the longer hand. Good tip!

libraries/botbuilder-ai/src/qnaMakerDialog.ts
garypretty: Done and thanks.

libraries/botbuilder-ai/src/qnaCardBuilder.ts
garypretty: Amended.

garypretty: Amended.

libraries/botbuilder-ai/src/qnaCardBuilder.ts
garypretty: Thanks - implemented.

libraries/botbuilder-ai/src/qnaCardBuilder.ts
garypretty: Done.

libraries/botbuilder-ai/src/qnaCardBuilder.ts
garypretty: Amended.

libraries/botbuilder-ai/src/qnaCardBuilder.ts
garypretty: Amended.

garypretty:
@Stevenic Thanks - implemented most of the changes. Left comments on a couple of items.

Did you see any reason why the 'instanceof Dialog' check would be failing?

PR #1771, opened 2020-02-24
add skillDialog and associated classes

### LINKED ISSUE #1635: SkillDialog (JS)
See [rollup Issue](https://github.com/microsoft/botframework-sdk/issues/5723) for spec/design.

stevengum:
Fixes #1635

## Specific Changes
  - Move skill-related interfaces and abstract classes to `botbuilder-core` from `botbuilder`
  - Add `BotFrameworkClient` abstract class to `botbuilder-core`
  - Move `InvokeResponse` interface to `botbuilder-core`


stevengum:
I cancelled the builds for https://github.com/microsoft/botbuilder-js/pull/1771/commits/1c0caa7285c67eff1f023b7cf129afea32fd957e as this code does not build (due to an unsatisfied `botbuilder`) dependency.

libraries/botbuilder-dialogs/src/skillDialog.ts
```diff
@@ -0,0 +1,135 @@
...
+import { BotFrameworkSkill } from 'botbuilder'; // bad import
+import { Activity, ActivityTypes, ConversationState, StatePropertyAccessor, TurnContext } from 'botbuilder-core';
+import { Dialog, DialogTurnResult } from './dialog';
+import { DialogContext } from './dialogContext';
+import { SkillDialogArgs } from './skillDialogArgs';
+import { SkillDialogOptions } from './skillDialogOptions';
+
+export class SkillDialog extends Dialog {
+    private readonly _activeSkillProperty: StatePropertyAccessor;
+    private readonly _conversationState: ConversationState; // Why is this restricted to ConversationState?
```
stevengum: Shouldn't this be `BotState`? 

https://github.com/microsoft/botbuilder-dotnet/blob/69112a39fa0aa7fb5cea6e506f41b36779c33681/libraries/Microsoft.Bot.Builder.Skills/Dialogs/SkillDialog.cs#L25

libraries/botbuilder-dialogs/src/skillDialog.ts
```diff
@@ -0,0 +1,135 @@
...
+
+        const dialogArgs = options as SkillDialogArgs;
+
+        if (dialogArgs.skill == undefined || dialogArgs.skill == null) {
+            throw new TypeError(`"skill" undefined or null in options`);
+        }
+
+        if (dialogArgs.activity === undefined || dialogArgs.activity === null) {
+            throw new TypeError(`"activity" is undefined or null in options.`);
+        }
```
stevengum: I'll be changing this to follow the normal falsy checks we do throughout the repository[...](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing)

libraries/botbuilder-dialogs/src/skillDialog.ts
```diff
@@ -0,0 +1,135 @@
...
+     * The options parameter in `beginDialog()` must be a `SkillDialogArgs` object with the initial parameters
+     * for the dialog.
+     * 
+     * @param SkillDialogOptions 
+     * @param dialogOptions 
+     * @param ConversationState 
+     * @param conversationState 
+     */
+    public constructor(dialogOptions: SkillDialogOptions, conversationState: ConversationState) {
+        super('SkillDialog'); // What about Dialog Id?
```
stevengum: @gabog just a nit, should we make `dialogId` configurable via `SkillDialogOptions`?

stevengum:
Per offline conversations with @stevenic and @benbrown, we're going to move the skill interfaces and abstract classes into `botbuilder-core`, which should solve the dependency problem.

libraries/botbuilder-dialogs/src/skillDialog.ts
stevengum: Going to stay with ConversationState because we're saving ConversationState and there are stronger semantics around it (SkillDialog needs to save the DialogStack which is normally saved in ConversationState).

libraries/botbuilder-dialogs/src/skillDialog.ts
```diff
@@ -0,0 +1,140 @@
...
+        if (dc.context.activity.type === ActivityTypes.Message || dc.context.activity.type === ActivityTypes.Event) {
+            // Just forward to the remote skill
+            await this.sendToSkill(dc.context, dc.context.activity);
+        }
+
+        return Dialog.EndOfTurn;
+    }
+
+    public async endDialog(context: TurnContext, instance: DialogInstance, reason: DialogReason): Promise<void> {
+        // Send of of conversation to the skill if the dialog has been cancelled.
```
carlosscastro: typo in comment

libraries/botbuilder-dialogs/src/skillDialog.ts
```diff
@@ -0,0 +1,140 @@
...
+        const skillConversationId = await this.dialogOptions.conversationIdFactory.createSkillConversationId(TurnContext.getConversationReference(activity) as ConversationReference);
+
+        // Always save state before forwarding
+        // (the dialog stack won't get updated with the skillDialog and things won't work if you don't)
+        await this.dialogOptions.conversationState.saveChanges(context, true);
+        const skillInfo = this.dialogOptions.skill;
+        const response = await this.dialogOptions.skillClient.postActivity(this.dialogOptions.botId, skillInfo.appId, skillInfo.skillEndpoint, skillInfo.skillEndpoint, skillConversationId, activity);
+
+        // Inspect the skill response status
+        if (!(response.status >= 200 && response.status <= 299)) {
```
carlosscastro: dont we have constants for this? Or otherwise it might be worth checking if in other places we do something like this which owuld be 'isSuccessStatusCode'. C# has a built in method for this actually, given how frequently folks ask this question. Potentially something to think about in general rather than for this PR, definitely would not block on it.

Issue #1668, opened 2020-02-05
Provide Cleanup Guidance around closing websockets


cleemullins:
Originally reported in #1655, and moved here so we've got an issue tracking the ask. 

> As there's not a single BotFrameworkAdapter but one for every WebSocket connection, it would also mean the new SkillHandler as well as ChannelServiceRoutes have to exist multiple times, which results in every conversation/connection resulting in a different endpoint via the SkillHandler (skillEndpoint.register(server, '/api/skills');).
Which currently is not easy to cleanup as the WebSocket does not seem to expose a disconnected event. This would result in a memory leak.

This is work we plan on shipping as part of R8, which ships ~March 15.

alopix:
Giving feedback on the linked ticket:
The setup is basically a combination of the two samples:
* https://github.com/microsoft/BotBuilder-Samples/blob/master/experimental/directline-speech/javascript_nodejs/02.echo-bot/index.js
* https://github.com/microsoft/BotBuilder-Samples/blob/master/samples/javascript_nodejs/80.skills-simple-bot-to-bot/simple-root-bot/index.js

```
server.on('upgrade', async (req, socket, head) => {
    // Create an adapter scoped to this WebSocket connection to allow storing session data.
    const streamingAdapter = new BotFrameworkAdapter(adapterSettings);
    // Set onTurnError for the BotFrameworkAdapter created for each connection.
    streamingAdapter.onTurnError = onTurnErrorHandler;

    await streamingAdapter.useWebSocket(req, socket, head, async (context) => {
        // After connecting via WebSocket, run this logic for every request sent over
        // the WebSocket connection.
        await myBot.run(context);
    });
});
```

combined with (both top-level)

```
// Create and initialize the skill classes
const authConfig = new AuthenticationConfiguration([], allowedSkillsClaimsValidator);
const handler = new SkillHandler(adapter, bot, conversationIdFactory, credentialProvider, authConfig);
const skillEndpoint = new ChannelServiceRoutes(handler);
skillEndpoint.register(server, '/api/skills');
```

Would not work, as `SkillHandler` requires access to the `adapter`, which is only created on connection -> `ChannelServiceRoutes` and registration with the `server` would have to be inside the websocket connection handler as well, so requires a different endpoint all the time.

johnataylor:
From another thread:

[...]

Issue #1655, opened 2020-02-03
New skill handling support in botbuilder JS SDK with Directline Speech


alopix:
## Versions
botbuilder 4.7
node 10.16.3
macOS 10.14

## Describe the bug
TL;DR;
It looks like the botbuilder-skills package is no longer being maintained, and instead similar functionality has been rolled into the botbuilder package. However, this new functionality does not appear to support Direct Line Speech out of the box - is there a timescale for that?

Full Description:
We discovered that the previous botbuilder-skills SDK seems to have been deprecated or at least not been worked on in a while and skill support been added to the main botbuilder JS SDK. To date we have been depending on the botbuilder-skills package.

Evaluating the changes necessary to migrate we noticed, that the example (https://github.com/microsoft/BotBuilder-Samples/tree/master/samples/javascript_nodejs/80.skills-simple-bot-to-bot) is specifically targeting DirectLine, as does the SDK itself. Support for DirectLine Speech is only possible with quite a bit of additional work as the way the adapter is being handled is not the same.

As there's not a single BotFrameworkAdapter but one for every WebSocket connection, it would also mean the new SkillHandler as well as ChannelServiceRoutes have to exist multiple times, which results in every conversation/connection resulting in a different endpoint via the SkillHandler (skillEndpoint.register(server, '/api/skills');).
Which currently is not easy to cleanup as the WebSocket does not seem to expose a disconnected event. This would result in a memory leak.

Apart from that, the solution does not necessarily scale with DirectLine Speech and multiple bot instances. A skill would only know the load balancer's URL and therefore might post back the response to an incorrect instance of the bot, which does not actually maintain the open client WebSocket connection.

A possible workaround for that would be some kind of message queue between skill and skill handler to distribute the responses – which is quite a bit of additional work.

I wonder if there's a different scaling solution Microsoft had in mind when designing this new skill architecture.

[bug]

daveta:
@darrenj : Can you comment?

cleemullins:
Hi @alopix Thanks for the details. 

We're releasing a new JS build today ([4.7.2, currently on our daily build server](https://botbuilder.myget.org/feed/botbuilder-v4-js-daily/package/npm/botbuilder/4.7.2)) to fix a bug around closed web sockets and unhandled exceptions. This is being regressed now, for release (hopefully) in the next 24 hours. 

[...]

PR #1783, opened 2020-02-25
Added buffered delivery mode support.

Stevenic:
Fixes: #1787
Fixes: #1668
Fixes: #1655

## Description
This adds support for a new `deliveryMode` called `buffered` which causes a bot to buffer any activities sent by the bot and then return them as an array in the http response to a received activity.


libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -457,13 +462,27 @@ export class TurnContext {
...
-                    // Set responded flag
-                    if (sentNonTraceActivity) { this.responded = true; }
-
-                    return responses;
+            if (this.activity.deliveryMode === 'buffered') {
+                // Append activities to buffer
+                const responses: ResourceResponse[] = [];
+                output.forEach((a) => {
+                    this.bufferedActivities.push(a);
+                    responses.push({ id: undefined });
```
stevengum: Are we just supporting sending and receiving message activities?

The current implementation doesn’t support operations that rely on Activity.id. 

What is the list of operations we do support or don’t support with buffered activity functionality?

libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -457,13 +462,27 @@ export class TurnContext {
...
                 });
+
+                // Set responded flag
+                if (sentNonTraceActivity) { this.responded = true; }
```
stevengum: Do we still send trace activities for this mode? The recipient of the bufferedActivities probably shouldn’t receive any traces.

libraries/botbuilder-core/src/turnContext.ts
stevengum: Does the parent store the activity.id on behalf of the skill, does it send them asynchronously in relation to the parent’s turn...? via an Event activity?

libraries/botbuilder-core/src/turnContext.ts
garypretty: > Do we still send trace activities for this mode? The recipient of the bufferedActivities probably shouldn’t receive any traces.

Would we not want to send all activities back in bufferedActivites and then leave it up to the parent to decide what to? In the case that we are calling a skill, the skill could conceivably send trace activities back which we, in turn, want to be sent by the parent back to the emulator, for example.

garypretty:
@Stevenic Looks good - this is very similar to a change @johnataylor and I made in the community adapters recently to simplify the buffering of the activities, where we need request / response.

Is the thinking here that a call to a skill, using a skill dialog I assume, will mark the outgoing request with the buffered delivery mode? Thus solving the affinity issue with skills.

libraries/botbuilder-core/src/turnContext.ts
stevengum: That’s true, on second thought I’d continue with the pattern that the parent and skill figure out what to do, rather than forcing behavior since the semantics are open-ended

johnataylor:
@garypretty yes absolutely, (perhaps there is some reuse possible) note the approach here should allow middleware to take a crack at the list of all activities if it wants to - for custom processing of sets of speak activities etc. (this is assuming a theoretical scenario where something like the alexa code is hosted and the application is talking to that with this buffer - i.e. logically equivalent to something your adapter is allowing developers to do with overriding.)

libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -457,13 +462,27 @@ export class TurnContext {
         });
 
         return this.emit(this._onSendActivities, output, () => {
-            return this.adapter.sendActivities(this, output)
-                .then((responses: ResourceResponse[]) => {
-                    // Set responded flag
-                    if (sentNonTraceActivity) { this.responded = true; }
-
-                    return responses;
+            if (this.activity.deliveryMode === 'buffered') {
```
johnataylor: agreement was to check in with 'bufferedReplies'

libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -341,6 +341,11 @@ export class TurnContext {
         return reference;
     }
 
+    /**
+     * List of activities to send when `context.activity.deliveryMode == 'buffered'`.
+     */
+    public readonly bufferedActivities: Partial<Activity>[] = [];
```
johnataylor: public readonly bufferedReplyActivities: Partial<Activity>[] = [];

libraries/botbuilder-core/src/turnContext.ts
stevengum: Resolved via 89f23cb0040ea628ad2c2c9d9ef58b8f15605aaa

PR #1795, opened 2020-02-26
Remove @types node dev dependency from botframework-streaming

### LINKED ISSUE #1663: [Streaming] Stop using @types/node in botframework-streaming

Zerryth:
Fixes #1663 

## Description
**Goal: remove our dependency on @types/node**
* Removed public types from node's `net` module from `botframework-streaming`
* Replaced node types with mock interfaces of node classes

## Specific Changes
* Added helper method to create Servers (`createNodeServer()`, `getServerFactory()`)

**Created interfaces**:
* `IEventEmitter`
* `INodeServer`

**Added to the following mock interfaces, to be the proper shape to replace node types**:
* `INodeBuffer`
* `INodeSocket` -- includes signatures from `Duplex`, so we didn't have to rely on `stream` types


libraries/botframework-streaming/src/namedPipe/namedPipeServer.ts
```diff
@@ -15,14 +15,15 @@ import {
...
+import { INodeSocket2, IStreamingTransportServer, IReceiveResponse } from '../interfaces';
+import { createNodeServer } from '../utilities';
 
 /**
 * Streaming transport server implementation that uses named pipes for inter-process communication.
 */
 export class NamedPipeServer implements IStreamingTransportServer {
-    private _outgoingServer: Server;
-    private _incomingServer: Server;
+    private _outgoingServer
```
christopheranderson: missing type on line 25/26

libraries/botframework-streaming/src/namedPipe/namedPipeServer.ts
```diff
@@ -73,14 +74,15 @@ export class NamedPipeServer implements IStreamingTransportServer {
         }
 
         const incoming = new Promise(resolve => {
-            this._incomingServer = new Server((socket: Socket): void => {
-                this._receiver.connect(new NamedPipeTransport(socket));
-                resolve();
-            });
+                this._incomingServer =  createNodeServer((socket: INodeSocket2): void => {
```
christopheranderson: I'm very skeptical of code that doesn't have error conditions, especially around networking. But it doesn't look like this is due to your PR, so could be considered out of scope.

libraries/botframework-streaming/tests/NamedPipe.test.js
```diff
@@ -1,6 +1,7 @@
 const net = require('net');
 const np = require('../lib');
 const npt = require('../lib/namedPipe/namedPipeTransport');
+const NodeServerUtils = require('../lib/utilities/createNodeServer');
```
christopheranderson:
```suggestion
const { createNodeServer, getNetServerConstructor } = require('../lib/utilities/createNodeServer');

```

Mostly a style nit. Ignore me if you'd like.

libraries/botframework-streaming/tests/NamedPipe.test.js
```diff
@@ -414,5 +415,29 @@ describe('Streaming Extensions NamedPipe Library Tests', () => {
...
+
+            expect(server).to.not.be.null;
+            expect(server).to.be.instanceOf(Object);
+        });
+
+        it('should return the constructor when calling getNodeServerConstructor()', () => {
+            const netServerCtor = NodeServerUtils.getNetServerConstructor();
+
+            expect(netServerCtor).to.not.be.null;
+            expect(typeof netServerCtor).to.equal('function');
```
christopheranderson: maybe use a `is` assertion function from TypeScript to assert that the function returns the right type?

``` typescript
function IsNetServer(o: any): o is Server {
    if(o.on && typeof o.on === 'function') {
        return true;
    } // or whatever the right logic is?
}
```

libraries/botframework-streaming/src/utilities/createNodeServer.ts
```diff
@@ -0,0 +1,30 @@
...
+ // What type hint should I put, if not importing 'net'?
+export const createNodeServer = function(callback?: (socket: INodeSocket2) => void) {
+    if (callback && typeof callback !== 'function') {
+        throw new TypeError(`Invalid callback; callback parameter must be a function to create Node 'net' Server.`);
+    }
+
+    return getNetServerConstructor()(callback);
+}
+
+export const getNetServerConstructor = function() {
```
christopheranderson: naming nit: a constructor is usually  `new Object` pattern. This is a factory method.

christopheranderson: You also need to assert the type signature here, I think. Else, you might run into issues with the types being output here. Check the .d.ts here to make sure it isn't outputting the `require('net').Server` signature.

libraries/botframework-streaming/src/namedPipe/namedPipeServer.ts
Zerryth: Created INodeServer interface

libraries/botframework-streaming/tests/NamedPipe.test.js
Zerryth: Added server type guards into `createNodeServer()`, as the tests are in plain JS.
Test that they don't throw with the typeguards in place

libraries/botframework-streaming/src/interfaces/ISocket.ts
```diff
@@ -1,3 +1,5 @@
+import { INodeBuffer } from "./INodeBuffer";
```
Stevenic: Move this below comment block

libraries/botframework-streaming/src/interfaces/INodeSocket.ts
```diff
@@ -6,13 +6,151 @@
...
+    write(str: string, encoding?: string, fd?: string): boolean; 
+    write(data: any, encoding?: string, callback?: Function): void; 
+    write(chunk: any, cb?: (error: Error | null | undefined) => void): boolean;
+    write(chunk: any, encoding?: string, cb?: (error: Error | null | undefined) => void): boolean;
+
+    ref(): any;
+    unref(): any;
+}
+
+interface AddressInfo {
```
Stevenic: It didn't complain that these two interfaces weren't explicitly exported?

libraries/botframework-streaming/src/namedPipe/namedPipeServer.ts
Stevenic: So if you look at createNodeServer() it throws if an error occurs which the new Promise() will catch and automatically turn into a rejection. I would have made createNodeServer() return a promise instead but this code should work ok. 

---
In reply to: [384669694](https://github.com/microsoft/botbuilder-js/pull/1795#discussion_r384669694) [](ancestors = 384669694)

libraries/botframework-streaming/src/interfaces/INodeSocket.ts
Zerryth: No, as I was just only using them within the file for `INodeSocket`

Zerryth:
@stevenic 

* moved the import statement below the comment
* and as per in person discussion, we're not making createNodeServer() return a promise
* I didn't export the 2 interfaces used only within `INodeSocket` still, but let me know if you want that done. Otherwise I'll leave it as is

Zerryth:
@johnataylor it's this P2 that needs review/approval
other one can be punted

Issue #1806, opened 2020-02-27
Botkit adapter must send a sign-in card instead of an OAuth card


ckkashyap:
Background: There are customer requests to support ABS OAuth in the adapter so that Bots that use the adapter may leverage the OAuth support that channels provide.

Ben and I have been able to build a POC by copying over OAuth related functions into the Facebook adapter. The only change in botbuilder that was needed was in libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts:248 - a sign-in card must be sent in case of adapter. In the POC, we simply checked for the name of the adapter but perhaps there is a better way to do it.

This appears to be a trivial change and can unblock the OAuth support in adapter. I will submit a PR for this soon.

ckkashyap:
PR created - https://github.com/microsoft/botbuilder-js/pull/1812

ckkashyap:
Fix has been merged.

PR #1812, opened 2020-02-28
Ensure that ABS OAuth cards are not sent by the community adapters

ckkashyap:
This PR addresses https://github.com/microsoft/botbuilder-js/issues/1806

OAuth prompts are not supported in the community adapters so they must not send OAuth cards. Instead, they must send sign-in cards.


iMicknl:
Is this not supported at all? It looks like OAuth is working in the JS Community Adapters, when you link it to ABS. https://github.com/BotBuilderCommunity/botbuilder-community-js/pull/149

It is still a work in progress, since I was waiting for some OAuth changes in core, however it looks like it is possible. Why block this use in core, instead of blocking it in the adapter?

PR #1813, opened 2020-02-28
Initial check-in of flatten, unique and

chrimc62:
Enable indicesAndValues to work over objects.
Add flatten(array, [depth]) which will flatten all arrays or to optional depth.
Add unique(array) which will remove duplicate items from array.

Fixes #

## Testing
Tests for all three fesatures.


boydc2014:
Looks good, not sure why the node-12 CI not passing, seems several prs are also not passing that one

PR #1824, opened 2020-03-01
Fix expression built-in function errors and test errors

Danieladu:
1. Add adaptive-expression into mocha test and test coverage
2. Fix errors in function `indicesAndValues`
3.fix several test errors


libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -1793,6 +1793,25 @@ export class ExpressionFunctions {
         return count;
     }
 
+    private static flattern(arr: any[], dept: number): any[]{
```
boydc2014: s/flattern/flatten

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2024,9 +2043,9 @@ export class ExpressionFunctions {
                 ExpressionType.Flatten,
                 ExpressionFunctions.apply(
                     args => {
-                        let array = args[0]
-                        let depth = args.length > 1 ? args[1] : 100
-                        return (array as any).flat(depth)
+                        let array = args[0];
+                        let depth = args.length > 1 ? args[1] : 100;
+                        return ExpressionFunctions.flattern(array, depth);
```
boydc2014: flatten

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -1793,6 +1793,25 @@ export class ExpressionFunctions {
         return count;
     }
 
+    private static flatten(arr: any[], dept: number): any[]{
+        dept = typeof dept === 'undefined' ? 1 : dept;
```
chrimc62: > ? 1  [](start = 42, length = 5)

This changes the meaning.  No depth should flatten all arrays, otherwise to depth.  This is different than the flat function, but is a better definition.

chrimc62:
Thanks for fixing this.  My fauilt.

---
In reply to: [593046110](https://github.com/microsoft/botbuilder-js/pull/1824#issuecomment-593046110) [](ancestors = 593046110)

PR #1872, opened 2020-03-07
clean up activityhandler invoke path

### LINKED ISSUE #1836: [PORT] clean up activityhandler and teamsactivityhandler
https://github.com/microsoft/botbuilder-dotnet/pull/3482

fixes #3446 

(duplicate code and new compiler warning introduced with sso fixes)

# Changed projects
* Microsoft.Bot.Builder
* Microsoft.Bot.Builder.Tests

johnataylor:
fixes #1836


libraries/botbuilder-core/src/activityHandler.ts
```diff
@@ -416,38 +400,44 @@ export class ActivityHandler extends ActivityHandlerBase {
...
      * and then continue by calling [defaultNextEvent](xref:botbuilder-core.ActivityHandler.defaultNextEvent).
     */
-    protected async onInvokeActivity(context: TurnContext): Promise<void|InvokeResponse> {
-        if(context.activity.name && (context.activity.name === verifyStateOperationName || context.activity.name === tokenExchangeOperationName)) {
-            await this.onSignInInvoke(context);
+    protected async onInvokeActivity(context: TurnContext): Promise<InvokeResponse> {
+        try {
+            if (context.activity.name && (context.activity.name === verifyStateOperationName || context.activity.name === tokenExchangeOperationName)) {
+                await this.onSignInInvoke(context);
+                return { status: 200 };
```
carlosscastro: There is a StatusCodes constant, you shold replace this for StatusCodes.OK

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -86,7 +86,7 @@ const OAUTH_ENDPOINT = 'https://api.botframework.com';
 const US_GOV_OAUTH_ENDPOINT = 'https://api.botframework.azure.us';
 
 // This key is exported internally so that the TeamsActivityHandler will not overwrite any already set InvokeResponses.
-export const INVOKE_RESPONSE_KEY: symbol = Symbol('invokeResponse');
+//export const INVOKE_RESPONSE_KEY: symbol = Symbol('invokeResponse');
```
swagatmishra2007: remove comment?

libraries/botbuilder-core/src/activityHandler.ts
swagatmishra2007: > return { status: 200 }; [](start = 15, length = 24)

if there is an oauthprompt in the dialog stack, that would have set an invokeResponse object in the turnstate dictionary for the Invoke_Response_key key, so returning an invokeresponse object here would either throw a key collision exception or worse, override the oauthprompt's invokeresponse with a 200 OK always.
Consider either restoring this back to return a void or checking if the Invoke_Response_key has been set to an invokeresponse. If already set, return that object otherwise return 200

johnataylor: the code in ActivityHandlerBase should handle that. Refer to line 70 in that file. The check you are looking for is there.

stevengum: @swagatmishra2007, the scenario you mention is handled by this section of [code](https://github.com/microsoft/botbuilder-js/pull/1872/files#diff-e465969a7fb5affdcc2195df6f6f4ab6R67-R72): 

```ts
// ActivityHandlerBase
case ActivityTypes.Invoke:
    const invokeResponse = await this.onInvokeActivity(context);
    // If onInvokeActivity has already sent an InvokeResponse, do not send another one.
    if (invokeResponse && !context.turnState.get(INVOKE_RESPONSE_KEY)) {
        await context.sendActivity({ value: invokeResponse, type: 'invokeResponse' });
    }
```

johnataylor: @swagatmishra2007 if you think this is still a problem, please open an issue, and highlight (or add) a test case that illustrates the problem, thanks.

johnataylor:
@stevengum are we both push nearly identical changes to this branch 🐳

johnataylor:
I'll wait for the build... not exactly sure which commit is building yours or mine ❓

PR #1876, opened 2020-03-07
Add expectedReplies, update skillDialog and botFrameworkAdapter

EricDahlvang:


libraries/botframework-connector/src/connectorApi/models/mappers.ts
```diff
@@ -956,6 +956,28 @@ export const ConversationsResult: msRest.CompositeMapper = {
   }
 };
 
+export const ExpectedReplies: msRest.CompositeMapper = {
```
stevengum: Not sure this is needed as the activities will come in a response to BotframeworkClient.postActivity()... but otherwise it looks good!

PR #1883, opened 2020-03-10
API naming for LG

### LINKED ISSUE #1879: [PORT] LG API changes
Port this PR - https://github.com/microsoft/botbuilder-dotnet/pull/3480

Danieladu:
fixes #1879
API naming for LG
1. Rename 'LGFile' to "Templates"
2. Rename "CreateActivity" in "ActivityFactory" to "FromObject"
3. Rename "SimpleMultiLangGenerator.cs " to "MultiLanguageLG.cs" and move into LG library,
  and remove LG library and ILanguageGenerator dependencies
4. Rename "EvaluateTemplate" to "Evaluate", and the original "Evaluate" is replaced by a new name "EvaluateText"


libraries/botbuilder-lg/src/templates.ts
```diff
@@ -6,39 +6,39 @@
...
+export class Templates {
 
     /// <summary>
     /// Gets or sets templates that this LG file contains directly.
     /// </summary>
     /// <value>
     /// templates that this LG file contains directly.
     /// </value>
-    public templates: LGTemplate[];
+    public items: Template[];
```
boydc2014: i know it's not typical to extends array<T> in JS/TS, so it LGTM to call this items. @tomlm any ideas.

tomlm: yeah the problem was we couldn't have property name same as class name in C#

---
In reply to: [390280511](https://github.com/microsoft/botbuilder-js/pull/1883#discussion_r390280511) [](ancestors = 390280511)

libraries/botbuilder-lg/src/templates.ts
```diff
@@ -6,39 +6,39 @@
...
-import { LGErrors } from './lgErrors';
-import { LGExtensions } from './lgExtensions';
+import { TemplateErrors } from './templateErrors';
+import { TemplateExtensions } from './templateExtensions';
 
 /// <summary>
 /// LG entrance, including properties that LG file has, and evaluate functions.
 /// </summary>
-export class LGFile {
+export class Templates {
```
tomlm: >Templates [](start = 13, length = 9)

you can simply do this:  
      class Templates extends Array<Template>

cleemullins:
A quick update on this: @tomlm and @stevenic are working to drive this to closure today, so we can push out the RC candidate of the JS SDK.

libraries/botbuilder-lg/src/templates.ts
```diff
@@ -119,6 +112,37 @@ export class LGFile {
...
+    }
+
+    /**
+     * Appends 1 or more templates to the collection.
+     * @param args List of templates to add.
+     */
+    public push(...args: Template[]): void {
+        args.forEach(t => this.items.push(t));
+    }
+
```
tomlm: Do you need a .length, or any other standard methods?

libraries/botbuilder-lg/tests/lg.test.js
```diff
@@ -9,89 +9,100 @@ function GetExampleFilePath(fileName) {
...
         assert.strictEqual(options.includes(evaled), true, `The result ${ evaled } is not in those options [${ options.join(',') }]`);
     });
 
     it('TestBasicTemplateReference', function() {
-        let LGFile = LGParser.parseFile(GetExampleFilePath('3.lg'));
-        console.log(LGFile.templates[0].body)
-        console.log(LGFile.templates[1].body)
-        let evaled = LGFile.evaluateTemplate('welcome-user', undefined);
+        let templates = Templates.parseFile(GetExampleFilePath('3.lg'));
+        console.log(templates.items[0].body);
```
tomlm: >items [](start = 30, length = 5)

I thought you removed this property?!?

libraries/botbuilder-lg/tests/lg.test.js
```diff
@@ -463,77 +474,77 @@ describe('LG', function() {
...
+        assert.strictEqual(templates.items[2].body.replace(/\r\n/g, '\n'), '- hi2 \n');
+
+        templates = templates.updateTemplate('newtemplate', 'newtemplateName', ['newage', 'newname'], '- new hi\r\n#hi');
+        assert.strictEqual(templates.items.length, 3);
+        assert.strictEqual(templates.imports.length, 0);
+        assert.strictEqual(templates.items[1].name, 'newtemplateName');
+        assert.strictEqual(templates.items[1].parameters.length, 2);
+        assert.strictEqual(templates.items[1].parameters[0], 'newage');
+        assert.strictEqual(templates.items[1].parameters[1], 'newname');
+        assert.strictEqual(templates.items[1].body.replace(/\r\n/g, '\n'), '- new hi\n- #hi\n');
```
tomlm: Same here, these all should be templates[1] instead of templates.items

libraries/botbuilder-lg/src/templates.ts
```diff
@@ -99,16 +92,16 @@ export class LGFile {
     /// </value>
     public options: string[];
 
-    public constructor(templates?: LGTemplate[],
-        imports?: LGImport[],
+    public constructor(items?: Template[],
```
tomlm: >items [](start = 23, length = 5)

templates as name?

libraries/botbuilder-lg/src/templates.ts
Stevenic: No... It's not a full array so I don't wont to break into jail by adding length

PR #1907, opened 2020-03-14
[OAuth]do not set signInLink to empty when OAuthAppCredentials is set

lzc850612:
Fixes #

## Description

When OAuthAppCredentials is passed in as part OAuthPromptSettings, do not set SignInLink to be empty as Emulator/other first party channels will ignore this and get a SignInLink without the additional OAuthAppCredentials setting.

## Specific Changes

  - 
  -
  -

### DIFF
--- a/libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts
+++ b/libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts
@@ -280,7 +280,12 @@ export class OAuthPrompt extends Dialog {
                 const signInResource = await (context.adapter as ExtendedUserTokenProvider).getSignInResource(context, this.settings.connectionName, context.activity.from.id, null, this.settings.oAuthAppCredentials);
                 let link = signInResource.signInLink;
                 const identity = context.turnState.get((context.adapter as BotAdapter).BotIdentityKey);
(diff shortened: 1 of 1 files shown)


libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts
```diff
@@ -280,7 +280,12 @@ export class OAuthPrompt extends Dialog {
                 const signInResource = await (context.adapter as ExtendedUserTokenProvider).getSignInResource(context, this.settings.connectionName, context.activity.from.id, null, this.settings.oAuthAppCredentials);
                 let link = signInResource.signInLink;
                 const identity = context.turnState.get((context.adapter as BotAdapter).BotIdentityKey);
-                if((identity && isSkillClaim(identity.claims)) || OAuthPrompt.isFromStreamingConnection(context.activity)) {
+                
+                // use the SignInLink when 
+                //   in speech channel or
+                //   bot is a skill or
+                //   an extra OAuthAppCredentials is being passed in
+                if((identity && isSkillClaim(identity.claims)) || OAuthPrompt.isFromStreamingConnection(context.activity) || this.settings.oAuthAppCredentials) {
```
carlosscastro: This 'if' should become a method given how complex it is, but I'm OK doing that for R9

PR #1911, opened 2020-03-16
Improve Expression with adding type checking functions

cosmicshuai:
1. Remove Array prebuilt-function. You can use createArray funtions.
2. Type Checking prebuilt functions:
isString, isInteger (*), isFloat, isArray, isObject, isDateTime (**), isBoolean

Notes:
\*  In JavaScript, when you call isInteger(1.0), it will true, so isFloat(1.0) will be false.
** isDateTime can will only return true when a ISO TimeStamp string is parsed


libraries/botbuilder-lg/tests/testData/examples/MultiLineExpr.lg
```diff
@@ -0,0 +1,9 @@
+>Demo How LG can evalute a multiple line defined expression
```
Danieladu: more kinds of test cases especially corner case
if there is no close bracket. 
```
# template
- ${name
- good
```

```
# template
- ${concat('hi', 
   jack)
```

```
# template
[Activity
  key = ${concat('1',
'2'} | hi
```

more...

libraries/botbuilder-lg/tests/testData/examples/MultiLineExpr.lg
```diff
@@ -0,0 +1,9 @@
...
+#definition
```
Danieladu: it would be better if we can add one or two test cases in Expression library

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2988,7 +2983,37 @@ export class ExpressionFunctions {
...
                     }),
                 ReturnType.Boolean,
-                ExpressionFunctions.validateIsMatch)
+                ExpressionFunctions.validateIsMatch),
+            
+            // Type Checking Functions
+            new ExpressionEvaluator(ExpressionType.isString, ExpressionFunctions.apply(
+                (args: any[]): boolean =>  typeof args[0] === 'string'),
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary),
```
Danieladu: not just `Unary`, the param's return type should be `string` or `object`.
we can consider to use 
```
validateUnaryString
```

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2988,7 +2983,37 @@ export class ExpressionFunctions {
...
+            new ExpressionEvaluator(ExpressionType.isInteger, ExpressionFunctions.apply(
+                (args: any[]): boolean =>  typeof args[0] === 'number' && args[0] % 1 === 0),
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary),
```
Danieladu: validateUnaryNumber

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2988,7 +2983,37 @@ export class ExpressionFunctions {
...
+            new ExpressionEvaluator(ExpressionType.isFloat, ExpressionFunctions.apply(
+                (args: any[]): boolean =>  typeof args[0] === 'number' && args[0] % 1 !== 0),
```
Danieladu: what's the difference between Number.isInteger and x % 1 !== 0, we can choose one from them

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2988,7 +2983,37 @@ export class ExpressionFunctions {
...
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary),
+            new ExpressionEvaluator(ExpressionType.isArray, ExpressionFunctions.apply(
+                (args: any[]): boolean => Array.isArray(args[0])),
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary),
```
Danieladu: ExpressionFunctions.validateArityAndAnyType(expression, 1, 1, ReturnType.Object);

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2988,7 +2983,37 @@ export class ExpressionFunctions {
...
+            new ExpressionEvaluator(ExpressionType.isObject, ExpressionFunctions.apply(
+                (args: any[]): boolean => typeof args[0] === 'object'),
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary),
```
Danieladu: ExpressionFunctions.validateArityAndAnyType(expression, 1, 1, ReturnType.Object);

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2988,7 +2983,37 @@ export class ExpressionFunctions {
...
+            new ExpressionEvaluator(ExpressionType.isBoolean, ExpressionFunctions.apply(
+                (args: any[]): boolean => typeof args[0] === 'boolean'),
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary),
```
Danieladu: ExpressionFunctions.validateUnaryBoolean

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2988,7 +2983,37 @@ export class ExpressionFunctions {
...
+            new ExpressionEvaluator(ExpressionType.isDateTime, ExpressionFunctions.apply(
+                (args: any[]): boolean => typeof args[0] === 'string' && this.verifyISOTimestamp(args[0]) === undefined),
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary)
```
Danieladu: validateUnaryString

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2988,7 +2983,37 @@ export class ExpressionFunctions {
...
-                ExpressionFunctions.validateIsMatch)
+                ExpressionFunctions.validateIsMatch),
+            
+            // Type Checking Functions
+            new ExpressionEvaluator(ExpressionType.isString, ExpressionFunctions.apply(
+                (args: any[]): boolean =>  typeof args[0] === 'string'),
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary),
+            new ExpressionEvaluator(ExpressionType.isInteger, ExpressionFunctions.apply(
+                (args: any[]): boolean =>  typeof args[0] === 'number' && args[0] % 1 === 0),
```
Danieladu: what's the difference between Number.isInteger and x % 1 !== 0, we can choose one from them

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -2988,7 +2983,37 @@ export class ExpressionFunctions {
...
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary),
+            new ExpressionEvaluator(ExpressionType.isFloat, ExpressionFunctions.apply(
+                (args: any[]): boolean =>  typeof args[0] === 'number' && args[0] % 1 !== 0),
+            ReturnType.Boolean,
+            ExpressionFunctions.validateUnary),
```
Danieladu: validateUnaryNumber

libraries/adaptive-expressions/src/expressionFunctions.ts
cosmicshuai: > not just `Unary`, the param's return type should be `string` or `object`.
> we can consider to use
> 
> ```
> validateUnaryString
> ```

My understanding is isString() can take any type argument, and check if the type of args[0] is 'string'. If validateUnaryString used here, once a number is parsed to the function. It will throw an error instead of return false.

Danieladu: ok make sense

libraries/adaptive-expressions/src/expressionFunctions.ts
Danieladu: use `ExpressionFunctions.isNumber` replace `typeof args[0] === 'number'`

libraries/adaptive-expressions/src/expressionFunctions.ts
Danieladu: use `ExpressionFunctions.isNumber` replace `typeof args[0] === 'number'`

Issue #1612, opened 2020-01-24
[botbuilder] If localTimestamp comes from a different time zone the original time zone is lost after conversion to date object


VictorGrycuk:
## Versions
What package version of the SDK are you using. **4.7.0**
What nodejs version are you using **v10.8.0**
What browser version are you using **none**
What os are you using **Windows 10 Pro**

## Describe the bug
The property [activity.localTimestamp](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder/src/botFrameworkAdapter.ts#L1421) does not retain the original time zone when the activity comes from a different time zone than where the bot is running.

When the [parseRequest](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder/src/botFrameworkAdapter.ts#L1415) method is executed, the property `activity.localTimestamp` is converted to `Date()` but the date time offset is lost as [JavaScript works solely on UTC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date).

We thought a couple solutions to fix or circumvent this problem:

1. Augment the date object to include the original timezone, either as [global namespace](https://stackoverflow.com/questions/40876478/how-do-you-add-a-property-to-an-existing-type-in-typescript) or [by exporting](https://github.com/microsoft/botframework-solutions/blob/master/sdk/typescript/libraries/botbuilder-skills/src/auth/msJWTAuthenticationProvider.ts#L7).
2. Keep `localTimestamp` as Date, but before the conversion extract the time zone and store it in [localTimezone property](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botframework-schema/src/index.ts#L336).
3. Remove the conversion to `Date` object and keep the original date as string. This will introduce a breaking change on user already operating on `localTimestamp` as Date.

Our **preferred** solution would be the first one as it will retain the Date compatibility while offering the expected information as in the .Net Library.

## To Reproduce

1. Go to the [Virtual Assistant's folder](https://github.com/microsoft/botframework-solutions/tree/master/templates/Virtual-Assistant-Template/typescript/samples/sample-assistant)
2. Install the dependencies with `npm install`
3. Compile the solution with `npm run build`
4. Deploy the Virtual Assistant using the following command:

```bash
[...]

PR #1924, opened 2020-03-18
Add localTimezoneOffset to Activity interface

EricDahlvang:
Fixes https://github.com/microsoft/botbuilder-js/issues/1612


libraries/botbuilder/src/activityValidator.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @module botbuilder
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { Activity } from 'botbuilder-core';
+
+export function validateActivity(resolve: any, activity: Activity): void{
```
stevengum: I'd prefer if the order of the parameters was switched, so instead of 
```ts
export function validateActivity(resolve: any, activity: Activity): void{
```

it would be
```ts
export function validateActivity(activity: Activity, resolve: any): void {
```

The Activity is the main bit we care about, and the resolve is something we use because of the original composition of the code.

libraries/botbuilder/src/activityValidator.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (typeof activity !== 'object') { throw new Error(`activityValidator.validateActivity(): invalid request body.`); }
+    if (typeof activity.type !== 'string') { throw new Error(`activityValidator.validateActivity(): missing activity type.`); }
```
stevengum: I would change the error message to use `validateActivity()` instead of `activityvalidator.validateActivity()`.

libraries/botbuilder/src/activityValidator.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (typeof activity.timestamp === 'string') { activity.timestamp = new Date(activity.timestamp); }
+    if (typeof activity.expiration === 'string') { activity.expiration = new Date(activity.expiration); }
+    if (typeof activity.localTimestamp === 'string') { 
+        // Since Javascript Date object is UTC, this code will pull the TimezoneOffset and put it
+        // in activity.localTimezoneOffset to preserve the value.
+        let [d, fullTime] = (activity.localTimestamp as string).split('T');
+        if(fullTime){
+            let [t, tz] = fullTime.split(/(?=[+-])/);
```
stevengum: These assignations can both `const [ ]` instead of `let [ ]`.

libraries/botbuilder/src/activityValidator.ts
stevengum: Actually, can you instead just have this function return an `Activity` instead of void, so in the calling code it looks like this

```ts
if (req.body) {
    try {
        const activity = validateActivity(req.body);
        resolve(activity);
    } catch (err) {
        reject(err);
    }
} else {
    let requestData = '';
    req.on('data', (chunk: string): void => {
        requestData += chunk;
    });
    req.on('end', (): void => {
        try {
            req.body = JSON.parse(requestData);
            const activity = validateActivity(req.body);
            resolve(activity);
        } catch (err) {
            reject(err);
        }
    });
}
```

EricDahlvang:
> Not a terrific fan of adding something to the Activity schema...

I'm not a fan of this either.  It's a 'one-off' for javascript, because of the UTC Date object behavior.  If you have some other ideas, let's discuss.

libraries/botbuilder/src/activityValidator.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { Activity, ActivityTimestamps } from 'botbuilder-core';
+
+export function validateActivity(activity: Activity): Activity {
+    if (typeof activity !== 'object') { throw new Error(`validateActivity(): invalid request body.`); }
+    if (typeof activity.type !== 'string') { throw new Error(`validateActivity(): missing activity type.`); }
+    if (typeof activity.timestamp === 'string') {
+        (activity as ActivityTimestamps).rawTimestamp = activity.timestamp;
```
johnataylor, line 15: It's a little weird to have a function called "validate" go ahead and change the thing its validating.
Validate _and fix_ might be more appropriate.

EricDahlvang: Good point.  Changed to validateAndFixActivity

PR #1958, opened 2020-03-30
[Expression] Add support of create object via {key: value} syntax

cosmicshuai:
Now in LG and Expression, user can create an object by {key, value, ...}
For example:
{user: "Wilson"}, {user: {age: 27}}
Also, the value can be some variable defined in the scope, 
For example, if the scope is {userName: "Allen"}
{user: userName}   will be evaluated as {user: "Allen"}


libraries/adaptive-expressions/src/parser/ExpressionAntlrLexer.g4
```diff
@@ -65,18 +72,18 @@ NEWLINE : '\r'? '\n' -> skip;
...
-CONSTANT : ('{' WHITESPACE* '}');
-
 INVALID_TOKEN_DEFAULT_MODE : . ;
 
 mode STRING_INTERPOLATION_MODE;
 
 STRING_INTERPOLATION_END : '`' {this.ignoreWS = true;} -> type(STRING_INTERPOLATION_START), popMode;
 
-TEMPLATE : '$' '{' (STRING | ~[\r\n{}'"])*? '}';
+EMPTY_OBJECT: '{' WHITESPACE* '}';
```
Danieladu: OBJECT_DEFINITION should wrap empty-object

libraries/adaptive-expressions/src/parser/expressionParser.ts
```diff
@@ -130,6 +129,27 @@ export class ExpressionParser implements ExpressionParserInterface {
...
+                    const node = kvPair.key().children[0];
+                    if (node instanceof TerminalNode) {
+                        if (node.symbol.type === ep.ExpressionAntlrParser.IDENTIFIER) {
+                            key = node.text;
+                        } else {
+                            key = node.text.substring(1, node.text.length - 1);
+                        }
+                    } 
+                    
+                    expr = this.makeExpression(ExpressionType.SetProperty, expr, new Constant(key), this.visit(kvPair.expression()));
```
Danieladu, line 146: we should consider if there exists the same keys and what should happen.
Throw exception, or just simply use `SetProperty` to override the previous one.

PR #1962, opened 2020-03-30
Add @microsoft/api-extractor to detect public API changes

### LINKED ISSUE #1903: Use @microsoft/api-extractor
Current integration testing has proven insufficient in avoiding breaking changes, and we've had cases were files were not properly exported. Using api-extractor can help with avoiding these scenarios:

image

https://www.npmjs.com/package/@microsoft/api-extractor

https://api-extractor.com/

stevengum:
Fixes #1903

>_Partial fix, `api-extractor` has not been enabled for botframework-connector, adaptive-expressions, and botbuilder-lg._

## Description
Using `@microsoft/api-extractor` enables us to easily identify public API changes and avoid breaking changes.

Developers should now use use `build:rollup` to generate a source-controlled Markdown file that contains the public API of a package for a human friendly experience in reviewing API changes.

Changing the public API and running `build:rollup` should result in updates to the package's `.api.md` file.

### Example: a developer changes the constructor for TestAdapter in botbuilder-core.
When they run `build:rollup`, it should result in **`botbuilder-core/etc/botbuilder-core.api.md`** being changed.
These changes should be committed before the PR is submitted for review, otherwise the CI will fail.

For an example PR that fails the API-Check, see #1963.
___

## Specific Changes
  - update `.gitattributes` to support JSON with comments
  - update `.gitignore` to ignore `temp/` and `dist/` folders 
  - add `"@microsoft/api-extractor": "^7.7.10"` as a dependency for all libraries.
  - add two npm scripts to supported libraries:
     - `"build:rollup"` - clears current `lib/`, rebuilds and then runs `api-extractor`
         - **Run by developers before committing changes**
         - `api-extractor run --verbose --local` creates a new `<package-name>.api.md` in `etc/` which is checked into source control
         - `<package-name>.api.md` contains the public APIs and is used to highlight any breaking changes
     - `"test:compat"` - executes `api-extractor run --verbose`
         - **Run by CI machines**
         - If this task fails, the changes contains public API changes and the updates from `build:rollup` committed and added to the PR

## Testing
Created a ADO Pipeline to call `lerna run test:compat` which fails on public API changes without a checked-in `<package-name>.api.md` file.


stevengum:
@BruceHaley as FYI.

The ADO pipeline that specifically checks the public APIs is BotBuilder-JS-API-Check. (Build [#116581](https://fuselabs.visualstudio.com/SDK_v4/_build/results?buildId=116581&view=results))

Issue #1966, opened 2020-03-31
[PORT] Update IMemory interface and tryEvaluate for expression


tomlm:
> Port this change from botbuilder-dotnet/master branch:
https://github.com/microsoft/botbuilder-dotnet/pull/3635

Add an optional parameter `options` in tryEvaluate.

Add an optional parameter `allowInsubstitution` in getValue method of IMemory. 

In any boolean context, use raw value, in any other context, use substituted value. 

Fix part of #3625 

Will do a follow up for LG strict mode improvement. 

# Changed projects
* AdaptiveExpressions
* Microsoft.Bot.Builder.Dialogs.Adaptive
* AdaptiveExpressions.Tests

PR #1967, opened 2020-03-31
[PORT] Update IMemory interface and tryEvaluate for expression

cosmicshuai:
fixes: #1966 

Add an optional parameter options in tryEvaluate.

Inside options, we allows user to set a callback function which been called whenever there is a null found in memory path, except for those in boolean context.

### DIFF
--- a/libraries/adaptive-expressions/src/expressionFunctions.ts
+++ b/libraries/adaptive-expressions/src/expressionFunctions.ts
@@ -1010,11 +1010,11 @@ export class ExpressionFunctions {
 
     private static wrapGetValue(state: MemoryInterface, path: string, options: Options): any {
         let result = state.getValue(path);
-        if (result !== undefined) {
+        if (result !== undefined && result !== null) {
             return result;
         }
 
-        if(options && options.nullSubstitution !== undefined) {
+        if(options.nullSubstitution !== undefined) {
             return options.nullSubstitution(path);
         }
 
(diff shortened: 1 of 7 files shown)


libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -1005,16 +1008,31 @@ export class ExpressionFunctions {
         return {value: result, error};
     }
 
-    private static extractElement(expression: Expression, state: MemoryInterface): { value: any; error: string } {
+    private static wrapGetValue(state: MemoryInterface, path: string, options: Options): any {
+        let result = state.getValue(path);
+        if (result !== undefined) {
```
boydc2014, line 1013: do we only to this for "undefined", how about "null", i guess i'm not sure the convention is null or undefined

libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -1005,16 +1008,31 @@ export class ExpressionFunctions {
...
+            return result;
+        }
+
+        if(options && options.nullSubstitution !== undefined) {
```
boydc2014, line 1017: can you try to comment out options, and just leave options.nullSubstitution, see, where options might not exisits?

libraries/adaptive-expressions/src/expressionFunctions.ts
cosmicshuai: > do we only to this for "undefined", how about "null", i guess i'm not sure the convention is null or undefined

will change to if(result), so it can handle both undefined and null

libraries/adaptive-expressions/src/expressionFunctions.ts
cosmicshuai: > can you try to comment out options, and just leave options.nullSubstitution, see, where options might not exisits?

The reason is the initialization is not consistent with dotnet. Will fix.

PR #1980, opened 2020-04-02
Remove "botbuilder-core" dependency from LG

Danieladu:
Originally LG depends on "botbuilder-core" to achieve the bot schema to build the Activity througth `ActivityFactory`.
But now, `ActivityFactory` has been moved to "botbuilder-core" library, so, there is no need to depends on any bot library in LG. Which would keep LG as a pure and independent library.


Danieladu:
coverage dropped 2.4% currently, but we will migrate the tests into AdaptiveDialog when it is ready.
So, the dropping of coverage is temporary.

PR #1981, opened 2020-04-02
[Port] make custom function easier

### LINKED ISSUE #1931: [PORT] make custom function easier
https://github.com/microsoft/botbuilder-dotnet/pull/3595

Make customized function in LG and expression easier by accepting just a lambda. 

At the cost of expose FunctionTable instead of a generic IDictionary.

Now you can do

```C#
            Expression.Functions.Add("contoso.sqrt", (args) =>
            {
                object retValue = null;
                if (args[0] != null)
                {
                    double dblValue;
                    if (double.TryParse(args[0], out dblValue))
                    {
                        retValue = Math.Sqrt(dblValue);
                    }
                }

                return retValue;
            });
```
to add a new function

# Changed projects
* AdaptiveExpressions
* Microsoft.Bot.Builder.LanguageGeneration.Tests

cosmicshuai:
Fixes #1931 

Make customized function in LG and expression easier by accepting just a lambda.

At the cost of expose FunctionTable instead of a generic IDictionary.

Now you can do
```JavaScript
Expression.functions.add('contoso.sqrt', (args) => {
            let result = null;
            if (args[0] !== undefined ) {
                const inputFloat = parseFloat(args[0]);
                result = Math.sqrt(inputFloat);
            }

            return result;
        });
```

to add a new function

### DIFF
--- a/libraries/adaptive-expressions/src/functionTable.ts
+++ b/libraries/adaptive-expressions/src/functionTable.ts
@@ -58,17 +58,20 @@ export class FunctionTable implements Map<string, ExpressionEvaluator> {
 
     }
 
-    public add(item: {key: string; value: ExpressionEvaluator} | string, value?: ExpressionEvaluator | customFunction): void{
+    public add(item: {key: string; value: ExpressionEvaluator}): void;
+    public add(key: string, value: ExpressionEvaluator): void;
+    public add(key: string, value: customFunction): void;
+    public add(param1: {key: string; value: ExpressionEvaluator} | string, param2?: ExpressionEvaluator | customFunction): void{
         if(arguments.length === 1) {
-            if (item instanceof Object) {
-                this.set(item.key, item.value);
+            if (param1 instanceof Object) {
+                this.set(param1.key, param1.value);
             }
         } else {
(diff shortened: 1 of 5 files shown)


libraries/adaptive-expressions/src/functionTable.ts
```diff
@@ -0,0 +1,105 @@
+/**
```
Danieladu, line 1: export it in index.ts

libraries/adaptive-expressions/src/functionTable.ts
```diff
@@ -0,0 +1,105 @@
...
+        if(ExpressionFunctions.standardFunctions.get(key)) {
+            throw Error(`You can't overwrite a built in function.`);
+        }
+
+        this.customFunctions.set(key, value);
+        return this;
+
+    }
+
+    public add(item: {key: string; value: ExpressionEvaluator} | string, value?: ExpressionEvaluator | customFunction): void{
```
Danieladu, line 61: Maybe use functions overloads is more readable?
https://www.typescriptlang.org/docs/handbook/functions.html#overloads

PR #1984, opened 2020-04-02
Allow creation of multiple Cosmos containers

### LINKED ISSUE #1975: Cannot use different Cosmos DB containers
## Versions
What package version of the SDK are you using. 4.8.0
What nodejs version are you using: 10.18.1

## Describe the bug
Cannot use different Cosmos DB containers because of global `_doOnce` in cosmosDbPartitionedStorage.ts.

## To Reproduce
```javascript
const storage = new CosmosDbPartitionedStorage({
				cosmosDbEndpoint: process.env.DB_SERVICE_ENDPOINT,
				authKey: process.env.DB_AUTH_KEY,
				databaseId: process.env.DATABASE_ID,
				containerId: process.env.CONTAINER,
			})
const otherStorage = new CosmosDbPartitionedStorage({
				cosmosDbEndpoint: process.env.DB_SERVICE_ENDPOINT,
				authKey: process.env.DB_AUTH_KEY,
				databaseId: process.env.DATABASE_ID,
				containerId: process.env.OTHER_CONTAINER,
			})
otherStorage.write(...)
```

## Expected behavior
Data gets written to the other container.

## Actual behavior
Data gets written to the first container.

[bug]

I might submit a PR to correct and test this.

mdrichardson:
Fixes #1975 

## Description

`doOnce` was preventing `cosmosDbPartitionedStorage` users from using multiple containers and/or databases. This is because `doOnce` acted on `cosmosDbPartitionedStorage` as a whole, rather than individual db/container combinations.

## Specific Changes

- Store a dict of `doOnce`-awaited functions, keyed by db+container
## Testing

Added two tests, which fail without this change and pass with it.

# Considerations

Is this the right approach, or would it be better to store the container as a property in `cosmosDbPartitionedStorage`? My hunch is that there might be concurrency issues with the class property approach.


libraries/botbuilder-azure/src/doOnce.ts
```diff
@@ -7,13 +7,20 @@
...
+        [dbAndContainerKey: string]: Promise<T>;
+    } = {};
 
-    public waitFor(fn: () => Promise<T>): Promise<T> {
-        if (!this.task) {
-            this.task = fn();
+    public waitFor(dbAndContainerKey: string, fn: () => Promise<T>): Promise<T> {
+        if (!this.task[dbAndContainerKey]) {
+            // Surround in a try/catch so that if container creation fails, it can retry and succeed later.
+            try {
```
EricDahlvang, line 17: Should this be something like:
```
            try {
                this.task[dbAndContainerKey] = fn();
            } catch (err) {
                delete this.task[dbAndContainerKey];
                console.error(err);
            }
```
So the Task is removed, and future callers can re-add it.

libraries/botbuilder-azure/src/doOnce.ts
```diff
@@ -7,13 +7,15 @@
...
 export class DoOnce<T> {
-    private task: Promise<T>;
+    private task: {
+        [dbAndContainerKey: string]: Promise<T>;
+    } = {};
 
-    public waitFor(fn: () => Promise<T>): Promise<T> {
-        if (!this.task) {
-            this.task = fn();
+    public waitFor(dbAndContainerKey: string, fn: () => Promise<T>): Promise<T> {
```
juharris, line 14: There's nothing really specific in this class about databases and containers so how about just calling the variable `key`?

PR #1985, opened 2020-04-02
Move Error logic out of botbuilder into authentication layer

### LINKED ISSUE #1445: [bf-connector] Cleanup Errors in authentication code
In `botframework-connector/authentication` we return normal JS errors with non-normalized error messages.

This has led to manual `Error.message` parsing in determining the correct HTTP Status code to send back to the caller of the bot. 

See https://github.com/microsoft/botbuilder-js/pull/1433#discussion_r349991588.

Zerryth:
Fixes #1445 

## Description
See [discussion](https://github.com/microsoft/botbuilder-js/pull/1433#discussion_r349991588) in comment.
* Create custom error class that throws richer errors, so we do not have to rely on parsing the error message to determine the correct status code.
* Move the logic out of botbuilder into the authentication layer

## Specific Changes
  - Create `IStatusCodeError` in botbuilder-schema. 
      - Currently there's a custom `StatusCodeError` class built in botbuilder, however we can't use it in the auth layers in botframework-connector 
      - This dependency would break browser compatibility and would be a circular dependency
      - Alternatively moving into botframework-schema breaks backwards compatibility
  - Created `AuthenticationError` class (ctor needs status code and message) in botframework-connector
     - Auth layers that activate when `BotFrameworkAdapter.authenticateConnection()` is called now throw this custom error, so we rely less heavily on string-parsing to determine status code to throw. (Throwing errors w/status code already done in [other parts of SDK](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder/src/channelServiceHandler.ts#L169) as well):. 
         - `JwtTokenValidation`
         - Channel Validation classes: `ChannelValidation`, `EmulatorValidation`, `GovernmentChannelValidation`, `SkillValidation`
         - `JwtTokenExtractor`
         - `EndorsementsValidation`
 - Made helper functions to determine auth code if the Adapter catches an error that does not have a status code, as fallback 
    - (such as errors thrown from 3rd party libraries like `jsonwebtoken` that handles validating tokens with Channel Service's asymmetical signatures)
    - Note: TypeScript doesn't allow you to [catch errors of a specific type](https://github.com/microsoft/TypeScript/issues/20024) :sob:


libraries/botframework-connector/src/auth/authenticationError.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import { IStatusCodeError, StatusCodes } from 'botframework-schema';
+
+export type StatusCode = number;
+
+export class AuthenticationError extends Error implements IStatusCodeError {
+    constructor(
+        public readonly message: string,
```
Stevenic, line 11: >public readonly [](start = 8, length = 15)

Error should already have a message property. Don't think you need to redefine it here.

Zerryth: Ah, that's right. Removed redefinition 👍 git

Zerryth:
Hmm, chrome tests are breaking in the build, and I don't believe it has anything to do with the PR changes...reached out to Bruce regarding that aspect.

Thanks @Stevenic for reviewing! (Woohoo a PR review finally 😄)
@carlosscastro Steve took a look at the PR, but said that it'd be best if you looked into too before merging though, please

libraries/botframework-connector/src/auth/jwtTokenExtractor.ts
```diff
@@ -94,7 +96,7 @@ export class JwtTokenExtractor {
         const keyId: string = decodedToken.header.kid;
         const metadata: any = await this.openIdMetadata.getKey(keyId);
         if (!metadata) {
-            throw new Error('Signing Key could not be retrieved.');
+            throw new AuthenticationError('Signing Key could not be retrieved.', StatusCodes.NOT_FOUND);
```
carlosscastro, line 99: Does this end in an unauthorized request? If yes, UNAUTHORIZED could be a more appropriate status code.

carlosscastro: i.e. if the request ends here, we want to return an http unauthorized and not a 400.

libraries/botframework-connector/src/auth/jwtTokenExtractor.ts
```diff
@@ -106,20 +108,29 @@ export class JwtTokenExtractor {
             if (Array.isArray(endorsements) && endorsements.length !== 0) {
                 const isEndorsed: boolean = EndorsementsValidator.validate(channelId, endorsements);
                 if (!isEndorsed) {
-                    throw new Error(`Could not validate endorsement for key: ${ keyId } with endorsements: ${ endorsements.join(',') }`);
+                    throw new AuthenticationError(
+                        `Could not validate endorsement for key: ${ keyId } with endorsements: ${ endorsements.join(',') }`,
+                        StatusCodes.BAD_REQUEST
```
carlosscastro, line 113: consider UNAUTHORIZED

libraries/botframework-connector/src/auth/jwtTokenExtractor.ts
```diff
@@ -106,20 +108,29 @@ export class JwtTokenExtractor {
...
                 }
 
                 // Verify that additional endorsements are satisfied. If no additional endorsements are expected, the requirement is satisfied as well
                 const additionalEndorsementsSatisfied = requiredEndorsements.every(endorsement => EndorsementsValidator.validate(endorsement, endorsements));
 
                 if (!additionalEndorsementsSatisfied) {
-                    throw new Error(`Could not validate additional endorsement for key: ${keyId} with endorsements: ${requiredEndorsements.join(',')}. Expected endorsements: ${requiredEndorsements.join(',')}`);
+                    throw new AuthenticationError(
+                        `Could not validate additional endorsement for key: ${keyId} with endorsements: ${requiredEndorsements.join(',')}. Expected endorsements: ${requiredEndorsements.join(',')}`,
+                        StatusCodes.BAD_REQUEST
```
carlosscastro, line 123: consider UNAUTHORIZED

libraries/botframework-connector/src/auth/jwtTokenExtractor.ts
```diff
@@ -106,20 +108,29 @@ export class JwtTokenExtractor {
...
+                    );
                 }
             }
 
             if (this.tokenValidationParameters.algorithms) {
                 if (this.tokenValidationParameters.algorithms.indexOf(decodedToken.header.alg) === -1) {
-                    throw new Error(`"Token signing algorithm '${ decodedToken.header.alg }' not in allowed list`);
+                    throw new AuthenticationError(
+                        `"Token signing algorithm '${ decodedToken.header.alg }' not in allowed list`,
+                        StatusCodes.BAD_REQUEST
```
carlosscastro, line 132: Consider UNAUTHORIZED

libraries/botframework-connector/tests/auth.test.js
```diff
@@ -15,6 +16,7 @@ describe('Bot Framework Connector - Auth Tests', function() {
                     throw new Error('Expected validation to fail.');
                 } catch (err) {
                     assert(err.message === `'authHeader' required.`, `unexpected error thrown: "${ err.message }"`);
+                    assert.equal(err.statusCode, 400);
```
carlosscastro, line 19: use bad request constant?

libraries/botframework-connector/src/auth/openIdMetadata.ts
```diff
@@ -51,11 +55,11 @@ export class OpenIdMetadata {
                 this.lastUpdated = new Date().getTime();
                 this.keys = (await getKeyResponse.json()).keys as IKey[];
             } else {
-                throw new Error(`Failed to load Keys: ${ getKeyResponse.status }`);
+                throw new AuthenticationError(`Failed to load Keys: ${ getKeyResponse.status }`, StatusCodes.BAD_REQUEST);
             }
 
         } else {
-            throw new Error(`Failed to load openID config: ${ res.status }`);
+            throw new AuthenticationError(`Failed to load openID config: ${ res.status }`, StatusCodes.NOT_FOUND);
```
carlosscastro, line 62: This is more that we failed to retrieve relevant information while trying to validate. Maybe 500, internal server error?

libraries/botframework-connector/src/auth/openIdMetadata.ts
```diff
@@ -51,11 +55,11 @@ export class OpenIdMetadata {
                 this.lastUpdated = new Date().getTime();
                 this.keys = (await getKeyResponse.json()).keys as IKey[];
             } else {
-                throw new Error(`Failed to load Keys: ${ getKeyResponse.status }`);
+                throw new AuthenticationError(`Failed to load Keys: ${ getKeyResponse.status }`, StatusCodes.BAD_REQUEST);
```
carlosscastro, line 58: If we are here is because we failed to load information relevant to the validation but it is not related to the request, and more to an error while trying to reach the intercom open id metadata endpoint. 500 might be more appropriate since the error is not due to a bad request.

libraries/botframework-connector/src/auth/endorsementsValidator.ts
```diff
@@ -24,7 +28,7 @@ export class EndorsementsValidator {
         }
 
         if (endorsements === null) {
-            throw new Error('endorsements required');
+            throw new AuthenticationError('endorsements required', StatusCodes.BAD_REQUEST);
```
carlosscastro, line 31: Unauthorized preferred here. This means that the token was not endorsed for the channel, which is a must and part of the auth validation.

libraries/botframework-connector/src/auth/jwtTokenExtractor.ts
Zerryth: Hmm...
I suppose it would _ultimately_ end up being unauthorized, because even though we can `decode` the token initially and get a decoded token, `decode` doesn't verify if the JWT signature is valid. (decode to find out what key we need, then we call `verify` to actually get a decoded token while also verifying if signature is valid)

So I guess trying to just `getKey(keyId)` might be not found (we're just asking if the metadata has a key with the specific id), but the ultimate result in the encompassing `validateToken` method semantically should be that the client is unauthorized.

Changing to UNAUTHORIZED 👍

libraries/botframework-connector/src/auth/skillValidation.ts
```diff
@@ -123,7 +125,7 @@ export namespace SkillValidation {
         channelId: string,
         authConfig: AuthenticationConfiguration): Promise<ClaimsIdentity> {
         if (!authConfig) {
-            throw new Error('SkillValidation.authenticateChannelToken(): invalid authConfig parameter');
+            throw new AuthenticationError('SkillValidation.authenticateChannelToken(): invalid authConfig parameter', StatusCodes.BAD_REQUEST);
```
stevengum, line 128: This looks incorrect. The `authConfig` is part of the bot's configuration, it's not received over the wire. Throwing an `INTERNAL_SERVER_ERROR` would be correct here.

libraries/botframework-connector/tests/auth.test.js
stevengum: @Zerryth there are additional tests in this PR (in `botbuilder`) that should also use the constant.

libraries/botframework-connector/tests/auth.test.js
```diff
@@ -49,6 +52,7 @@ describe('Bot Framework Connector - Auth Tests', function() {
                     throw new Error('Expected validation to fail.');
                 } catch (err) {
                     assert(err.message.substring('Unauthorized. Invalid AppId passed on token:'), `unexpected error thrown: "${ err.message }"`);
+                    assert.strictEqual(err.statusCode, StatusCodes.UNAUTHORIZED);;
```
stevengum, line 55: double semicolons seems excessive 😄

libraries/botframework-connector/tests/auth.test.js
```diff
@@ -424,5 +446,56 @@ describe('Bot Framework Connector - Auth Tests', function() {
                 assert(isValid);
             });
         });
+
+        describe('AuthenticationError', () => {
```
stevengum, line 450: Please put these tests in another file... I don't want more tests going into `auth.test.js`

libraries/botframework-connector/tests/auth.test.js
Zerryth: LOL

Zerryth: DOUBLE THE POWER?

libraries/botframework-connector/src/auth/authenticationError.ts
```diff
@@ -0,0 +1,44 @@
...
+        return !!(err && err.statusCode && typeof err.statusCode === "number");
+    }
+
+    /**
+     * Used to determine a status code from the error message for non-`IStatusCodeError`'s.
+     * @param err The error thrown, used to determine an appropriate status code.
+     */
+    public static determineStatusCodeAndBuildMessage(err: any): string {
+        let errMessage: string = (err && err.message) ? err.message : 'Internal Server Error';
+        let code: number = AuthenticationError.determineStatusCode(errMessage);
```
stevengum, line 27: These should be `const`

libraries/botframework-connector/src/auth/authenticationError.ts
```diff
@@ -0,0 +1,44 @@
...
+export class AuthenticationError extends Error implements IStatusCodeError {
+    constructor(
+        message: string,
+        public readonly statusCode: StatusCode
+    ) {
+        super(message);
+    }
+
+    public static isStatusCodeError(err: any): err is IStatusCodeError {
+        return !!(err && err.statusCode && typeof err.statusCode === "number");
```
stevengum, line 18: This can be simplified to 

```ts
return !!(err && typeof err.statusCode === 'number');
```

libraries/botframework-connector/src/auth/authenticationError.ts
```diff
@@ -0,0 +1,44 @@
...
+    public static determineStatusCodeAndBuildMessage(err: any): string {
+        let errMessage: string = (err && err.message) ? err.message : 'Internal Server Error';
+        let code: number = AuthenticationError.determineStatusCode(errMessage);
+        const connectionHeader = `Connection: 'close'\r\n`;
+        
+        return `HTTP/1.1 ${ code } ${ StatusCodes[code] }\r\n${ errMessage }\r\n${ connectionHeader }\r\n`;
+    }
+    
+    private static determineStatusCode(message: string): StatusCode {
+        if (typeof(message) === 'string') {
```
stevengum, line 34: `typeof message === 'string'`

libraries/botframework-connector/tests/auth.test.js
Zerryth: wowza yeah there were a lot of numbers in botbuilder tests that can now be replaced with more readable `StatusCode` enum 

updated them to do so

libraries/botframework-connector/tests/auth.test.js
Zerryth: moved to `authenticationError.test.js`

Zerryth:
@stevengum @carlosscastro 
* Changed status codes thrown with the `AuthenticationErrors` pointed out
* `AuthenticationError` unit tests are now separated into `authenticationError.test.js`
* Replaced inexplicit numbers in unit tests with `StatusCodes` enum values instead

LMK if anything else is needed 😀

PR #2015, opened 2020-04-13
[Expression] Support byte array input for base64 function

cosmicshuai:
fixes: #2005 
Now base64 built-in function can take byte array input.
if the input is byte array, it will use Convert.ToBase64String directly, otherwise, the input will be converted to string first, then to byte array.

base64ToBinary will convert a base64 string to byte array


libraries/adaptive-expressions/package.json
```diff
@@ -24,6 +24,7 @@
     "@types/moment-timezone": "^0.5.12",
     "@types/xmldom": "^0.1.29",
     "antlr4ts": "0.5.0-alpha.1",
+    "atob": "^2.1.2",
```
Danieladu, line 27: need `@types/atob`

Issue #1574, opened 2020-01-13
telemetryLoggerMiddleware does not log Adaptive Card responses


nathanlindorff:
**Is your feature request related to a problem? Please describe.**
I am using the Virtual Assistant template which uses the telemetryLoggerMiddleware. After using it for a while, it became apparent when Adaptive cards are being used as a reply they are not being picked up by the telemetry logger. Only `text` and `speak` are being populated into the event properties in the [`fillSendEventProperties`](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/telemetryLoggerMiddleware.ts#L229) method.

_[telemetryLoggerMiddleware.ts line 243](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/telemetryLoggerMiddleware.ts#L243)_
```
if (activity.text && activity.text.trim()) {
    properties[this.telemetryConstants.textProperty] = activity.text;
}

if (activity.speak && activity.speak.trim()) {
    properties[this.telemetryConstants.speakProperty] = activity.speak;
}
```

**Describe the solution you'd like**
I would like TelemetryLoggerMiddleware, or a variant of it, to log attachments like Adaptive cards. This could either be by default if detected, or by configuration.

**Describe alternatives you've considered**
To work around this issue, I could make my own version of the telemetryLoggerMiddleware, and may need to do so in the short term. In my option it would be better to have it as a part of the framework, as using Adaptive Cards is a documented way of sending messages. These should be captured.

[enhancement]

v-kydela:
@nathanlindorff - I'm a little confused because you have a link that's labeled `fillUpdateEventProperties` but it links to `fillReceiveEventProperties` and then you've posted a code snippet from `fillSendEventProperties`. Can you explain why?

nathanlindorff:
PEBKAC cut and paste error sorry. I'll fix it up. It all should point to fillSendEventProperties

nathanlindorff:
What I ended up doing in my scenario was subclassing TelemetryLoggerMiddleware and overriding `fillSendEventProperties`.

I added a new 'attachment' property to which I wrote the stringified version of the attachment. I also did some specific work to extract the text from the Adaptive card into a readable form. The latter was more specific to my cards. In retrospect unless there is a standard way to turn cards to text this may be the better option than framework changes. 

Perhaps a documentation update would be more appropriate, indicating that if Cards are used, the telemetry middleware doesn't log any details.

[...]

PR #2025, opened 2020-04-14
Add attachments to telemetry

garypretty:
Addresses #1574 

Logs attachments on outgoing activities in telemetry, if the LogPersonalInformation flag is set to true.

### DIFF
--- a/libraries/botbuilder-core/src/telemetryConstants.ts
+++ b/libraries/botbuilder-core/src/telemetryConstants.ts
@@ -15,4 +15,5 @@ export class TelemetryConstants {
     public readonly textProperty: string = 'text';
     public readonly speakProperty: string = 'speak';
     public readonly userIdProperty: string = 'userId';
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-core/src/telemetryConstants.ts
```diff
@@ -15,4 +15,5 @@ export class TelemetryConstants {
     public readonly textProperty: string = 'text';
     public readonly speakProperty: string = 'speak';
     public readonly userIdProperty: string = 'userId';
+    public readonly attachmentsProperty: string = 'attachments';
```
stevengum, line 18: These class properties should have been static like in [C#](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder/TelemetryConstants.cs). From a cursory glance on the class' references developers shouldn't need to instantiate this class to access these values.

PR #2044, opened 2020-04-15
Bot Change Detection

Stevenic:
Fixes https://github.com/microsoft/botframework-sdk/issues/5495

## Description
These changes add logic to automatically detect that a bots dialogs have changed since a persisted dialog stack was started.  A hash code is stored with each persisted container dialog that is then checked on each subsequent execution. 

When a change is detected a new 'dialogChanged' event is raised which the bot can optionally handle to deal with the change in some custom way. If that event isn't handled and error will be thrown which will trigger the bots standard error handling logic.

note: `WaterfallDialog` does not currently support change detection as there's no great way to do this. You can check for the number of steps to change but that's about it. 

## Specific Changes
- Added new `DialogSetchangeHash` property.
- Added new `ContainerDialog.checkForChanges()` method.
- Updated `ComponentDialog` to call new method.
- Updated `AdaptiveDialog` to call new method.

## Testing
Added unit tests to cover the new feature.


libraries/botbuilder-dialogs/tests/componentDialog.test.js
```diff
@@ -347,8 +347,95 @@ describe('ComponentDialog', function () {
...
+
+        const adapter = new TestAdapter(async turnContext => {
+            const dc = await dialogs.createContext(turnContext);
+            const results = await dc.continueDialog();
+
+            if (results.status === DialogTurnStatus.empty) {
+                await dc.beginDialog('test');
+            } else {
+                assert(results.status === DialogTurnStatus.complete, `results.status should be 'complete' not ${ results.status }`);
+                assert(results.result === undefined, `results.result should be undefined, not ${ results.result }`);
```
stevengum, line 380: To avoid having to add descriptive error messages, this could be simplified to

```suggestion
                assert.strictEqual(results.status, DialogTurnStatus.complete);
                assert.strictEqual(results.result, undefined);
```

libraries/botbuilder-dialogs-adaptive/src/adaptiveDialog.ts
```diff
@@ -143,6 +143,8 @@ export class AdaptiveDialog<O extends object = {}> extends DialogContainer<O> {
     public async beginDialog(dc: DialogContext, options?: O): Promise<DialogTurnResult> {
         this.onPushScopedServices(dc.context);
         try {
+            await this.checkForChanges(dc);
```
tomlm, line 146: >    await this.checkForChanges(dc); [](start = 8, length = 35)

Should checking for changes be built into the dialog context itself?  It's the thing that is orchestrating switching between dialogs isn't it?

libraries/botbuilder-dialogs/src/dialogEvents.ts
```diff
@@ -11,5 +11,6 @@ export class DialogEvents {
     static readonly repromptDialog = "repromptDialog";
     static readonly cancelDialog = "cancelDialog";
     static readonly activityReceived = "activityReceived";
+    static readonly versionChanged = "dialogChanged";
```
tomlm, line 14: >dialogChanged [](start = 38, length = 13)

name doesn't match variable name?

libraries/botbuilder-dialogs/src/waterfallDialog.ts
```diff
@@ -88,6 +88,12 @@ export class WaterfallDialog<O extends object = {}> extends Dialog<O> {
         }
     }
 
+    public get version(): string {
+        // Simply return the id + number of steps to help detect when new steps have
+        // been added to a given waterfall.
+        return `${this.id}:${this.steps.length}`;
+    }
```
tomlm, line 95: I wonder if this should be a method.  

This requires that the langauge supports virtual properties.  For example, WaterfallDialog derives from Dialog and so needs to "override" the base version because it's a property.

libraries/botbuilder-dialogs-adaptive/src/adaptiveDialog.ts
```diff
@@ -192,6 +194,8 @@ export class AdaptiveDialog<O extends object = {}> extends DialogContainer<O> {
     public async continueDialog(dc: DialogContext): Promise<DialogTurnResult> {
         this.onPushScopedServices(dc.context);
         try {
+            await this.checkForVersionChange(dc);
```
tomlm, line 197: >  await this.checkForVersionChange(dc); [](start = 10, length = 39)

I don't think this works. 
Doesn't every component have to expose version information?
Doesn't AdaptiveDialog version have to capture all of the stuff? 
```
            StringBuilder sb = new StringBuilder();
            sb.Append(base.GetVersion());  // this is the .Dialogs
            sb.Append(this.Id ?? string.Empty);
            sb.Append(this.AutoEndDialog);
            sb.Append(this.Recognizer.GetVersion());
            sb.Append(this.Selector.GetVersion());
            sb.Append(this.Generator.GetVersion());
            sb.Append(this.Schema.GetVersion());
            foreach (var trigger in Triggers)
            {
                sb.Append(trigger.GetVersion());  // Trigger itself is made up of conditions, and all of the nested ActionScopes
            }
```

libraries/botbuilder-dialogs/src/dialogContainer.ts
```diff
@@ -29,4 +30,29 @@ export abstract class DialogContainer<O extends object = {}> extends Dialog<O> {
...
+     * Checks to see if a containers child dialogs have changed since the current dialog instance
+     * was started.
+     * 
+     * @remarks
+     * This should be called at the start of `beginDialog()`, `continueDialog()`, and `resumeDialog()`.
+     * @param dc Current dialog context.
+     */
+    protected async checkForVersionChange(dc: DialogContext): Promise<void> {
+        const current = dc.activeDialog.version;
+        dc.activeDialog.version = this.dialogs.getVersion();
```
tomlm, line 44: > this.dialogs.getVersion(); [](start = 33, length = 27)

Instead of this call put this into virtual GetInternalVersion()

PR #2081, opened 2020-04-22
Add ConnectorClientOptions to BotFrameworkAdapterSettings 

### LINKED ISSUE #2013: Add ConnectorClientOptions to BotFrameworkAdapterSettings
The ability to set `ServiceClientOptions` when ConnectorClients are created

Hacking the `createConnectorClientInternal` method prototype on `BotFrameworkAdapter`, and accessing the ConnectorClient's properties directly:
```js
BotFrameworkAdapter.prototype.createConnectorClientInternal = function (serviceUrl, credentials) {
    // Set noRetryPolicy: true so ServiceClient code does not add one.
    // This code adds the default retry policies, with a modified exponentialRetryPolicy
    const client = new ConnectorClient(credentials, { baseUri: serviceUrl, noRetryPolicy: true });
    
    //retryCount, retryInterval, minRetryInterval, maxRetryInterval, requestTimeout
    client._requestPolicyFactories.push(retryWithTimeoutPolicy(3, 1000 * 2, 1000 * 1, 1000 * 4, 1000 * 5));
    client._requestPolicyFactories.push(systemErrorRetryPolicy());
    client._requestPolicyFactories.push(throttlingRetryPolicy());

    return client;
};
```

[enhancement]

EricDahlvang:
Fixes #2013


libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -8,9 +8,10 @@
 
 import { STATUS_CODES } from 'http';
 import { arch, release, type } from 'os';
+import { HttpClient } from '@azure/ms-rest-js';
```
stevengum, line 11: Is this the interface or an abstract class?

EricDahlvang: It is just the interface. I wondered if i should leave it out.?

stevengum: If it’s only used in the private method as a parameter I would leave it out, change the type to any and perhaps leave a comment.

EricDahlvang: removed

PR #2137, opened 2020-04-29
Add HealthCheck Invoke handling

### LINKED ISSUE #1640: Add OOTB ping method to bots (JS)
See the [rollup issue](https://github.com/microsoft/botframework-sdk/issues/5660) for details.

johnataylor:
Fixes #1640

### DIFF
--- a/libraries/botbuilder-core/src/activityHandler.ts
+++ b/libraries/botbuilder-core/src/activityHandler.ts
@@ -3,7 +3,7 @@
  * Licensed under the MIT License.
  */
 
-import { MessageReaction } from 'botframework-schema';
+import { MessageReaction, HealthCheckResponse } from 'botframework-schema';
 import { ActivityHandlerBase } from './activityHandlerBase';
 import { InvokeResponse } from './invokeResponse';
 import { verifyStateOperationName, tokenExchangeOperationName, tokenResponseEventName } from './signInConstants';
@@ -414,11 +414,17 @@ export class ActivityHandler extends ActivityHandlerBase {
     */
     protected async onInvokeActivity(context: TurnContext): Promise<InvokeResponse> {
         try {
-            if (context.activity.name && (context.activity.name === verifyStateOperationName || context.activity.name === tokenExchangeOperationName)) {
-                await this.onSignInInvoke(context);
-                return { status: StatusCodes.OK };
+            switch (context.activity.name) {
+                case verifyStateOperationName:
+                case tokenExchangeOperationName: {
+                    await this.onSignInInvoke(context);
+                    return { status: StatusCodes.OK };
+                }
+                case "healthCheck":
+                    return await ActivityHandler.createInvokeResponse(await this.onHealthCheckAsync(context));
+                default:
+                    throw new Error('NotImplemented');
             }
-            throw new Error('NotImplemented');
         }
         catch (err) {
             if (err.message === 'NotImplemented') {
@@ -441,6 +447,23 @@ export class ActivityHandler extends ActivityHandlerBase {
     */
     protected async onSignInInvoke(context: TurnContext): Promise<void> {
         throw new Error('NotImplemented');
+    }
+
+    /*
+     * Handle _healthCheck invoke activity type_.
+     * 
+     * @param context The context object for the current turn.
+     * 
+     * @remarks
+     * Overwrite this method to customize or extended the built in healthCheck behavior.
+    */
+    protected async onHealthCheckAsync(context: TurnContext): Promise<HealthCheckResponse> {
+        var adapter = <any>context.adapter;
+        if (adapter.hasOwnProperty('healthCheck') && typeof adapter.healthCheck === 'function') {
+            return adapter.healthCheck(context);
+        } else {
+            return { healthResults: { success: true, messages: [ 'Health check succeeded.' ] } };
+        }
     }
 
     /**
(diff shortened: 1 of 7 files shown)


libraries/botbuilder/src/teamsActivityHandler.ts
```diff
@@ -44,43 +44,43 @@ export class TeamsActivityHandler extends ActivityHandler {
             } else {
                 switch (context.activity.name) {
                     case 'fileConsent/invoke':
-                        return TeamsActivityHandler.createInvokeResponse(await this.handleTeamsFileConsent(context, context.activity.value));
+                        return ActivityHandler.createInvokeResponse(await this.handleTeamsFileConsent(context, context.activity.value));
```
cleemullins, line 47: I don't quite understand these changes. Is the old code a bug?

Stevenic: He just moved it to the base class because he needs it down there now

---
In reply to: [417531893](https://github.com/microsoft/botbuilder-js/pull/2137#discussion_r417531893) [](ancestors = 417531893)

libraries/botbuilder-core/src/activityHandler.ts
```diff
@@ -414,11 +414,17 @@ export class ActivityHandler extends ActivityHandlerBase {
...
-            if (context.activity.name && (context.activity.name === verifyStateOperationName || context.activity.name === tokenExchangeOperationName)) {
-                await this.onSignInInvoke(context);
-                return { status: StatusCodes.OK };
+            switch (context.activity.name) {
+                case verifyStateOperationName:
+                case tokenExchangeOperationName: {
+                    await this.onSignInInvoke(context);
+                    return { status: StatusCodes.OK };
+                }
+                case "healthCheck":
```
Stevenic, line 423: >healthCheck [](start = 22, length = 11)

Should this be a constant as well?

libraries/botbuilder/src/teamsActivityHandler.ts
stevengum: createInvokeResponse was moved to the ActivityHandler in this PR, but @johnataylor should it have been moved to ActivityHandlerBase?

johnataylor: no i moved a helper function from TeamsActivityHandler up into the more generic ActivityHandler - it used to be Invokes (and therefore this helper) were all Teams specific.

libraries/botbuilder-core/src/activityHandler.ts
```diff
@@ -443,6 +449,23 @@ export class ActivityHandler extends ActivityHandlerBase {
...
+     * Handle _healthCheck invoke activity type_.
+     * 
+     * @param context The context object for the current turn.
+     * 
+     * @remarks
+     * Overwrite this method to customize or extended the built in healthCheck behavior.
+    */
+    protected async onHealthCheck(context: TurnContext): Promise<HealthCheckResponse> {
+        const  adapter = <any>context.adapter;
+        if (adapter.healthCheck && typeof adapter.healthCheck === 'function') {
```
Stevenic, line 462: >adapter.healthCheck && [](start = 12, length = 22)

you can just do the "typeof" check... You don't need the first check

libraries/botbuilder-core/src/activityHandler.ts
johnataylor: we are missing constants for most of the invokes - for example all the Teams invokes - we have them for some of the security stuff - the thought was we would just do them all together. Currently this is also pretty internal and not primary application programming model so we weren't super concerned

stevengum: This point was brought up in C# as well: https://github.com/microsoft/botbuilder-dotnet/pull/3796/files#r414133232

This could be put into a constant, but it's expected that it's mainly used by our libraries, not the developer's code.

Cleanup for internal (in this case, dispatching) constants could be done in another milestone.

libraries/botbuilder-core/src/activityHandler.ts
johnataylor: OK I'll update that. You guys are cool with this runtime type identification - that's good! you see why I had to do it this way - because of the botbuilder/botbuilder-core split.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -1357,6 +1357,19 @@ export class BotFrameworkAdapter extends BotAdapter implements ExtendedUserToken
...
+        const healthResults = <HealthResults>{
+            success: true,
+            "user-agent": USER_AGENT,
+            messages: [ 'Health check succeeded.' ]
+        };
+        if (this.credentials.appId) {
+            const token = await this.credentials.getToken();
+            healthResults.authorization = `Bearer ${ token }`;
+        }
+        return { healthResults: healthResults };
```
stevengum, line 1370:
```suggestion
if (!(await this.credentialProvider.isAuthenticationDisabled())) {
            const credentials = context.turnState.get(ConnectorClientCacheKey).credentials ||  this.credentials;
            const token = await this.credentials.getToken();
            healthResults.authorization = `Bearer ${ token }`;
        }
        return { healthResults: healthResults };
```

libraries/botbuilder/src/teamsActivityHandler.ts
johnataylor: we don't have much logic in ActivityHandlerBase at all - we use it in ActivityHandler (well now we do) so I just moved it that far.

libraries/botbuilder/src/botFrameworkAdapter.ts
johnataylor: OK updated now. (And you almost got it right - I assume you intended to use the symbol for the cache key and the local credentials on line 1373 😃)

Thanks for the review - this is better - and actually slightly closer to the C#.

PR #2143, opened 2020-04-30
Add choice set

zidaneymar:
Fixes #

## Description

add choiceSet
## Specific Changes

Port choice set implementation from C#


zidaneymar:
@chon219 hi zichuan, I just changed the expression type for the choices field, I have run the test and everything is ok, do we have any sample bots to test those changes?

libraries/botbuilder-dialogs-adaptive-tests/src/choiceSet.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        sample.push('test3')
+
+        const json = JSON.stringify(sample)
+        const value = new ChoiceSet(JSON.parse(json))
+
+        assert.equal('test1', value[0].value);
+        assert.equal('test2', value[1].value);
+        assert.equal('test3', value[2].value);
+    });
+});
```
chon219, line 129: We may need to move these unit tests to botbuilder-dialogs-adaptive package.

libraries/botbuilder-dialogs-adaptive/package.json
```diff
@@ -45,7 +45,7 @@
     "ts-node": "^4.1.0"
   },
   "scripts": {
-    "test": "tsc && tsc ./tests/expressionProperty.test.ts && nyc mocha tests/",
+    "test": "tsc && tsc ./tests/expressionProperty.test.ts && tsc ./tests/choiceSet.test.ts && nyc mocha tests/",
```
chon219, line 48: how about `tsc ./tests/*.ts` ?

Issue #1991, opened 2020-04-03
Skills + OAuth: Differentiate between regular and skill mode when receiving tokens (JS)


carlosscastro:
When doing Oauth, and a Skill gets an activity, the skill thinks it is in regular mode and not skill mode, so it uses the wrong connector client.

the fix is to ahve the Token Service add an additional claim and have the SDK act accordingly and know in what mode it is based on the claim.

cleemullins:
Per @carlosscastro ETA is Monday 4/27 or 4/28, as the C# will land first.

carlosscastro:
@stevengum  and I reviewed C# together and he kindly agreed to do the js side. Thanks :)

PR #2144, opened 2020-04-30
Skills use correct ConnectorClient in OAuthPrompt, add ConnectorClientBuilder

stevengum:
Fixes https://github.com/microsoft/botbuilder-js/issues/1991

## Specific Changes
  - Create `ConnectorClientBuilder` in botbuilder package with property `createConnectorClientWithIdentity`
      - `createConnectorClient` is a synchronous method on the BotFrameworkAdapter, hence the change in name
      - `createConnectorClientWithIdentity` is a property, not a function so that we can deprecate it (and make it optional) if we want to continue using the `ConnectorClientBuilder` interface
  - Port OAuthPrompt changes (https://github.com/microsoft/botbuilder-dotnet/pull/3822) to differentiate between regular and skill mode when receiving tokens
      - The Prompt casts `context.adapter` to `any` instead of `ConnectorClientBuilder` for browser compatibility; but does a property check to see if the adapter contains a `createConnectorClientWithIdentity` function

### DIFF
--- a/libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts
+++ b/libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts
@@ -9,7 +9,7 @@ import { Activity, ActivityTypes, Attachment, CoreAppCredentials, BotAdapter, Ca
 import { Dialog, DialogTurnResult } from '../dialog';
 import { DialogContext } from '../dialogContext';
 import { PromptOptions, PromptRecognizerResult,  PromptValidator } from './prompt';
-import { isSkillClaim } from './skillsHelpers';
+import { isSkillClaim, getAppIdFromClaims } from './skillsHelpers';
 
 /**
  * Response body returned for a token exchange invoke activity.
@@ -124,6 +124,7 @@ export interface OAuthPromptSettings {
  * ```
  */
 export class OAuthPrompt extends Dialog {
+    private readonly PersistedCaller: string = 'botbuilder-dialogs.caller';
     /**
      * Creates a new OAuthPrompt instance.
      * @param dialogId Unique ID of the dialog within its parent `DialogSet` or `ComponentDialog`.
@@ -150,6 +151,7 @@ export class OAuthPrompt extends Dialog {
         state.state = {};
         state.options = o;
         state.expires = new Date().getTime() + timeout;
+        state[this.PersistedCaller] = OAuthPrompt.createCallerInfo(dc.context);
 
         // Attempt to get the users token
         const output: TokenResponse = await this.getUserToken(dc.context);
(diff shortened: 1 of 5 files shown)


libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts
```diff
@@ -320,10 +322,33 @@ export class OAuthPrompt extends Dialog {
...
             token = context.activity.value as TokenResponse;
+
+            // Fix-up the DialogContext's state context if this was received from a skill host caller.
+            const state: CallerInfo = dc.activeDialog.state[this.PersistedCaller];
+            if (state !== null) {
+                // Set the ServiceUrl to the skill host's Url
+                dc.context.activity.serviceUrl = state.callerServiceUrl;
+
+                // Recreate a ConnectorClient and set it in TurnState so replies use the correct one
+                if (!(typeof (context.adapter as any).createConnectorClientWithIdentity === 'function')) {
```
johnataylor, line 338: so we're now doing this in  couple of places - suggests we have something missing in the core abstractions!

libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts
```diff
@@ -124,6 +124,7 @@ export interface OAuthPromptSettings {
  * ```
  */
 export class OAuthPrompt extends Dialog {
+    private readonly PersistedCaller: string = 'botbuilder-dialogs.caller';
```
carlosscastro, line 127: This made me remember that this didn't get done in c#, as somebody merged my PR overnight. Don't think it's in python either. I'll tlak to tracy and do my best to remember to update c#

stevengum: I was just looking at @tracyboehrer's [PR](https://github.com/microsoft/botbuilder-python/pull/1014/files) I'm not sure if we need to do this here... This key may be stored one level higher than where developers interact.

stevengum: @stevenic or @johnataylor could you chime in?

libraries/botbuilder-dialogs/src/prompts/oauthPrompt.ts
stevengum: Agreed, the botbuilder/Microsoft.Bot.Builder libraries are tightly coupled to the Connector code which is not always desirable. The Connector libraries tie the SDKs to HTTP, but the transport protocol isn't guaranteed to be HTTP with browsers, electron, mobile apps (e.g. Xamarin), IoT devices, or notebooks.

One area that could help in this decoupling is the introduction of a `"TransportClient"` interface similar to HttpClient in .netcore.

Stevenic: One simple cleanup would be to use a type guard... [https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)


---
In reply to: [418193184](https://github.com/microsoft/botbuilder-js/pull/2144#discussion_r418193184) [](ancestors = 418193184)

stevengum: stevenic and I synced offline, the type guard would still only result in a casting of any because the required interface is `botframework-connector` dependent and lives in `botbuilder`.

PR #2145, opened 2020-04-30
Add Dialog / PageView telemetry

### LINKED ISSUE #1792: [PORT] Update TelemetryClient to enable tracking of dialog ('page') views
As per C# issue https://github.com/microsoft/botbuilder-dotnet/issues/3441

C# PR -> https://github.com/microsoft/botbuilder-dotnet/pull/3440

This is to support the ability to track user flows within application insights and understand how users move between dialogs within a conversation using the built in user flows / funnels pages within App Insights portal.

[enhancement]

garypretty:
Fixes #1792.  Parity PR for C# https://github.com/microsoft/botbuilder-dotnet/pull/3440


libraries/botbuilder-core/src/extensions.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { BotTelemetryClient, BotPageViewTelemetryClient, Severity } from './botTelemetryClient';
+
+/**
+ * Some util and extension functions
+ */
+export class Extensions {
```
stevengum, line 14: Let's avoid naming it this if possible 😄, @benbrown, @stevenic could one of you assist Gary with updating his PR?

garypretty: @stevengum I agree with you, though I was trying to be consistent as there is another class for things like this in the adaptive library, so if we change the name here, maybe we should change that too?

garypretty: @stevengum @Stevenic @benbrown If this is just a naming things and the rest of the PR is ok, if any of you get a chance to make the name change (once a suitable name has been landed on) that would be good. I am out for the weekend from now. If not and you can suggest an appropriate name, then approve the PR pending that and I will make the change first thing Monday.

stevengum: I have availability; Steve is reviewing it right now, if doesn't have a particular way he wants it then I'll dive in here and push some changes

libraries/botbuilder-core/src/extensions.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    public static TelemetryTrackDialogView(telemetryClient: BotTelemetryClient, dialogName: string, properties?: {[key: string]: any}, metrics?: {[key: string]: number }): void
```
Stevenic, line 16: >TelemetryTrackDialogView [](start = 18, length = 24)

Case is wrong. Should be camelCased.

libraries/botbuilder-core/src/extensions.ts
Stevenic: It seems like these could just be functions exported from botTelemetryClient.ts. They don't technically need to be exported as part of the class... 

---
In reply to: [419700069](https://github.com/microsoft/botbuilder-js/pull/2145#discussion_r419700069) [](ancestors = 419700069)

PR #2149, opened 2020-05-01
Added LuisAdaptiveRecognizer

### LINKED ISSUE #1637: [PORT] External Entities & Dynamic lists
https://github.com/microsoft/botbuilder-dotnet/pull/3300

Fix for #3156 which now enables external entity recognizers and dynamic lists for declarative LUIS recognizers.
*** Breaking change for declarative LUIS Recognizers ***
They now use expressions so to bind you need to do "=luis.settings.endpointKey" rather than "{luis.settings.endpointKey}".  There is also a PR to update bf-cli.

Dropped $role of lg and memorypath since not used.
Changed it so $role of expression would add string type.
Updated TestBot to include settings.
Moved all of the PropertyExpression stuff to its own library.  Once it no longer depends on .lg we can move it into the actual expression library.

# Changed projects
* Microsoft.Bot.Expressions.Properties
* Microsoft.Bot.Builder.AI.LUIS.Tests
* Microsoft.Bot.Builder.Ai.LUISV3.tests
* Microsoft.Bot.Builder.LanguageGeneration.Tests

[R8,adaptive]

Stevenic:
Fixes #1637

## Description
Added missing LuisAdaptiveRecognizer class

### DIFF
--- a/libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
+++ b/libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
@@ -59,8 +59,11 @@ export class LuisAdaptiveRecognizer extends Configurable implements Recognizer {
 
     public async recognize(dialogContext: DialogContext, activity: Activity) {
         // Validate passed in activity matches turn activity 
-        const context = dialogContext.context
-        if (activity && (!context.activity == null || context.activity.type != ActivityTypes.Message || context.activity.text != activity.text)) {
+        const context = dialogContext.context;
+        const utteranceMatches: boolean = !activity || 
+            (context.activity.type === activity.type &&  context.activity.text === activity.text);
+        
+        if (!utteranceMatches) {
             throw new Error(`TurnContext is different than text`);
         }
 
--- a/libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
+++ b/libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
@@ -1,0 +1,94 @@
+/**
+ * @module botbuilder-dialogs-adaptive
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+import { Recognizer } from '../recognizers';
+import { StringExpression, ArrayExpression } from 'adaptive-expressions';
+import { LuisPredictionOptions, LuisRecognizerOptionsV3, LuisRecognizer, LuisApplication } from 'botbuilder-ai';
+import { Configurable, DialogContext } from 'botbuilder-dialogs';
+import { BotTelemetryClient, Activity, ActivityTypes } from 'botbuilder-core';
+
+export class LuisAdaptiveRecognizer extends Configurable implements Recognizer {
+    /**
+     * Recognizers unique ID.
+     */
+    public id: string;
+
+    /**
+     * LUIS application ID.
+     */
+    public applicationId: StringExpression;
+
+    /**
+     * LUIS dynamic list.
+     */
+    public dynamicLists: ArrayExpression<any>;
+
+    /**
+     * LUIS endpoint to query.
+     * @remarks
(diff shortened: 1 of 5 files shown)


libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
```diff
@@ -0,0 +1,94 @@
...
+     * Key used to talk to a LUIS endpoint.
+     */
+    public endpointKey: StringExpression;
+
+    /**
+     * External entity recognizer.
+     * @remarks
+     * This recognizer is run before calling LUIS and the results are passed to LUIS.
+     */
+    // public externalEntityRecognizer: Recognizer;
```
stevengum, line 47: Should this be uncommented?

libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
```diff
@@ -0,0 +1,94 @@
...
+     * LUIS dynamic list.
+     */
+    public dynamicLists: ArrayExpression<any>;
+
+    /**
+     * LUIS endpoint to query.
+     * @remarks
+     * For example: "https://westus.api.cognitive.microsoft.com"
+     */
+    public endpoint: StringExpression;
```
stevengum, line 35: Is this the completed endpoint with the AppId? Or is this the host?

What value is [`ExpressionProperty.prototype.getValue`](https://github.com/microsoft/botbuilder-js/blob/1a2a64a7a0fa2cb9da65f9dfdf88d22fa78a8000/libraries/adaptive-expressions/src/expressionProperties/expressionProperty.ts#L57) supposed to return?

libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Telemetry client.
+     */
+    public telemetryClient: BotTelemetryClient;
+
+    public async recognize(dialogContext: DialogContext, activity: Activity) {
+        // Validate passed in activity matches turn activity 
+        const context = dialogContext.context
+        if (activity && (!context.activity == null || context.activity.type != ActivityTypes.Message || context.activity.text != activity.text)) {
```
stevengum, line 63: `!context.activity == null` should never evaluate to true.

I've refactored the code a little in an attempt to improve readability and used strict non-equality checks:

```suggestion
        const context = dialogContext.context;
        const utteranceMatches: boolean = (context.activity !== undefined && context.activity !== null)
            || context.activity.type !== ActivityTypes.Message
            || context.activity.text != activity.text;
        
        if (activity && utteranceMatches) {
```

I left `context.activity.text != activity.text`, but maybe that should be changed as well?

libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
Stevenic: No... We can't support it right now as it requires changes to the base luis recognizer.

libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
Stevenic: I copied it verbatim from C# so I believe its just endpoint URL as in the comments.

libraries/botbuilder-dialogs-adaptive/src/luis/luisAdaptiveRecognizer.ts
Stevenic: Good catch... I forgot to remove the "= null" part

Stevenic: Your version had issues as well but the changed version should work

---
In reply to: [418658435](https://github.com/microsoft/botbuilder-js/pull/2149#discussion_r418658435) [](ancestors = 418658435)

PR #2155, opened 2020-05-01
Add tenantId and userRole to TeamsChannelAccount

EricDahlvang:
Fixes https://github.com/microsoft/botbuilder-js/issues/2157


stevengum:
I'm looking into the build issues.

stevengum:
#2159 should address the build issue

PR #2159, opened 2020-05-01
Fix CI and align versions of moment and moment-timezone

stevengum:
## Description
Fix CI and align versions of moment and moment-timezone

## Specific Changes
  - Align moment and moment-timezone versions
  - Change imports for moment, moment-timezone and LRUCache
  - Added esModuleInterop flag to adaptive-expressions tsconfig.json
  - Removed `"strict": false` from adaptive-expressions tsconfig.json
  - Removed disabling of eslint rule in `expressionFunctions.ts`

## Testing
Ran `lerna bootstrap` to test the ability to build adaptive-expressions without relying on hoisted dependencies.

### DIFF
--- a/libraries/adaptive-expressions/package.json
+++ b/libraries/adaptive-expressions/package.json
@@ -29,7 +29,7 @@
     "jspath": "^0.4.0",
     "lodash": "^4.17.15",
     "lru-cache": "^5.1.1",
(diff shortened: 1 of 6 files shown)


stevengum:
RCA was a breaking change in `2.25.0` of moment, but the moment team released a quick fix in `2.25.1`

PR #2160, opened 2020-05-01
Stevenic/adaptive fixes

Stevenic:
## Description
- Added missing LG converter.
- Fixed a case conversion issues.
- Worked around a serialization issue for cards coming from LG.

### DIFF
--- a/libraries/botbuilder-dialogs-adaptive/src/generators/languageGeneratorManager.ts
+++ b/libraries/botbuilder-dialogs-adaptive/src/generators/languageGeneratorManager.ts
@@ -42,12 +42,12 @@ export class LanguageGeneratorManager {
 
     public static resourceExplorerResolver(locale: string, resourceMapping: Map<string, IResource[]>): ImportResolverDelegate {
         return  (source: string, id: string): {content: string; id: string} => {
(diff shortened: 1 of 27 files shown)


libraries/botbuilder-dialogs-adaptive/src/generators/languageGeneratorManager.ts
```diff
@@ -34,7 +34,7 @@ export class LanguageGeneratorManager {
         const resources = await this._resourceExporer.getResources('lg');
         for (const resource of resources) {
             const generator = await this.getTemplateEngineLanguageGenerator(resource);
-            this.languageGenerator.set(resource.id().toLocaleLowerCase(), generator);
+            this.languageGenerator.set(resource.id(), generator);
```
stevengum, line 37: Should we still call `.toLowerCase()`?

Stevenic: no... thats the bug


---
In reply to: [418757742](https://github.com/microsoft/botbuilder-js/pull/2160#discussion_r418757742) [](ancestors = 418757742)

PR #2161, opened 2020-05-02
Update SkillDialog to handle all activity types (e.g. set DeliveryMode for Invokes)

stevengum:
Fixes https://github.com/microsoft/botbuilder-js/issues/2139

## Specific Changes
  - Add `StatusCodes.MULTIPLE_CHOICES = 300`
  - Add protected method `SkillDialog.onValidateActivity()`
  - `SkillDialog` now forwards all `ActivityTypes` to the Skill.
  - Add new C# SkillDialog tests and refactor old tests
  - Add missing test for TurnContext when deliveryMode === 'expectReplies'
  - Fix error message in `BotFrameworkAdapter.createConnectorClientWithIdentity()`
  - Add tests for `BotFrameworkAdapter.createConnectorClientWithIdentity()`
  - Add test for `BotFrameworkAdapter.processActivity()` when deliveryMode === 'expectReplies', type is `Invoke` and reply activities have type `invokeResponse`

### DIFF
--- a/libraries/botbuilder-dialogs/src/skillDialog.ts
+++ b/libraries/botbuilder-dialogs/src/skillDialog.ts
@@ -60,7 +60,7 @@ export class SkillDialog extends Dialog {
     }
 
     public async beginDialog(dc: DialogContext, options?: any): Promise<DialogTurnResult> {
-        const dialogArgs = SkillDialog.validateBeginDialogArgs(options);
+        const dialogArgs = this.validateBeginDialogArgs(options);
 
         await dc.context.sendTraceActivity(`${ this.id }.beginDialog()`, undefined, undefined, `Using activity of type: ${ dialogArgs.activity.type }`);
 
@@ -70,7 +70,7 @@ export class SkillDialog extends Dialog {
         // Apply conversation reference and common properties from incoming activity before sending.
         const skillActivity = TurnContext.applyConversationReference(clonedActivity, TurnContext.getConversationReference(dc.context.activity), true) as Activity;
 
-        // Store the deliveryMode of the first forwarded activity
+        // Store delivery mode and connection name in dialog state for later use.
         dc.activeDialog.state[this.DeliveryModeStateKey] = dialogArgs.activity.deliveryMode;
         dc.activeDialog.state[this.SsoConnectionNameKey] = dialogArgs.connectionName;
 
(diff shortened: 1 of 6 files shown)


libraries/botbuilder-dialogs/src/skillDialog.ts
```diff
@@ -162,35 +177,18 @@ export class SkillDialog extends Dialog {
...
-        // Create a conversationId to interact with the skill and send the activity
-        let skillConversationId: string;
-        try {
-            skillConversationId = await this.dialogOptions.conversationIdFactory.createSkillConversationIdWithOptions(conversationIdFactoryOptions);
-        } catch (err) {
-            if (err.message !== 'Not Implemented') throw err;
-            // If the SkillConversationIdFactoryBase implementation doesn't support createSkillConversationIdWithOptions(),
-            // use createSkillConversationId() instead.
-            skillConversationId = await this.dialogOptions.conversationIdFactory.createSkillConversationId(TurnContext.getConversationReference(activity) as ConversationReference);
+        if (activity.type === ActivityTypes.Invoke) {
```
johnataylor, line 184: having to switch expectReplies on for an invoke seems a little odd.

libraries/botbuilder/src/botFrameworkAdapter.ts
```diff
@@ -843,8 +843,18 @@ export class BotFrameworkAdapter extends BotAdapter implements ConnectorClientBu
             context.turnState.set(BotCallbackHandlerKey, logic);
             await this.runMiddleware(context, logic);
 
-            // Retrieve cached invoke response.
-            if (request.type === ActivityTypes.Invoke) {
```
johnataylor, line 847: and this change presumably for consistency with the other place this condition occurs...

stevengum: Yes, these changes are ported from C# with exactly the same spirit (and tests) in mind. That said, I will add some more complex unit tests tomorrow to cover the interactions between the BFAdapter and ActivityHandler, and the two with SkillDialog.

libraries/botbuilder-dialogs/src/skillDialog.ts
gabog: if the dialog is interacting with a skill that only works with invokes, this makes sure that EoC replies are buffered with the invokeResponse and the dialog ends when an EoC comes back.

PR #2201, opened 2020-05-07
Adaptive dialogs bug fix

chon219:
Fixes: #2210
Fixes: #2209

## Description

We found some minor bugs in adaptive dialogs and we need fix them in this PR.

## Specific Changes

  - Registered HttpRequest as well as related converters
  - Added missing onComputeId() for OAuthInput
  - Avoid possible null pointer exceptions in Inputs
  - Avoid possible null pointer exceptions in RegexRecognizer


cleemullins:
@chon219, This is an R10 fix, and with R10 we're starting a new policy. PRs need to have an issue. Can you please create an issue, tag & triage it appropiatly, and then tie it to this PR?

cleemullins:
WE'll be enabling the github enforcement of this soon, but we should start now.

chon219:
> @chon219, This is an R10 fix, and with R10 we're starting a new policy. PRs need to have an issue. Can you please create an issue, tag & triage it appropiatly, and then tie it to this PR? 

OK, I'll create related issues later. Thanks for reminding.

libraries/botbuilder-dialogs-adaptive/src/input/attachmentInput.ts
```diff
@@ -20,7 +20,7 @@ export class AttachmentInput extends InputDialog {
     public outputFormat: EnumExpression<AttachmentOutputFormat> = new EnumExpression<AttachmentOutputFormat>(AttachmentOutputFormat.first);
 
     protected onComputeId(): string {
-        return `AttachmentInput[${ this.prompt.toString() }]`;
+        return `AttachmentInput[${ this.prompt && this.prompt.toString() }]`;
```
stevengum, line 23: Good catch

libraries/botbuilder-dialogs-adaptive/src/input/oauthInput.ts
```diff
@@ -190,6 +190,10 @@ export class OAuthInput extends InputDialog {
         return adapter.signOutUser(dc.context, this.connectionName.getValue(dc.state));
     }
 
+    protected onComputeId(): string {
+        return `OauthInput[${ this.prompt && this.prompt.toString() }]`;
```
stevengum, line 194: This should be OAuthInput to reflect the class name

libraries/botbuilder-dialogs-adaptive/src/recognizers/regexRecognizer.ts
```diff
@@ -71,13 +71,15 @@ export class RegexRecognizer implements Recognizer {
...
-                            entity['start'] = match.index;
-                            entity['end'] = match.index + text.length;
-                            entityPool.push(entity);
+                            if (text) {
+                                const entity: Entity = {
+                                    type: name,
+                                };
+                                entity['text'] = text;
+                                entity['start'] = match.index;
+                                entity['end'] = match.index + text.length;
```
stevengum, line 80: This smells a little, but it's the same code as in C#. What's an example of the `"name"` that's used as the value for `Entity.type`?

We're using the Entity as defined in the [protocol spec](https://github.com/microsoft/botframework-obi/blob/master/protocols/botframework-activity/botframework-activity.md#Entity); but it's just holding numbers and strings here. 

The instancePool down below looks like it's expecting a [ModelResult](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-dialogs/src/choices/modelResult.ts) ([C#](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.Dialogs/Choices/ModelResult.cs)), not a low-level Entity. @stevenic @tomlm or @chrimc62 can you chime in?

chon219: Yes, we have `new Entity(groupName)` in C#, I'm not quite sure about this, but we may need to update this if you found this is an incorrect implementation.

stevengum: Because this object eventually appears as `any` to the developer, it's fine to leave it as is... This is just a bit unfortunate 😞

PR #2217, opened 2020-05-09
Allow loading settings from turn state

### LINKED ISSUE #2216: Allow loading object property into settings
JS SDK load settings from process.env, so we cannot put a complex property (e.g. an object) into settings.

JS SDK should load settings from turn state besides process.env, thus we can put complex settings into it.

**Additional context**

[enhancement]

chon219:
Fixes #2216 

## Description

SettingsMemoryScope should load "settings" from turn state besides loading settings from process.env. We can put extra settings in DialogManager's intialTurnState to be treated as settings.

## Specific Changes

  - Updated SettingsMemoryScope to load settings from turn state
  -
  -

## Testing

Updated corresponding unit test.


libraries/botbuilder-dialogs/src/memory/scopes/settingsMemoryScope.ts
```diff
@@ -10,22 +10,23 @@ import { ScopePath } from '../scopePath';
...
         super(ScopePath.settings, false);
     }
 
     public getMemory(dc: DialogContext): object {
-        // Clone strings from env
-        const settings: object = {};
+        let settings: object = {};
+        if (dc.context.turnState.has(ScopePath.settings)) {
+            settings = dc.context.turnState.get(ScopePath.settings);
+        }
```
stevengum, line 24: This is an additive change in behavior; please add another unit test that demonstrates what happens when the `process.env` key-value pairs collide with the settings from `dc.context.turnState['settings']`.

stevengum: @chon219, is this an attempt to avoid breaking changes for developers getting their settings from `process.env`?

chon219: @stevengum Yes, I'm trying to make it compatible with previous version, is this a requirement?

stevengum: @chon219, yes this is a requirement.

I think it's safe to assume that the additive code in `SettingsMemoryScope.getMemory()` is what we're going to ship R10 with. Can you add a test for the scenario I described in my [earlier comment](https://github.com/microsoft/botbuilder-js/pull/2217/files#r426188801)?

chon219: @stevengum sure, I'll add a test later.

chon219: @stevengum I added a few tests to test if settings from process.env have a higher priority to override settings from turn state.

PR #2223, opened 2020-05-12
Remove node exports from utilities

### LINKED ISSUE #2224: Webpack fails with botframework-streaming due to Node.js "net" dependency
## Versions
4.9.0

## Describe the bug
`botframework-streaming/src/utilities/index.ts` exports `createNodeServer` which requires the Node.js `'net'` module.

Because Webpack supports `require`, it attempts to call `require('net')` and `'net'` is not available, so webpack fails.

## Potential Fix
A quick fix is removing the export of createNodeServer from the index, and then NamedPipeServer importing directly from the `createNodeServer.ts` file:

`import { createNodeServer } from '../utilities/createNodeServer';`

Zerryth:
Fixes #2224 

## Description
loading bf-streaming in browser yields error:
image

`browserWebSocket.ts` imports `utilities`, which included `createNodeServer`

Removed reference to this node export in `utilities`


cleemullins:
I'm not convinced we need to take this in 4.9. Hasn't this been here for... months?

stevengum:
For 4.9.1.

PR #2228, opened 2020-05-12
(AGAINST MASTER) Remove node exports from utilities

### LINKED ISSUE #2224: Webpack fails with botframework-streaming due to Node.js "net" dependency

Zerryth:
Fixes #2224

## Description
Per offline discussion, Chris M. asked that we do the PR against master
See https://github.com/microsoft/botbuilder-js/pull/2223 for PR against 4.9 counterpart


stevengum:
@ckkashyap validated this PR

PR #2246, opened 2020-05-18
Harden Prompts

mdrichardson:
Fixes https://github.com/microsoft/BotBuilder-Samples/issues/1728
Fixes: #2271 

## Description

Some of the prompts break if user uploads an attachment instead of responding with text because of some of the `string` methods that get called on `activity.text` (which doesn't exist if an attachment is sent).

## Specific Changes

Add

```
if (!utterance) {
  return result;
}
```

to:

1. ChoicePrompt
2. DatetimePrompt
3. NumberPrompt

The original issue was only for DatetimePrompt, but I added the other two. Other prompts didn't seem relevant for this.

## Testing

Tested with CoreBot and it fixed original issue.

## Note

I'm currently checking to see if the other SDK languages need something like this.


libraries/botbuilder-dialogs/src/prompts/choicePrompt.ts
```diff
@@ -106,6 +106,9 @@ export class ChoicePrompt extends Prompt<FoundChoice> {
         const result: PromptRecognizerResult<FoundChoice> = { succeeded: false };
         const activity: Activity = context.activity;
         const utterance: string = activity.text;
+        if (utterance == null) {
```
GeoffCoxMSFT, line 109: Consider adding a unit test in choicePrompt.test.js to cover the null utterance case.

GeoffCoxMSFT, line 109: Consider 

`if (!utterance) `

if you want to handle the undefined and zero-length cases as well.

mdrichardson: Done. Also added empty string test for applicable prompts

mdrichardson:
@GeoffCoxMSFT Good suggestions. Made the changes and am requesting re-review. Note: none of the pipeline tests failures are due to this PR.

libraries/botbuilder-dialogs/tests/datetimePrompt.test.js
```diff
@@ -184,4 +184,36 @@ describe('DatetimePrompt', function () {
...
+        // Create a DialogState property, DialogSet and DateTimePrompt.
+        const dialogState = convoState.createProperty('dialogState');
+        const dialogs = new DialogSet(dialogState);
+        dialogs.add(new DateTimePrompt('prompt'));
+
+        await adapter.send('Hello')
+            .assertReply('Enter a date.')
+            .send('')
+            .assertReply('Enter a date.')
+            .send({ type: 'message', text: null })
```
stevengum, line 214: Replace `'message'` with `ActivityTypes.Message`

PR #2251, opened 2020-05-19
Modify TelemetryConstants to use static properties

### LINKED ISSUE #2063: TelemetryConstants should use static properties not instance properties
These class properties should have been static like in [C#](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder/TelemetryConstants.cs). From a cursory glance on the class' references developers shouldn't need to instantiate this class to access these values.

_Originally posted by @stevengum in https://github.com/microsoft/botbuilder-js/pull/2025_

mrestorff:
Fixes #2063

## Description
As noted in #2063, instantiation of TelemetryConstants should not be necessary. Converted TelemetryConstant properties to static ones and adjusted TelemetryLoggerMiddleware accordingly.

## Specific Changes

  - modify TelemetryConstants to use static properties
  - remove instantiation of TelemetryConstants from TelemetryLoggerMiddleware
  - add TSDoc comment to TelemetryConstants, same as C#


libraries/botbuilder-core/src/telemetryConstants.ts
```diff
@@ -1,19 +1,22 @@
...
+    public static readonly fromIdProperty: string = 'fromId';
+    public static readonly fromNameProperty: string = 'fromName';
+    public static readonly localeProperty: string = 'locale';
+    public static readonly recipientIdProperty: string = 'recipientId';
+    public static readonly recipientNameProperty: string = 'recipientName';
+    public static readonly replyActivityIdProperty: string = 'replyActivityId';
+    public static readonly textProperty: string = 'text';
+    public static readonly speakProperty: string = 'speak';
+    public static readonly userIdProperty: string = 'userId';
+    public static readonly attachmentsProperty: string = 'attachments';
```
stevengum, line 21: This class is internal, therefore these are not breaking changes.

PR #2268, opened 2020-05-21
replace atob package to be compatible with web worker

### LINKED ISSUE #2260: package atob does not work in Web worker
package atob depends on `window` to evaluate the result in browser environment.
But in web worker, `window` is undefined.
attach:

image

Danieladu:
Fixes #2260

## Description
package [atob](https://www.npmjs.com/package/atob) does not work correctly in web worker environment, replace it with package [atob-lite](https://www.npmjs.com/package/atob-lite)


boydc2014:
hi, @Stevenic  can you help check this one, we need this to fix for composer's scenario, which move all parser into web worker that don't work with previous "atob" package.

PR #2283, opened 2020-05-28
Removed fallback exception from version change logic.

Stevenic:
## Description
We added a new feature in 4.9 to automatically detect dialog changes. Some customers are complaining about the fallback logic we have to automatically throw an exception if the bot doesn't handle the new event we added.  We've decided to remove the fallback exception and make the feature entirely opt-in.


cleemullins:
The C# PR that matches this is here:
https://github.com/microsoft/botbuilder-dotnet/pull/4000

Issue #2281, opened 2020-05-28
getPastTime and getFutureTime have time-zone problems


beyackle:
## Describe the bug
getPastTime and getFutureTime are applying a time zone correction when they shouldn't, resulting in incorrect results.

## To Reproduce
1. Create a bot that sends a response of "- Now: ${utcNow()} - Yesterday: ${getPastTime(24, 'Hour')} - Tomorrow: ${getFutureTime(24, 'Hour')}"
2. Run the bot and observe that the past and future times are calculated misinterpreting the current time as if it were UTC, while utcNow correctly displays the current time in UTC.

The same behavior happens with (1,'Day') as well.

## Expected behavior

Date math should return the correct responses without misapplying a time zone correction.

## Screenshots
![image](https://user-images.githubusercontent.com/61990921/83178189-0c1e1e00-a0d5-11ea-9e2c-827251cb3d05.png)

[bug]

PR #2292, opened 2020-06-01
fix default timestamp format in Adaptive Expressions

cosmicshuai:
close: #2281
default timestamp format  now is 'YYYY-MM-DDTHH:mm:ss.SSS[Z]', which can produce the correct formatted results for the datetime related functions


libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -48,7 +48,12 @@ export class ExpressionFunctions {
     /**
      * The default date time format string.
      */
-    public static readonly DefaultDateTimeFormat: string = 'YYYY-MM-DDTHH:mm:ss.sssZ';
+    public static readonly DefaultDateTimeFormat: string = 'YYYY-MM-DDTHH:mm:ss.SSS[Z]';
+
+    /**
+     * The default date time format string.
```
Danieladu, line 54: refine the comments

cosmicshuai: fixed

PR #2329, opened 2020-06-10
[PORT]Add EOL function for both browser and Nodejs

### LINKED ISSUE #2288: [PORT] Add EOL builtin function
https://github.com/microsoft/botbuilder-dotnet/pull/3992

close: #3787 
Add `EOL()` builtin function which supplies the line ending for the current OS.

# Changed projects
* AdaptiveExpressions
* AdaptiveExpressions.Tests

Danieladu:
Fixes #2288

## Description
Add EOL function for both browser and Nodejs

- windows platform -> '\r\n'
- macOS/unix platform -> '\n'

- windows platform -> '\r\n'
- macOS/unix -> '\n'

## TODO
Add unit test for the browser environment

### DIFF
--- a/libraries/adaptive-expressions/src/index_browser.ts
+++ b/libraries/adaptive-expressions/src/index_browser.ts
@@ -6,4 +6,4 @@
  * Licensed under the MIT License.
  */
 export * from './index_common';
-export * from './builtinFunctions_node';
+export * from './builtinFunctions_browser';
--- a/libraries/adaptive-expressions/tests/expressionParser.test.js
+++ b/libraries/adaptive-expressions/tests/expressionParser.test.js
@@ -6,6 +6,7 @@ const assert = require('assert');
 const moment = require('moment');
 const bigInt = require('big-integer');
 const { useFakeTimers } = require('sinon');
+const os = require('os');
 
 const one = ['one'];
 const oneTwo = ['one', 'two'];
@@ -230,6 +231,7 @@ const dataSource = [
     ['count(newGuid())', 36],
     ['indexOf(newGuid(), \'-\')', 8],
     ['indexOf(newGuid(), \'-\')', 8],
+    ['EOL()', os.EOL],
     ['indexOf(hello, \'-\')', -1],
     ['indexOf(nullObj, \'-\')', -1],
     ['indexOf(hello, nullObj)', 0],
--- a/libraries/adaptive-expressions/package.json
+++ b/libraries/adaptive-expressions/package.json
@@ -17,6 +17,7 @@
     "url": "https://github.com/Microsoft/botbuilder-js.git"
   },
   "main": "./lib/index.js",
+  "browser": "./lib/index_browser.js",
   "typings": "./lib/index.d.ts",
   "dependencies": {
     "@microsoft/recognizers-text-data-types-timex-expression": "1.1.4",
--- a/libraries/adaptive-expressions/src/builtinFunctions_browser/eol.ts
+++ b/libraries/adaptive-expressions/src/builtinFunctions_browser/eol.ts
@@ -1,0 +1,36 @@
+/**
+ * @module adaptive-expressions
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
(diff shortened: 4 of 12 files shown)


Danieladu:
I have added a new function `EOL`. Consider the discrepancy between Browser and Node environment.
So I set up two entrances for the two environments respectively.  
Appreciate to be given some comments

Danieladu:
> @Danieladu can you describe the way you tested the browser compatible code in this PR?

Currently, I test the browser compatible in my locale browser, I have no idea and experience about how to test it in the CI pipeline. Maybe I need some time to do some investigation.
Mark the pr as WIP.

libraries/adaptive-expressions/src/expressionFunctionsForNode.ts
```diff
@@ -0,0 +1,39 @@
...
+import { ExpressionEvaluator } from './expressionEvaluator';
+import { ExpressionType } from './expressionType';
+import { ExpressionFunctions } from './expressionFunctions';
+import { ReturnType, Expression } from './expression';
+import * as os from 'os';
+
+/**
+ * Inject Node-specific functions into the common builtin-function list.
+ */
+export class ExpressionFunctionsForNode {
```
GeoffCoxMSFT, line 18: Given that the _initialize is executed immediately, the injectedFunctions is only called once, and the _ prefix indicates a private method others shouldn't call - I wonder if this needs to be exported or could be an immediately executed function instead?

libraries/adaptive-expressions/src/expressionFunctionsForNode.ts
```diff
@@ -0,0 +1,39 @@
...
+    public static _initialize(): void {
```
GeoffCoxMSFT, line 19: If you do want this public, then I would avoid the _ prefix.
If it is public, what happens if someone calls it again?

libraries/adaptive-expressions/src/expressionFunctionsForBrowser.ts
```diff
@@ -0,0 +1,44 @@
...
+import { ExpressionEvaluator } from './expressionEvaluator';
+import { ExpressionType } from './expressionType';
+import { ExpressionFunctions } from './expressionFunctions';
+import { ReturnType, Expression } from './expression';
+
+/**
+ * Inject Browser-specific functions into the common builtin-function list.
+ */
+export class ExpressionFunctionsForBrowser {
+    public static _initialize(): void {
```
GeoffCoxMSFT, line 18: Same comments as I left in the expressionFunctionsForNode.ts

libraries/adaptive-expressions/tests/expressionParser.test.js
```diff
@@ -1,10 +1,11 @@
 /* eslint-disable @typescript-eslint/no-use-before-define */
 /* eslint-disable @typescript-eslint/no-var-requires */
-const {Expression, SimpleObjectMemory, ExpressionFunctions, Options} = require('../lib');
+const {Expression, SimpleObjectMemory, ExpressionFunctions, Options} = require('../lib/index');
```
GeoffCoxMSFT, line 3: Why add the index? It will be automatically found for you.

libraries/adaptive-expressions/src/index_browser.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * @module adaptive-expressions
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+export * from './expressionFunctions';
```
GeoffCoxMSFT, line 8: Consider having index_common.ts that exports things shared between browser and node.  Have both index.ts and index_browser.ts export * from index_common.  That way folks don't have to try and keep the common stuff in sync between the two files

libraries/adaptive-expressions/src/expressionFunctionsForNode.ts
Danieladu: That's a good point. I will fix this one.

Danieladu:
Close it temporarily. We will continue finding the best practice to support the browser runtime and the browser test environment in Expression.

PR #2342, opened 2020-06-13
Refactored resource explorer as well as related components

### LINKED ISSUE #1959: [PORT] DCR: Change .dialog file string resource lookup to fallback to non-dialog files.
https://github.com/microsoft/botbuilder-dotnet/pull/3640

#3641 DCR

* Add fallback for string reference binding "foo.xxx" => looks for "foo.xxx.dialog" and if not found attempts to bind to "foo.xxx",

The point of these changes is that it allows new extensions to be added to the resource explorer, and if there are object bindings for those extensions they will be used.  If a file is valid json but NOT a dialog file (Such as .schema files) then they can be used directly without having to have .dialog extension.

For example, this captures the 2 scenarios:
1. **"recognizer":"foo.lu"** => will bind to attempt to bind to "foo.lu.dialog", and if lubuild has run find it. if there is no .dialog it will attempt to load the foo.lu file and appropriately fail.
2. **"schema":"foo.schema"** => will attempt to bind to foo.schema.dialog, and then fall back to foo.schema, which is a valid JSON object to bind and it will load it. (Thus removing the need to have schema file be **foo.schema.dialog**.

Some additional changes:
* Move folderResource.Extensions to resourceExporer.ResourceTypes and make non-static. This allos resource explorer to be configured to pick up additional resource types.
* removed some unused fields in there.
* cleaned up some nonsense I disovered in .csproj files
* added .qna files to pattern list

# Changed projects
* Microsoft.Bot.Builder.AI.QnA
* Microsoft.Bot.Builder.Azure

[DCR,adaptive]

chon219:
Fixes #1959 

## Description

Changed some components from interfaces to abstract class to align with dotnet sdk
Refactored resource explorer to support more resource types besides ".dialog"
Refactored event emitters to expose add/change/remove events on resources
Clean up unused code and unused dependencies in botbuilder-dialogs-declarative

## Specific Changes

  - `Resource`: interface -> abstract class
  - `ResourceProvider`: interface -> abstract class
  - use `chokidar` instead of `node-watch` to expose added/changed/removed events on resources for tracking resource changes more precisely
  - let `LanguageGeneratorManager` use resource explorer to track LG files changes
  - removed unused components like `FileResourceProvider`
  - removed unused resource files

## Testing

Test cases added to test resource changes tracking by resource explorer.


libraries/botbuilder-dialogs-declarative/src/resources/fileResource.ts
```diff
@@ -6,41 +6,38 @@
...
-        const temp = fs.readFileSync(filePath, 'utf-8');
-        return temp;
+        const filePath = this._fullname;
+        const text = fs.readFileSync(filePath, 'utf-8');
+        return text;
     }
 
     public toString(): string {
-        return this.id();
+        return this._id;
```
stevengum, line 41: This change is technically parity with .NET, but why do we still have FileResource.toString() in .NET if this is just a function that does the equivalent of [`Resource.id`](https://github.com/microsoft/botbuilder-js/pull/2342/files#diff-b737d102b45d121bda767b2e389e8074R15)?

chon219: > This change is technically parity with .NET, but why do we still have FileResource.toString() in .NET if this is just a function that does the equivalent of [`Resource.id`](https://github.com/microsoft/botbuilder-js/pull/2342/files#diff-b737d102b45d121bda767b2e389e8074R15)?

I'm not the one who designed this interface, I guess in C# every class can have its own ToString() method to help generate a string content representing itself, which might be useful in some cases, that is why(I guess) we can see there are ToString() methods everywhere. However, in JS we never use these methods to provide string contents for objects, so it is ok to remove them.

libraries/botbuilder-dialogs-declarative/package.json
```diff
@@ -20,29 +20,21 @@
...
+    "@types/jsonpath": "^0.2.0",
     "codelyzer": "^4.1.0",
     "mocha": "^6.2.3",
     "nyc": "^15.1.0",
     "source-map-support": "^0.5.3",
     "ts-node": "^4.1.0"
   },
   "scripts": {
-    "test": "tsc && nyc mocha tests/",
+    "test": "tsc && nyc mocha tests/ --exit",
```
GeoffCoxMSFT, line 37: Intentional?

libraries/botbuilder-dialogs-adaptive/src/generators/languageGeneratorManager.ts
```diff
@@ -23,55 +23,64 @@ export class LanguageGeneratorManager {
     /// <summary>
     /// multi language lg resources. en -> [resourcelist].
     /// </summary>
-    private _multiLanguageResources: Map<string, IResource[]>;
+    private _multiLanguageResources: Map<string, Resource[]>;
 
     public constructor(resourceManager: ResourceExplorer) {
         this._resourceExporer = resourceManager;
-
+        this._resourceExporer.changed = this.resourceExplorerChanged.bind(this);
```
GeoffCoxMSFT, line 30: FYI - If you write the resourceExplorerChanged function as a lamba, then you don't need to bind(this) in order for callbacks to work.

const resourceExplorerChanged = async(event: ResourceChangeEvent, resources: Resource[]): Promise<void> => { ...}

GeoffCoxMSFT:
IMHO - it is fine to have an abstract base class for providing default implementation shared across derived classes.  However, I don't think this has to mean removing the interfaces.  By removing the interface, the other classes become dependent on a specific implementation.  Interfaces do a good job of defining the contract in Typescript and get compiled away.  There aren't the same tradeoffs between ABCs and interfaces like there are in .NET.

libraries/botbuilder-dialogs-declarative/package.json
chon219: Yes, we added some tests to test if resource explorer can monitor folder changes continuously, which may prevent mocha tests exiting as usual. Any better suggestions?

libraries/botbuilder-dialogs-adaptive/src/generators/languageGeneratorManager.ts
chon219: Cool, thanks.

PR #2343, opened 2020-06-14
Updated Dialog Manager to work with skills

### LINKED ISSUE #1970: Update DialogManager to work with Skills (JS)
See the [rollup issue](https://github.com/microsoft/botframework-sdk/issues/5792) for details.

### LINKED ISSUE #2263: [PORT] Set Locale for EoC in DialogManager
https://github.com/microsoft/botbuilder-dotnet/pull/3875

align to #3739

# Changed projects

chon219:
Fixes #1970 
Fixes #2263 

## Description

Updated DialogManager to support skills

## Specific Changes

  - Updated DialogManager to support skills
  - Split onTurn in DialogManager into 2 private methods to process skill/parent requests
  -

## Testing

Added tests for Dialog Manager to cover skills


stevengum:
@chon219, I will take a look at this today. Thanks for jumping on it!

libraries/botbuilder-dialogs/src/dialogHelper.ts
```diff
@@ -95,7 +95,12 @@ export async function runDialog(dialog: Dialog, context: TurnContext, accessor:
  * Helper to determine if we should send an EoC to the parent or not.
  * @param context 
  */
-function sendEoCToParent(context: TurnContext): boolean {
+export function shouldSendEndOfConversationToParent(context: TurnContext, turnResult: DialogTurnResult): boolean {
+    if (!(turnResult.status == DialogTurnStatus.complete || turnResult.status == DialogTurnStatus.cancelled)) {
+        // The dialog is still going, don't return EoC.
+        return false;
+    }
+
```
stevengum, line 103: @chon219, this is a combination that combines [`DialogManager.ShouldSendEndOfConversationToParent()`](https://github.com/microsoft/botbuilder-dotnet/blob/2e2f20b65017305f3697d0d5b398dcd2bbfa6a04/libraries/Microsoft.Bot.Builder.Dialogs/DialogManager.cs#L378-L384) and [`DialogExtensions.SendEoCToParent()`](https://github.com/microsoft/botbuilder-dotnet/blob/2e2f20b65017305f3697d0d5b398dcd2bbfa6a04/libraries/Microsoft.Bot.Builder.Dialogs/DialogExtensions.cs#L100), correct?

I think we should also take this change in C#. What do you think @gabog?

libraries/botbuilder-dialogs/src/dialogHelper.ts
```diff
@@ -113,8 +118,21 @@ function sendEoCToParent(context: TurnContext): boolean {
     return false;
 }
 
+// Helper to send a trace activity with a memory snapshot of the active dialog DC.
+export async function sendStateSnapshotTrace(dc: DialogContext, traceLabel: string): Promise<void> {
```
stevengum, line 122: I don't think we need to export this from dialogHelper since it's only used in DialogManager in C#, can you move it to mirror the C# implementation and can you fix the docstring here?

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
-        // Send trace of memory to emulator
-        let snapshotDc = dc;
-        while (snapshotDc.child) {
-            snapshotDc = snapshotDc.child;
+        return { turnResult: turnResult };
+    }
+
+    private async handleSkillOnTurn(dc: DialogContext): Promise<DialogTurnResult> {
+        // The bot is running as a skill.
+        let turnContext = dc.context;
```
stevengum, line 218:
```suggestion
        const turnContext = dc.context;
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
+
+        // Process remote cancellation.
+        if (turnContext.activity.type == ActivityTypes.EndOfConversation && dc.activeDialog && isFromParentToSkill(turnContext))
+        {
```
stevengum, line 222:
```suggestion
        if (turnContext.activity.type === ActivityTypes.EndOfConversation && dc.activeDialog && isFromParentToSkill(turnContext)) {
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
-            valueType: 'https://www.botframework.com/schemas/botState',
-            value: snapshot,
-            label: 'Bot State'
-        });
 
-        return { turnResult: turnResult };
+        // Handle reprompt
+        // Process a reprompt event sent from the parent.
+        if (turnContext.activity.type == ActivityTypes.Event && turnContext.activity.name == DialogEvents.repromptDialog)
+        {
```
stevengum, line 236:
```suggestion
        if (turnContext.activity.type === ActivityTypes.Event && turnContext.activity.name === DialogEvents.repromptDialog) {
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
+            if (!dc.activeDialog)
+            {
+                return {
+                    status: DialogTurnStatus.empty
+                };
+            }
```
stevengum, line 242:
```suggestion
            if (!dc.activeDialog) {
                return { status: DialogTurnStatus.empty };
            }
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
+
+            await dc.repromptDialog();
+            return {
+                status: DialogTurnStatus.waiting
+            };
```
stevengum, line 247:
```suggestion
            return { status: DialogTurnStatus.waiting };
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
+        }
+
+        // Continue execution
+        // - This will apply any queued up interruptions and execute the current/next step(s).
+        var turnResult = await dc.continueDialog();
+        if (turnResult.status == DialogTurnStatus.empty)
+        {
+            // restart root dialog
+            var startMessageText = `Starting ${ this._rootDialogId }.`;
```
stevengum, line 256:
```suggestion
        const turnResult = await dc.continueDialog();
        if (turnResult.status === DialogTurnStatus.empty) {
            // restart root dialog
            const startMessageText = `Starting ${ this._rootDialogId }.`;
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
+            await turnContext.sendTraceActivity('DialogManager.onTurn()', undefined, undefined, startMessageText);
+            turnResult = await dc.beginDialog(this._rootDialogId);
+        }
+
+        await sendStateSnapshotTrace(dc, 'Skill State');
+
+        if (shouldSendEndOfConversationToParent(turnContext, turnResult))
+        {
+            var endMessageText = `Dialog ${ this._rootDialogId } has **completed**. Sending EndOfConversation.`;
```
stevengum, line 265:
```suggestion
        if (shouldSendEndOfConversationToParent(turnContext, turnResult)) {
            const endMessageText = `Dialog ${ this._rootDialogId } has **completed**. Sending EndOfConversation.`;
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
+
+        return turnResult;
+    }
+
+    private async handleBotOnTurn(dc: DialogContext): Promise<DialogTurnResult> {
+        let turnResult: DialogTurnResult;
+
+        // the bot is running as a root bot. 
+        if (!dc.activeDialog)
+        {
```
stevengum, line 281:
```suggestion
        if (!dc.activeDialog) {
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
+            // start root dialog
+            turnResult = await dc.beginDialog(this._rootDialogId);
+        }
+        else
+        {
```
stevengum, line 286:
```suggestion
        } else {
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
+            // Continue execution
+            // - This will apply any queued up interruptions and execute the current/next step(s).
+            turnResult = await dc.continueDialog();
+
+            if (turnResult.status == DialogTurnStatus.empty)
+            {
```
stevengum, line 292:
```suggestion
            if (turnResult.status === DialogTurnStatus.empty) {
```

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -51,64 +53,74 @@ export interface DialogManagerConfiguration {
 }
 
 export class DialogManager extends Configurable {
-    private dialogSet: DialogSet = new DialogSet();
-    private rootDialogId: string;
-    private dialogStateProperty: string;
+    private _rootDialogId: string;
+    private readonly _dialogStateProperty: string;
```
stevengum, line 57: These changes are unnecessary, we don't _normally_ follow this naming convention in JS... But since _initialTurnState already has the `_`, let's just leave it in so that the class is consistent.

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -51,64 +53,74 @@ export interface DialogManagerConfiguration {
...
+    /**
+     * Bots persisted conversation state.
+     */
+    public conversationState: ConversationState;
+
+    /**
+     * Optional. Bots persisted user state.
+     */
+    public userState?: UserState;
+
```
stevengum, line 76: Small but nice touch... Thanks for moving these.

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -51,64 +53,74 @@ export interface DialogManagerConfiguration {
...
-        this.rootDialogId = dialog.id;
+    public set rootDialog(value: Dialog) {
+        this.dialogs = new DialogSet();
+        if (value) {
+            this._rootDialogId = value.id;
+            this.dialogs.telemetryClient = value.telemetryClient;
+            this.dialogs.add(value);
+        } else {
+            this._rootDialogId = undefined;
+        }
```
stevengum, line 95: This is a bit more complicated than the original logic... Can you add tests for it?

libraries/botbuilder-dialogs/src/dialogHelper.ts
chon219: @stevengum I think so. We have two implementations of this function in c# but it seems that we only need one.

libraries/botbuilder-dialogs/src/dialogHelper.ts
chon219: @stevengum do you mean we need to make it a private method of DialogManager? We have several functions in dialog helper designed to be imported and used in DialogManager, they won't be exported to the outside of package, do we need to merge all of them with DialogManager?

libraries/botbuilder-dialogs/src/dialogManager.ts
chon219: Tests added, this is ported from c# version of `rootDialog`'s set method.

libraries/botbuilder-dialogs/src/dialogHelper.ts
stevengum: This function is "half-public" in the sense that if someone imported this specific file, they would have access to the function, even though the function is not exported from the index.ts. Since the function seems to be only used in DialogManager, it should live there for now as a private method.

I might be missing some references to this function, but in the PR Ctrl+F only returns 3 references besides the declaration; the import and its usage in DialogManager.

chon219: Got it, moved `sendStateSnapshotTrace` into dialog manager.

libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -192,20 +210,93 @@ export class DialogManager extends Configurable {
...
+            // - This will apply any queued up interruptions and execute the current/next step(s).
+            turnResult = await dc.continueDialog();
+
+            if (turnResult.status == DialogTurnStatus.empty) {
+                // restart root dialog
+                turnResult = await dc.beginDialog(this._rootDialogId);
+            }
+        }
+
+        await this.sendStateSnapshotTrace(dc, 'Bot State');
```
stevengum, line 298: Can you add a test for this?

chon219: sure, I'll add a test here.

chon219: state trace test added.

libraries/botbuilder-dialogs-adaptive/src/skillExtensions.ts
```diff
@@ -0,0 +1,39 @@
...
+    _self.initialTurnState.set(skillClientKey, skillClient);
+    return _self;
+};
+
+export const skillConversationIdFactoryKey = Symbol('SkillConversationIdFactory');
+
+/**
+ * Configures the skill conversation id factory to use.
+ */
+DialogManager.prototype.useSkillConverationIdFactory = function(skillConversationIdFactory: SkillConversationIdFactoryBase): DialogManager {
```
stevengum, line 35: Per the sync-up on 7/2/20, this isn't the way to do extension methods in TS/JS as there _isn't_ a good way to port extension methods verbatim. (see microsoft/typescript#9, look for posts from **RyanCavanaugh**)

There are some non-trivial problems that we create by going down this path, problems that would be avoided through either inheritance (a new `"AdaptiveDialogManager"`) or standalone helper functions (example: [teamsActivityHelpers.ts](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder/src/teamsActivityHelpers.ts)).

Additionally, our Skills and DialogManager concepts and implementation exist at a level outside of the Adaptive "namespace" and could be placed directly on the DialogManager. By placing these extension methods in an Adaptive package, we're likely going to confuse developers.

_(And we own the code base, so there will be additional confusion around our methods)_

### Workaround:

As a temporary workaround, @stevenic suggested keeping the keys and instead having developers set these initialTurnState values themselves.

```ts
import { skillConversationIdFactoryKey } from 'botbuilder-dialogs-adaptive'; 
import { DialogManager } from 'botbuilder-dialogs';

const dm = new DialogManager(/* some dialog */);
dm.initialTurnState.set(skillConversationIdFactoryKey, /* some idFactory */);
```

For programmatically registering values onto the DialogManager.initialTurnState, the code here is lightweight enough to implement via the application that uses Adaptive libraries as its runtime.

libraries/botbuilder-dialogs/tests/dialogManager.test.js
```diff
@@ -227,6 +228,50 @@ describe('DialogManager', function() {
         strictEqual(_dmTurnResult.turnResult.status, DialogTurnStatus.empty);
     });
 
+    it('Trace skill state', async () => {
```
stevengum, line 231: Thank you for adding these tests 😄

PR #2369, opened 2020-06-16
Fix renaming of Task Response type

### LINKED ISSUE #2332: Correct Teams typings for TaskModuleResponses
The typings around acceptable values for TaskModuleResponses are incorrect, see below:

image

https://github.com/microsoft/botbuilder-js/pull/1789/files#diff-aabeb3186244709c1c8df5dad0a42a39

We may be missing correct test coverage in this area.

ert485:
fixes #2332 
## Description

`'continue'` should be a valid type for a `TaskModuleResponseBase`->`type`, but the types only allow `'edit'` and `'send'` (both of which fail)

#1789 incorrectly changed `Type2` to `BotMessagePreviewActionType` instead of `BotMessagePreviewType`

## Specific Changes

  - use `BotMessagePreviewType` where `Type2` previously was used


stevengum:
Thanks @ert425!

PR #2386, opened 2020-06-19
[PORT][Expression]Enable list concatenation for function "concat"

### LINKED ISSUE #2373: [PORT] [Expression]Enable list concatenation for function "concat"
https://github.com/microsoft/botbuilder-dotnet/pull/4066

`concat`:
1. If all items are lists, concatenate them with list mode.
`concat([1, 2], [3, 4])` -> `[1, 2, 3, 4]`

2. If there exists any item that is not a list, use string concatenation mode:
`concat('a', 'b', 1, 2)` -> `'ab12'`
`concat('a', [1, 2])` -> `aSystem.Collections.Generic.List 1[System.Object]` // unreadable, best to avoid

3. If null exists, skip it
`concat([1, 2], null)` -> `[1, 2]`
`concat('a', 1, null)` -> `'a1'`

issue link: #4014

# Changed projects
* AdaptiveExpressions
* AdaptiveExpressions.Tests

Danieladu:
Fixes #2373

`concat`:
1. If all items are lists, concatenate them with list mode.
`concat([1, 2], [3, 4])` -> `[1, 2, 3, 4]`

2. If there exists any item that is not a list, use string concatenation mode:
`concat('a', 'b', 1, 2)` -> `'ab12'`
`concat('a', [1, 2])` -> `aSystem.Collections.Generic.List 1[System.Object]` // unreadable, best to avoid

3. If null exists, skip it
`concat([1, 2], null)` -> `[1, 2]`
`concat('a', 1, null)` -> `'a1'`


libraries/adaptive-expressions/src/expressionFunctions.ts
```diff
@@ -1962,6 +1962,20 @@ export class ExpressionFunctions {
         return res;
     }
 
+    private static commonStringify(input: any): string {
+        if (input === null || input === undefined) {
```
chrimc62, line 1966: >(input === null || input === undefined [](start = 11, length = 38)

!input would be clearer and more idiomatic. It is also fine for the empty string case.

Danieladu: but it is not fine for the boolean type `false`

PR #2431, opened 2020-06-25
Have ActivityPrompt.repromptDialog call onPrompt with isRetry set to false

v-kydela:
Fixes https://github.com/microsoft/botbuilder-dotnet/issues/2449

## Description
This fixes an inconsistency between prompt.ts and activityPrompt.ts. In `ActivityPrompt`, resuming the prompt counted as a retry while in `Prompt` it did not. This is because of the `isRetry` parameter of `onPrompt`.

Here is my reasoning to have them both be false rather than both be true. The kind of situation when a prompt would be resumed is when a user types "help" and then a help dialog is added to the stack and then ends. It makes more sense for the prompt to then say "Please enter your age" than to say "I'm sorry I couldn't understand that. Enter a number between 1 and 150."

In Python `is_retry` is already false in both cases so this choice helps with parity.


libraries/botbuilder-dialogs/tests/activityPrompt.test.js
```diff
@@ -162,7 +162,7 @@ describe('ActivityPrompt', function () {
 
             const results = await dc.continueDialog();
             if (results.status === DialogTurnStatus.empty) {
-                await dc.prompt('prompt', 'Please send an activity.');
+                await dc.prompt('prompt', { prompt: 'Please send an activity.', retryPrompt: 'Activity not received.' });
```
stevengum, line 165: Can you update the tests to match @EricDahlvang's changes in .NET?

https://github.com/microsoft/botbuilder-dotnet/pull/4298/files#diff-2f1e96fad160f0d167a84f3c85ba687c

v-kydela: Done

PR #2434, opened 2020-06-26
Add null checks to telemetryLoggerMiddleware

### LINKED ISSUE #2274: telemetryLoggerMiddleware code issue for TS
#### What project is affected?
Add a Skill in Power virtual agents
https://docs.microsoft.com/en-us/power-virtual-agents/configuration-add-skills#validation-performed-during-registering-a-skill

#### What language is this in?
Typescript

#### What happens?
When we try to executes steps mentioned in above document where are trying to add BF skill to PVA agent, following piece of code breaks 

/node_modules/botbuilder-core/lib/telemetryLoggerMiddleware.js
Function: fillReceiveEventProperties(...) has following lines 

```
properties[this.telemetryConstants.recipientIdProperty] = activity.recipient.id;
properties[this.telemetryConstants.recipientNameProperty] = activity.recipient.name;
```

As recipient is coming as NULL this is throwing an exception. So we had to modify the lines as follows which just sets recipient.id and recipient.name to empty string. 

```
properties[this.telemetryConstants.recipientIdProperty] = ( activity && activity.recipient && activity.recipient.id ) ? activity.recipient.id : '';

properties[this.telemetryConstants.recipientNameProperty] = ( activity && activity.recipient && activity.recipient.name ) ? activity.recipient.name : '';

```
#### What are the steps to reproduce this issue?
Steps 1: Create a BF TS skill (nodejs)
step 2: create a PVA bot (follow https://docs.microsoft.com/en-us/power-virtual-agents/fundamentals-get-started)
Step 3: Add a skill you created at step 1 to your PVA bot as step 2 using manifest file (follow https://docs.microsoft.com/en-us/power-virtual-agents/configuration-add-skills#validation-performed-during-registering-a-skill
)

#### What were you expecting to happen?
 We are expecting that recipient should be populated or this code check need to be added in. 

#### Can you share any logs, error output, etc.?
No

#### Any screenshots or additional context?

image

garypretty:
Fixes #2274 

Fixes an issue where an error was thrown during skill manifest validation from PVA.


libraries/botbuilder-core/src/telemetryLoggerMiddleware.ts
```diff
@@ -191,14 +191,14 @@ export class TelemetryLoggerMiddleware implements Middleware {
         properties[this.telemetryConstants.fromIdProperty] = activity.from ? activity.from.id : '';        
         properties[this.telemetryConstants.conversationNameProperty] = activity.conversation.name || '';        
         properties[this.telemetryConstants.localeProperty] = activity.locale || '';
-        properties[this.telemetryConstants.recipientIdProperty] = activity.recipient.id;
-        properties[this.telemetryConstants.recipientNameProperty] = activity.recipient.name;
+        properties[this.telemetryConstants.recipientIdProperty] = (activity && activity.recipient && activity.recipient.id) ? activity.recipient.id : '';
+        properties[this.telemetryConstants.recipientNameProperty] = (activity && activity.recipient && activity.recipient.name) ? activity.recipient.name : '';
```
stevengum, line 195: This will blow up on the first line if activity is null or undefined:

```ts
properties[this.telemetryConstants.fromIdProperty] = activity.from ? activity.from.id : ''; 
```

Another way of looking at it is that there isn't anything to log if there is no Activity, correct?

@garypretty can you confirm if we should add an initial `activity !== null && activity !== undefined` check to the following protected methods?

`onReceiveActivity()`
`fillReceiveEventProperties()`

garypretty: @stevengum we have a null check for activity before we call onReceiveActivity (which in turn calls fillReceiveEventProperties), so activity should never be null or undefined. https://github.com/microsoft/botbuilder-js/blob/73a63b52810071684991443d8a7ef6f5008eb51d/libraries/botbuilder-core/src/telemetryLoggerMiddleware.ts#L73-L79

stevengum: @garypretty I see your point; additionally the code here will also be in parity with C#. 

However, I think there's a flaw in both languages as all the methods in C# and TypeScript are virtual methods and public or protected.

If I override onTurn and then call the original onReceiveActivity, onReceiveActivity and fillReceiveEventProperties in the base class do not have the guarantee that context.activity is not null.

garypretty: Good catch - I can update both to check for null.

PR #2435, opened 2020-06-26
Add ActivityEx implementation

### LINKED ISSUE #1753: [Parity] ActivityEx is not present in Botbuilder-JS
We detected that the implementation of [ActivityEx ](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Schema/ActivityEx.cs)which is used in [DefaultActivityHandler ](https://github.com/microsoft/botframework-solutions/blob/master/templates/csharp/VA/VA/Bots/DefaultActivityHandler.cs#L78)(more specifically the [AsEventActivity method](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Schema/ActivityEx.cs#L113)) is present in the C# version but we didn't find the equivalent in Botbuilder-JS for TypeScript version. 
 
Migrate [ActivityEx ](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Schema/ActivityEx.cs)into Botbuilder-JS.

**Additional context**
At this moment, we are migrating the changes in the C# Virtual Assistant to the TypeScript Virtual Assistant to support the GA Skills SDK. We found the following scenarios:
	• To prevent the migration being blocked in [DefaultActivityHandler](https://github.com/microsoft/botframework-solutions/blob/master/templates/csharp/VA/VA/Bots/DefaultActivityHandler.cs#L78), we manually implemented the validation from AsEventActivity as a workaround; we found this same method in the Skill's [MainDialog](https://github.com/microsoft/botframework-solutions/blob/10ae4d2f0fa595e9d1b1f5d5ba11f15fd3160e9a/templates/csharp/Skill/Skill/Dialogs/MainDialog.cs#L230).
	• A similar case in the Virtual Assistant's [MainDialog](https://github.com/microsoft/botframework-solutions/blob/10ae4d2f0fa595e9d1b1f5d5ba11f15fd3160e9a/templates/csharp/VA/VA/Dialogs/MainDialog.cs#L311), the method AsMessageActivity is used in C# but is missing in TypeScript.
	• In the Skill's [DefaultAdapter ](https://github.com/microsoft/botframework-solutions/blob/10ae4d2f0fa595e9d1b1f5d5ba11f15fd3160e9a/templates/csharp/Skill/Skill/Adapters/DefaultAdapter.cs#L42)class, the method CreateEndOfConversationActivity is used in C# but is missing in TypeScript.

@darrenj can help providing any further information on this.

denscollo:
Fixes #1753

## Description
- Add the **_ActivityEx_** namespace with all the migrated methods from [ActivityEx](https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Schema/ActivityEx.cs) in .NET.
- Add unit tests for all the _ActivityEx_ methods.

### Testing
The image below shows the new tests passing.

image


denscollo:
Thanks @stevengum for the feedback. We just applied it and the CI passed. Thanks!

libraries/botframework-schema/tests/activityEx.test.js
```diff
@@ -0,0 +1,577 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+const assert = require('assert');
+const { ActivityTypes } = require('../lib/index');
+const { ActivityEx } = require('../lib/ActivityEx');
```
stevengum, line 8:
```suggestion
const { ActivityEx, ActivityTypes } = require('../');
```

libraries/botframework-schema/tests/activityEx.test.js
```diff
@@ -0,0 +1,577 @@
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+const assert = require('assert');
```
stevengum, line 6:
```suggestion
const { strictEqual } = require('assert');
```

And remove `assert.` occurrences

libraries/botframework-schema/src/activityEx.ts
```diff
@@ -0,0 +1,405 @@
...
+        return false;
+    }
+
+  /**
+   * Indicates whether this activity is of a specified activity type.
+   * @param source The source activity.
+   * @param activityType The activity type to check for.
+   * @returns True if the activity is of the specified activity type; otherwise, false.
+   */
+    export function isActivity(source:Partial<Activity>, activityType: string): boolean {
```
stevengum, line 372: Where is the test coverage for this function? What happens when `typeof source.type !== 'string'`?

libraries/botframework-schema/src/activityEx.ts
```diff
@@ -0,0 +1,405 @@
...
+       /*
+        * NOTE: While it is possible to come up with a fancy looking "one-liner" to solve
+        * this problem, this code is purposefully more verbose due to optimizations.
+        *
+        * This main goal of the optimizations was to make zero allocations because it is called
+        * by all of the .asXXXActivity methods which are used in a pattern heavily upstream to
+        * "pseudo-cast" the activity based on its type.
+        */
```
stevengum, line 380: Is this true? Or is this a blind copy-paste from the .NET implementation?

stevengum (changes requested):
Please see comments and run the files through `eslint`. Additionally, the helper functions in activityEx.test.js should be camel cased.

libraries/botframework-schema/src/activityEx.ts
denscollo: Hi @stevengum, this code was covered by the tests of the `asXXXActivity` methods as it is called by them. However, we added new tests to cover all possibilities.

libraries/botframework-schema/src/activityEx.ts
denscollo: Hi @stevengum, this was copied from .NET implementation but we consider it applies here as well, as the methods that call this one were almost identically migrated from .NET. We updated the comment to clarify this

libraries/botframework-schema/src/activityEx.ts
stevengum: Sorry, that of course tracks. However, thank you for adding individual tests as the `asXXXActivity` and `isActivity` methods are both public.

stevengum (approved):

PR #2446, opened 2020-06-29
Unionize TaskModuleResponse.task type definition

### LINKED ISSUE #1588: TaskModuleResponseBase interface definition does not support additional necessary fields
## Versions
4.7

## Describe the bug
The definition of TaskModuleResponseBase only includes the type field. However, there are several other fields used when responding to teams. 

This is problematic when using the typedefs and defining the `handleTeamsTaskModuleFetch` method on the ActivityHandler class.

## Expected behavior
The typedef should include the other possible fields, or at least allow wildcard fields.

[bug]

EricDahlvang:
Fixes #1588

### DIFF
--- a/libraries/botframework-schema/etc/botframework-schema.api.md
+++ b/libraries/botframework-schema/etc/botframework-schema.api.md
@@ -807,7 +807,7 @@ export interface MessagingExtensionAction extends TaskModuleRequest {
 // @public
 export interface MessagingExtensionActionResponse {
     composeExtension?: MessagingExtensionResult;
-    task?: TaskModuleResponseBase;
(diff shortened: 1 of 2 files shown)


EricDahlvang:
Hold merging this... @stevengum pointed out another place we should use this same union

stevengum:
@EricDahlvang and I still want to take this PR with the additional changes he mentioned.

Pedantically speaking, this is PR is a breaking change. However, Teams will not respect `TaskModuleResponseBases` that aren't one of the subtypes: `TaskModuleContinueResponse | TaskModuleMessageResponse`, so sending anything else will result in an error.

PR #2447, opened 2020-06-29
Make ConnectorClientOptions.baseUri optional

### LINKED ISSUE #2439: baseUri should be optional for ConnectorClientOptions
## Versions
NodeJS v10.15.3
BotBuilder NodeJS SDK v4.9.2
My app is written is typescript

## Describe the bug
When Using customized ConnectorClientOptions for BotFrameworkAdapterSettings, the baseUri is required, but passing a value for it causes the sdk to stop working - bot is unable to send messages

## To Reproduce
const botFrameworkAdapterSettings: BotFrameworkAdapterSettings = {
  appId,
  appPassword,
  clientOptions: { baseUri: undefined, httpClient: new CustomAxiosHttpClient() }
};
const adapter: BotFrameworkAdapter = new BotFrameworkAdapter(botFrameworkAdapterSettings);

## Expected behavior
The baseUri should be optional at the very least, that will allow me to omit it, and still being able to transpile my typescript app without errors

EricDahlvang:
Fixes #2439

### DIFF
--- a/libraries/botframework-connector/src/connectorApi/models/index.ts
+++ b/libraries/botframework-connector/src/connectorApi/models/index.ts
@@ -7,6 +7,10 @@ export * from "botframework-schema";
  * An interface representing ConnectorClientOptions.
  */
 export interface ConnectorClientOptions extends ServiceClientOptions {
+  /**
+   * (Optional) baseUri will be set automatically within BotFrameworkAdapter, 
+   * but is required if using the ConnectorClient outside of the adapter.
+   */
   baseUri?: string;
 }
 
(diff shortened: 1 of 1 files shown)


libraries/botframework-connector/src/connectorApi/models/index.ts
```diff
@@ -7,7 +7,7 @@ export * from "botframework-schema";
  * An interface representing ConnectorClientOptions.
  */
 export interface ConnectorClientOptions extends ServiceClientOptions {
-  baseUri: string;
+  baseUri?: string;
```
stevengum, line 10: @EricDahlvang can you add a comment indicating why this is optional? Specifically around the behavior that takes place inside of the BotFrameworkAdapter that applies the correct `baseUri`.

If someone uses the ConnectorClient without the BFAdapter, the `baseUri` is required.

EricDahlvang: good point.  comment added

PR #2466, opened 2020-07-01
Open id metadata cache refresh if key is missing

### LINKED ISSUE #2442: New OpenIdMetadata signing keys will not be picked up for up to 5 days
## Versions
4.9

(the rest is irrelevant)

## Describe the bug
If MSFT adds a new signing key to the OpenIdMetadata it will not be picket up for 5 days and fail Directline Speech connections.

The issue seems to be in OpenIdMetadata: https://github.com/microsoft/botbuilder-js/blob/60ff913ea0199afb7e5ff96f6ebf25648a75479b/libraries/botframework-connector/src/auth/openIdMetadata.ts#L26

On incoming requests, JwtTokenExtractor will validate the JWT token using the cached public keys loaded via OpenIdMetadata.

MSFT adding (and using!) a new signing key will not be picked up for up to 5 days, which means that if a new signing key is being used, L42/43 will not find it in the cached keys and return undefined, which in turn fails OpenIdMetadata L98 (https://github.com/microsoft/botbuilder-js/blob/60ff913ea0199afb7e5ff96f6ebf25648a75479b/libraries/botframework-connector/src/auth/jwtTokenExtractor.ts#L98).

## To Reproduce
Currently not reproducible as there's no new signing key. But the description should explain how it works.

## Expected behavior
My suggestion would be to always check if the signing key actually exists in the cache and refresh the tokens if it doesn't. Only then can you be sure that the signing key is completely invalid.

## Screenshots
-

## Additional context
We saw this issue on Saturday for a few hours and restarting our service fixed the issue as it refreshed the memory cache.

[bug]

EricDahlvang:
Fixes #2442

### DIFF
--- a/libraries/botframework-connector/tests/openIdMetadata.test.js
+++ b/libraries/botframework-connector/tests/openIdMetadata.test.js
@@ -1,0 +1,104 @@
+const assert = require('assert');
+const nock = require('nock');
+const { JwtTokenValidation } = require('../lib');
+const { StatusCodes } = require('botframework-schema');
+const { OpenIdMetadata } = require('../lib/auth/openIdMetadata');
+
+describe('OpenIdMetadata', function() {
+
+    describe('Cache Tests', function() {
+        this.timeout(20000);
+
+        const mockUrl = 'http://mockUrl';
+        const mockMetadataUrl = '/api/metadataUrl';
+        const jwks_uriUrl = '/api/jwks_uriUrl';
+
+        // the below key and parts were retrieved from an Emulator token
+        const e = 'AQAB';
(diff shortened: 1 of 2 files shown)


stevengum:
@carlosscastro when you have the chance can you review this as well?

@EricDahlvang thanks for adding the test! 😄

libraries/botframework-connector/tests/openIdMetadata.test.js
```diff
@@ -0,0 +1,104 @@
+const assert = require('assert');
```
carlosscastro, line 1: copyright header

EricDahlvang: added

PR #2479, opened 2020-07-03
add directlinespeech support to prompts

stevengum:
Fixes in JS: https://github.com/microsoft/botframework-sdk/issues/5950

## Description
Adds directlinespeech to channel.ts which determines the suggestedAction and cardAction support in the prompts. 

The buttonCount limits follow Web Chat and Directline because this is in the context of Direct Line Speech via Web Chat.

## Specific Changes
  - Adds directlinespeech to channel.ts helper functions
  - Add tests for channel.ts changes
  - Regenerate `botbuilder-dialogs.api.md`

## Testing
Unit tests and manually tested via [06.using-cards](https://github.com/microsoft/BotBuilder-Samples/tree/f45a096cdc8b6938d6f1090e8bdc7475706ab81b/samples/javascript_nodejs/06.using-cards) & [03.speech/a.direct-line-speech](https://github.com/microsoft/BotFramework-WebChat/tree/3bab93722158e73645db15ac675da1f7c8ae8931/samples/03.speech/a.direct-line-speech) on chromium edge.

FYI @varunbmallik, @compulim, @NickEricson, @dmvtech


libraries/botbuilder-dialogs/etc/botbuilder-dialogs.api.md
```diff
@@ -254,6 +253,7 @@ export abstract class DialogContainer<O extends object = {}> extends Dialog<O> {
     readonly dialogs: DialogSet;
     findDialog(dialogId: string): Dialog | undefined;
     protected getInternalVersion(): string;
+    telemetryClient: BotTelemetryClient;
```
compulim, line 256: I am unsure if this move is needed, or just accidentally got in.

libraries/botbuilder-dialogs/tests/choices_channel.test.js
```diff
@@ -1,5 +1,6 @@
 const assert = require('assert');
 const { supportsSuggestedActions, supportsCardActions, hasMessageFeed, getChannelId } = require('../lib');
+const { Channels } = require('botbuilder-core');
```
compulim, line 3: Sort imports?

libraries/botbuilder-dialogs/tests/choices_channel.test.js
```diff
@@ -98,4 +99,15 @@ describe('channel methods', function() {
         const channel = getChannelId({ activity: { } });
         assert(channel === '', `expected "", instead received ${ channel }`);
     });
+
+    // "directlinespeech" tests
+    it(`should return true for supportsSuggestedActions() with directlinespeech and 100`, function () {
+        const validNumOfSuggestedActions = supportsSuggestedActions(Channels.DirectlineSpeech, 100);
+        assert(validNumOfSuggestedActions, `returned false.`);
```
compulim, line 106: `returned true.`?

I am very confused since other tests have similar issues.

libraries/botbuilder-dialogs/etc/botbuilder-dialogs.api.md
stevengum: This is a side-effect of regenerating the API report for the library; it wasn't an intentional change, but this line would eventually be added when the report is regenerated.

libraries/botbuilder-dialogs/tests/choices_channel.test.js
stevengum: The `assert` used here is an alias for `assert.ok` which performs a truthy evaluation; the "returned false." is a custom error message when the `assert` fails.

These tests should be refactored, but that's out of this PR's scope and in @axelsrz's Code Quality pillar. I'll coordinate with him on updating these and other tests.

libraries/botbuilder-dialogs/tests/choices_channel.test.js
stevengum: How embarassing... Done!

PR #2489, opened 2020-07-07
Return ResourceResponse from SkillHandler.processActivity

### LINKED ISSUE #2358: SkillHandler doesn't return ResourceResponse when forwarding activities (JS)
See [parent](https://github.com/microsoft/botframework-sdk/issues/5919)

EricDahlvang:
Fixes #2358


libraries/botbuilder/src/skills/skillHandler.ts
```diff
@@ -218,7 +221,11 @@ export class SkillHandler extends ChannelServiceHandler {
         AppCredentials.trustServiceUrl(skillConversationReference.conversationReference.serviceUrl);
 
         await (this.adapter as BotFrameworkAdapter).continueConversation(skillConversationReference.conversationReference, skillConversationReference.oAuthScope, callback);
-        return { id: uuid() };
+        
+        if (!resourceResponse) {
+            resourceResponse = { id: uuid() };
```
gabog, line 226: Can you check with @johnataylor and make sure the implementation is consistent with what he is implementing in dotnet?

gabog: Checked the dotnet version and this looks good.

PR #2497, opened 2020-07-07
IUserTokenProvider.getTokenStatus interface updates and test

### LINKED ISSUE #1247: IUserTokenProvider is missing getTokenStatus (parity with c#)
https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/userTokenProvider.ts 

vs

https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder/IUserTokenProvider.cs#L66

[bug]

EricDahlvang:
Fixes #1247


libraries/botbuilder-core/tests/testAdapter.test.js
```diff
@@ -496,4 +496,19 @@ describe(`TestAdapter`, function () {
...
         adapter.send('hi');
     });
+
+    it(`should return statuses from getTokenStatus`, function (done) {
+        const adapter = new TestAdapter((context) => {
+            context.adapter.getTokenStatus(context, 'user').then(statuses => {
+                assert(statuses);
+                assert(statuses.length == 2);
+                assert(statuses.reduce((j, status) => (j||status.ConnectionName === "ABC"), false));
+                done();
```
stevengum, line 506: If an error is thrown in `.then()` in the `logic` it'll give an unhelpful `done not called` error and an uncaughtpromiseexception.

Since [`logic`](https://github.com/microsoft/botbuilder-js/blob/c39c7744eaf1bc82e897789202420659a03c460f/libraries/botbuilder-core/src/testAdapter.ts#L101) returns `Promise<void>`, can you rewrite the `logic` using async-await syntax?  You could still use `done` in combination with a try-catch to make it easier to parse. calling done with an error will cause the test to fail.

libraries/botbuilder-core/src/testAdapter.ts
```diff
@@ -310,6 +310,33 @@ export class TestAdapter extends BotAdapter implements ExtendedUserTokenProvider
         }
     }
 
+    public async getTokenStatus(context: TurnContext, userId: string, includeFilter?: string, oAuthAppCredentials?: any): Promise<any[]> {
+        const filter = (includeFilter ? includeFilter.split(',') : undefined);
+        const match = this._userTokens.filter(x => x.ChannelId === context.activity.channelId 
+                                            && x.UserId === context.activity.from.id
```
stevengum, line 316: If context.activity or context.activity.from is falsy this method will throw an error, can you add falsy checks when the method is invoked? (And tests too 😄)

EricDahlvang: Sure, but this is test code ... not the product.  It feels weird to writes tests validating parameters of test code.

libraries/botbuilder-core/tests/testAdapter.test.js
```diff
@@ -498,17 +498,52 @@ describe(`TestAdapter`, function () {
...
     });
+
+    it(`should throw when context parameter is not sent`, function (done) {
+        const adapter = new TestAdapter(async (context) => {
+            try {
+                await context.adapter.getTokenStatus();
+            } catch (err) {
+                assert(err.message === 'testAdapter.getTokenStatus(): context with activity is required',
+                `expected "testAdapter.getTokenStatus(): context with activity is required" Error message, not "${ err.message }"`);
+                done();
```
stevengum, line 526: I'm not a huge fan of this suggestion, but if the assert fails the test fails due to a timeout while waiting for `done` to be called.

You should be able to avoid this by doing the following:

```suggestion
                done(assert.strictEqual(err.message, 'testAdapter.getTokenStatus(): context with activity is required'));
```

PR #2515, opened 2020-07-10
add readableFlowing property to INodeSocket

mdrichardson:
All JS builds are breaking I believe due to a linting or pipeline update(?); I didn't really see any other relevant recent commits. INodeSocket needs to implement `readableFlowing` so I added it as a property that does nothing. It can't be optional or the build will still break. It uses the same typing and scope as `Socket` defined in Node globals:

image

[Here's the failure](https://fuselabs.visualstudio.com/SDK_v4/_build/results?buildId=145306&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=df8dc50d-985d-501e-ee67-7f805e7e444c&l=49):

```cmd
src/webSocket/nodeWebSocket.ts(39,69): error TS2345: Argument of type 'INodeSocket' is not assignable to parameter of type 'Socket'.
> botframework-config@4.4.0-preview.145306 build /Users/runner/work/1/s/libraries/botframework-config
  Property 'readableFlowing' is missing in type 'INodeSocket' but required in type 'Socket'.
```

[The reference for the change](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45997) doesn't amount to much more than "it was missing, so we added it".

### DIFF
--- a/libraries/botframework-streaming/src/interfaces/INodeSocket.ts
+++ b/libraries/botframework-streaming/src/interfaces/INodeSocket.ts
@@ -28,6 +28,7 @@ export interface INodeSocket {
     readonly writableLength: number;
     readonly readableHighWaterMark: number;
     readonly readableLength: number;
+    readonly readableFlowing: boolean | null;
 
     address(): AddressInfo | string;
 


chon219:
I pulled the latest code from master branch and got the following error when I ran "npm run build", is there anything broken?

```
lerna ERR! npm run build exited 2 in 'botframework-streaming'
lerna ERR! npm run build stdout:

> botframework-streaming@4.1.6 build /home/chon/workspace/botbuilder-js/libraries/botframework-streaming
> tsc -p tsconfig.json && tsc -p tsconfig-browser.json

src/namedPipe/namedPipeClient.ts(60,53): error TS2345: Argument of type 'Socket' is not assignable to parameter of type 'INodeSocket'.
  Property 'readableFlowing' is missing in type 'Socket' but required in type 'INodeSocket'.
src/namedPipe/namedPipeClient.ts(61,55): error TS2345: Argument of type 'Socket' is not assignable to parameter of type 'INodeSocket'.
```

stevengum:
@chon219, if you run lerna bootstrap —hoist again you’ll have the latest @types/node dependency which should resolve your build error.

chon219:
> @chon219, if you run lerna bootstrap —hoist again you’ll have the latest @types/node dependency which should resolve your build error.

Thanks, I would try again.

chon219:
> @chon219, if you run lerna bootstrap —hoist again you’ll have the latest @types/node dependency which should resolve your build error.

I've updated @types/node dependency to 10.17.26 via running `lerna bootstrap --hoist` but still got the same build error, it's so weird.

mdrichardson:
@chon219 I just cloned and am not running into this. If you try cloning the repo into a new folder, do you still experience this? And you're using the `botbuilder-js` repo, right? Do you see the `readableFlowing` property in `libraries/botframework-streaming/src/interfaces/INodeSocket.ts `?

stevengum:
@chon219 if you do a git clean -fdx at the root level and then do lerna bootstrap —hoist that will ensure the old node types and package-lock.jsons are gone.

chon219:
@mdrichardson yes, there is a readableFlowing property
image

I cloned a new copy of botbuilder-js and it builds well without errors, but I don't know why this would happen...

chon219:
> @chon219 if you do a git clean -fdx at the root level and then do lerna bootstrap —hoist that will ensure the old node types and package-lock.jsons are gone.

Got it, thanks!

PR #2533, opened 2020-07-14
unhandled undefined conditions#2532

### LINKED ISSUE #2532: unhandled undefined conditions in componentDialog
## Versions
What package version of the SDK are you using. - 4.8.0
What nodejs version are you using  - 13.11.0
What browser version are you using -  Emulator
What os are you using - Mac OS

## Describe the bug
unchecked undefined variable causing problems in bot application

## To Reproduce
1. from a Bot mainDialog starts a secondary dialog by calling beginDialog(secondaryDialog). In this secondaryDialog the initStep throw some exception.

TypeError: Cannot read property 'status' of undefined
    at HoursWorkedDialog.<anonymous> (/Users/c/gitsrc/ext/teams-chatbot/node_modules/botbuilder-dialogs/lib/componentDialog.js:86:30)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/c/gitsrc/ext/teams-chatbot/node_modules/botbuilder-dialogs/lib/componentDialog.js:4:58) {stack: 'TypeError: Cannot read property 'status' of u…tbuilder-dialogs/lib/componentDialog.js:4:58)', message: 'Cannot read property 'status' of undefined'}

or error:

TypeError: Cannot read property 'status' of undefined
    at MyScheduleDialog.<anonymous> (/Users/c/gitsrc/ext/teams-chatbot/node_modules/botbuilder-dialogs/lib/componentDialog.js:100:28)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/c/gitsrc/ext/teams-chatbot/node_modules/botbuilder-dialogs/lib/componentDialog.js:4:58)
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {stack: 'TypeError: Cannot read property 'status' of u…ctions (internal/process/task_queues.js:97:5)', message: 'Cannot read property 'status' of undefined'}

## Expected behavior
This error should not be thrown. The beginDialog should be able to handle undefined turnResult.

[bug]

cindylarison:
Fixes #2532
## Description
unhandled undefined condition in componentDialog

## Specific Changes
adding undefined checks

## Testing
tested with my bot

### DIFF
--- a/libraries/botbuilder-dialogs/src/componentDialog.ts
+++ b/libraries/botbuilder-dialogs/src/componentDialog.ts
@@ -88,7 +88,7 @@ export class ComponentDialog<O extends object = {}> extends DialogContainer<O> {
         const turnResult: DialogTurnResult<any> = await this.onBeginDialog(innerDC, options);
 
         // Check for end of inner dialog
-        if (turnResult.status !== DialogTurnStatus.waiting) {
+        if (turnResult !== undefined && turnResult.status !== DialogTurnStatus.waiting) {
             if (turnResult.status === DialogTurnStatus.cancelled) {
                 await this.endComponent(outerDC, turnResult.result);
                 const cancelledTurnResult: DialogTurnResult = { status: DialogTurnStatus.cancelled, result: turnResult.result }
@@ -109,7 +109,7 @@ export class ComponentDialog<O extends object = {}> extends DialogContainer<O> {
         const turnResult: DialogTurnResult<any> = await this.onContinueDialog(innerDC);
 
         // Check for end of inner dialog
-        if (turnResult.status !== DialogTurnStatus.waiting) {
+        if (turnResult !== undefined && turnResult.status !== DialogTurnStatus.waiting) {
             // Return result to calling dialog
             return await this.endComponent(outerDC, turnResult.result);
         }


johnataylor:
A reasonable fix might be to test for undefined and then throw an Exception with the text "dialog result from inner dialog should not be null"

In the C# the exception type would be NullReferenceException but with this text.

Anyhow, the current contract is that this shouldn't be null and if it is then an exception is throw. We can make the exception more self-describing but that is all.

cindylarison:
> A reasonable fix might be to test for undefined and then throw an Exception with the text "dialog result from inner dialog should not be null"
> 
> In the C# the exception type would be NullReferenceException but with this text.
> 
> Anyhow, the current contract is that this shouldn't be null and if it is then an exception is throw. We can make the exception more self-describing but that is all.

Hi, I just want to know how I should handle this error in my bot.

for both cases, I have a mainDialog which invoke secondary dialogs that runs REST API to get some information. if these REST API failed the secondary dialogs will through exception. These exceptions are causing the null turnContext. Because the secondary dialogs are not returned. Should I catch the exception with in the secondary dialog and just return some context? how do I signal there is an error?

cindylarison:
Hi @johnataylor ,

I just tested my bot. Even when I catch the errors within the secondary dialogs, my mainDialog still see undefined context.  I don't see anyway I can control that.

johnataylor:
@cindylarison let me take a look at that and I'll update this issue.

johnataylor:
closing this PR. we can't take a change of behavior like this here at this point. possibly a clearer exception text - if you feel strongly about that please open an issue (you can reference this PR and the other issue.)

PR #2535, opened 2020-07-15
[PORT] Add support for operations on arrays

### LINKED ISSUE #1994: [PORT] Add support for operations on arrays
https://github.com/microsoft/botbuilder-dotnet/pull/3669

This adds support for operations on arrays.  The API surface area is the same except that the OnClearProperty event was removed.  The internals were also simplified to make it clearer what is going on.  There will be a separate check-in that updates the samples to use arrays.  I wanted to keep this review simpler.

# Changed projects

chon219:
Fixes #1994 

## Description

> microsoft/botbuilder-dotnet#3669

This adds support for operations on arrays. The API surface area is the same except that the OnClearProperty event was removed. The internals were also simplified to make it clearer what is going on. There will be a separate check-in that updates the samples to use arrays. I wanted to keep this review simpler.

## Specific Changes

  - 
  -
  -


libraries/botbuilder-dialogs/src/memory/dialogPath.ts
```diff
@@ -13,6 +13,9 @@ export class DialogPath {
     /// Currently expected properties.
     public static readonly expectedProperties = 'dialog.expectedProperties';
 
+    /// Default operation to use for entities where there is no identified operation entity.
+    public static readonly defaultOperation = 'dialog.defaultOperation';
```
chrimc62, line 17: >defaultOperation [](start = 54, length = 16)

Does this need to be $defaultOperation?  I ask because that is the key in adaptiveDialog.

libraries/botbuilder-dialogs-adaptive/src/adaptiveDialogComponentRegistration.ts
```diff
@@ -12,7 +12,7 @@ import { AdaptiveTypeBuilder } from './adaptiveTypeBuilder';
...
 import { Ask, AttachmentInput, ChoiceInput, ConfirmInput, DateTimeInput, NumberInput, OAuthInput, TextInput, AttachmentOutputFormat, ChoiceOutputFormat } from './input';
-import { OnActivity, OnAssignEntity, OnBeginDialog, OnCancelDialog, OnChooseEntity, OnChooseIntent, OnChooseProperty, OnClearProperty, OnCondition, OnConversationUpdateActivity, OnDialogEvent, OnEndOfActions, OnEndOfConversationActivity, OnError, OnEventActivity, OnHandoffActivity, OnIntent, OnInvokeActivity, OnMessageActivity, OnMessageDeleteActivity, OnMessageReactionActivity, OnMessageUpdateActivity, OnQnAMatch, OnRepromptDialog, OnTypingActivity, OnUnknownIntent } from './conditions';
```
chrimc62, line 15: Why was this removed?

chrimc62:
>     /**

Document

---
Refers to: libraries/botbuilder-dialogs-adaptive/src/entityInfo.ts:12 in 9e072ea. [](commit_id = 9e072eaf5ed7f7c782bc7d9413fcb2c6303aaaaa, deletion_comment = False)

chrimc62:
>     /**

Document

---
Refers to: libraries/botbuilder-dialogs-adaptive/src/schemaHelper.ts:11 in 9e072ea. [](commit_id = 9e072eaf5ed7f7c782bc7d9413fcb2c6303aaaaa, deletion_comment = False)

libraries/botbuilder-dialogs/src/memory/dialogPath.ts
chon219: > > defaultOperation [](start = 54, length = 16)
> 
> Does this need to be $defaultOperation? I ask because that is the key in adaptiveDialog.

I don't think so, we have a private property in adaptiveDialog named _defaultOperationKey_ which represents the key $defaultOperation, which is a part of schema.

libraries/botbuilder-dialogs-adaptive/src/entityAssignments.ts
```diff
@@ -0,0 +1,62 @@
...
+     * @param actionContext Memory context.
+     */
+    public write(actionContext: ActionContext): void {
+        actionContext.state.setValue(events, this);
+    }
+
+    /**
+     * Returns the next entity event to surface.
+     */
+    public nextAssignment(): Partial<EntityAssignment> {
```
stevengum, line 44: This seems like it should just be a getter. And possibly renamed?

chon219: Got it.

Issue #2432, opened 2020-06-26
Increase coverage for botbuilder library


axelsrz:
Increase unit test coverage of botbuilder package to 80%

cleemullins:
Consider for Southworks to complete.

denscollo:
We divided the issue into the existing classes in which we'll increase the CC, and create a PR for each one.

This is the list and its status:
#2484 ChannelServiceHandler
#2485 InspectionMiddleware
BotFrameworkAdapter
BotFrameworkHttpClient
TeamsInfo
TeamsActivityHandler
StatusCodeError
ChannelServiceRoutes

axelsrz:
Closing due to complition.

PR #2540, opened 2020-07-15
Increase code coverage in botbuilder library - ChannelServiceRoutes class

denscollo:
Addresses #2432
## Description
This PR adds the `channelServiceRoutes.test.js` with unit tests for the [ChannelServiceRoutes](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder/src/channelServiceRoutes.ts) class, increasing code coverage for the class to 88,97%.

## Testing
image
image


EricDahlvang:
This branch is out of date.  Please rebase from master (cannot do this automatically from github, because the PR is from a fork)

libraries/botbuilder/tests/channelServiceRoutes.test.js
```diff
@@ -0,0 +1,776 @@
...
+        describe('readBody()', async () => {
+            it('should return request.body from "on" events', async () => {
+                const source = { test: true };
+
+                const request = {
+                    ...req,
+                    body: null
+                };
+
+                setTimeout(() => {
```
EricDahlvang, line 765: Can this be rewritten to avoid using a setTimeout?  Or, please add some comments explaining why it is necessary for this test.

denscollo:
@EricDahlvang  We just applied the feedback and update it with the master branch, so it's ready. Thanks!

libraries/botbuilder/tests/channelServiceRoutes.test.js
```diff
@@ -762,14 +762,17 @@ describe('channelServiceRoutes', function() {
...
-                }, 100);
-
-                ChannelServiceRoutes.readBody(request).then((body) => {
-                    assert.deepStrictEqual(body, source, `expected: ${ JSON.stringify(source) }. received: ${ JSON.stringify(body) }`);
+                // readBody Promise wrapper to resolve after the readBody request.on('data') and request.on('end') are assigned.
+                await new Promise(async resolve => {
+                    ChannelServiceRoutes.readBody(request).then((body) => {
+                        assert.deepStrictEqual(body, source, `expected: ${ JSON.stringify(source) }. received: ${ JSON.stringify(body) }`);
+                    });
+                    resolve();
```
stevengum, line 770: If an error is thrown inside of the async lambda, the test will likely still pass. Please test that errors are correctly surfaced.

denscollo: @stevengum Thanks, we'll check this again, and add the then/catch clause to this tests also.

denscollo:
@stevengum  We just pushed the feedback. Thanks!

PR #2547, opened 2020-07-17
Clean up emulator check in Cosmos Test

mdrichardson:
## Description

Cosmos tests would throw errors in prep and cleanup scripts if both:

1. Cosmos Emulator installed at appropriate path
2. Cosmos Emulator was not running

## Specific Changes

The Emulator check now only tries a GET request at the Emulator endpoint. If there is no response, it skips the prep and cleanup steps that would require emulator.


mdrichardson:
@stevengum I wouldn't say that it's _strictly_ necessary. Because `cleanup` erases all DBs and it isn't called until `after` in the "Base Storage Tests" section, there's a small chance that an existing DB could interfere with _those_ tests.

PR #2567, opened 2020-07-23
Updating Entity interface

### LINKED ISSUE #2561: Entity should allow custom fields
## Versions
4.n

## Describe the bug
Entity type should allow custom fields:
https://github.com/microsoft/botframework-sdk/blob/master/specs/botframework-activity/botframework-activity.md#entity

The mapper is currently stripping custom Entity fields.

## Additional context
From @stevengum 👍

Ideally, our Entity interface should look like the following:
image

The corresponding mapper should also support the updated interface.

___

Alternatively, any Mapper that references the Entity CompositeMapper (i.e. SemanticAction and Activity CompositeMappers) should replace the current the Composite type “Entity” with the autorest equivalent of OpenAPI’s “allOf” while using “Entity” and “Object” Mappers.
 
As an example, the Activity’s entities property would then be defined as such in the interface:
 
image

 

[bug]

Virtual-Josh:
Fixes #2561


libraries/botframework-schema/etc/botframework-schema.api.md
```diff
@@ -46,7 +46,7 @@ export interface Activity {
     code?: EndOfConversationCodes | string;
     conversation: ConversationAccount;
     deliveryMode?: DeliveryModes | string;
-    entities?: Entity[];
+    entities?: Entity & {[key: string]: any})[];
```
stevengum, line 49: This is incorrect with the change below which indicates that the `Entity` will take KvP mapping string to any. Was this change performed by hand?

Virtual-Josh: reverted this change after talking to steven

stevengum:
Original issue was resolved via https://github.com/microsoft/botbuilder-js/pull/2648, closing PR.

PR #2586, opened 2020-07-27
Correctly Call QnA Train API & Handle 204 No-Content Response

### LINKED ISSUE #2518: Wrong JSON Serialization
https://github.com/microsoft/botbuilder-js/blob/d79b10b3997fef42a22617601913ed68022a6390/libraries/botbuilder-ai/src/qnamaker-utils/trainUtils.ts#L49

```
 var payloadBody = JSON.stringify({
            feedbackRecords: feedbackRecords
        });
```

`feedbackRecords`  is already in the format of FeedbackRecords, so need not be wrapped in `{feedbackRecords: records}` for serialization.

This breaks trainAPI requests in active learning as bad http request.

Zerryth:
Fixes #2518 
___

## Description

Currently the SDK sends a malformed payload body to QnA Maker's [Train API](https://docs.microsoft.com/en-us/azure/cognitive-services/qnamaker/how-to/improve-knowledge-base#train-api).

Fix will:

* Send correct payload body to Train API
* Handle the Train API's response of `204 No-Content` properly, 
    * as currently [`HttpRequestUtils`](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/qnamaker-utils/httpRequestUtils.ts#L55) always tries to decode response to json, even when there isn't any content, which will throw a `FetchError: invalid json response body at https://xxxxx.azurewebsites.net/qnamaker/knowledgebases/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/train reason: Unexpected end of JSON input`

## Specific Changes
  - As expected by the Train API, in the request's payload body, send the *actual array* of `FeedbackRecord`'s instead of the whole `FeedbackRecords` object, which has the array as a member. 
  - Handle sending correct `QnAMakerResult` for no-content responses from QnA Maker service


libraries/botbuilder-ai/src/qnamaker-utils/httpRequestUtils.ts
```diff
@@ -45,14 +45,20 @@ export class HttpRequestUtils {
...
+                timeout: timeout,
+                body: payloadBody
+            });
+            const jsonResult = qnaResult.status == 204 ? this.getSuccessfulTrainApiResult() : await qnaResult.json();
+
+            return jsonResult;
+
+        } catch (error) {
+            throw error;
+        }
```
stevengum, line 61: The catch only rethrows the error, so it is unnecessary.

libraries/botbuilder-ai/src/qnamaker-utils/httpRequestUtils.ts
```diff
@@ -45,14 +45,20 @@ export class HttpRequestUtils {
...
-        
-        return await qnaResult.json();
+        try {
+            const qnaResult = await fetch(requestUrl, {
+                method: 'POST',
+                headers: headers,
+                timeout: timeout,
+                body: payloadBody
+            });
+            const jsonResult = qnaResult.status == 204 ? this.getSuccessfulTrainApiResult() : await qnaResult.json();
```
stevengum, line 55: @Zerryth @vipeketi, can we validate the requestUrl ahead of time to know that we're specifically calling the Train API? Do other APIs return 204 status codes?

libraries/botbuilder-ai/tests/qnaMaker.test.js
```diff
@@ -778,6 +780,105 @@ describe('QnAMaker', function () {
...
+                        endpoint.qnaHost + trainApi,
+                        payloadBody,
+                        endpoint
+                    );
+                    
+                    assert.deepStrictEqual(qnaResult, successfullyTrainedResult);
+
+                } catch (error) {
+                    throw error;
+                }
```
stevengum, line 811: Unnecessary catch

libraries/botbuilder-ai/tests/qnaMaker.test.js
```diff
@@ -778,6 +780,105 @@ describe('QnAMaker', function () {
...
         });
     });
+
+    describe('Active Learning', function() {
+        it('executeHttpRequest() should return JSON result from Train API response of 204 No-Content', function() {
+            const { endpoint, trainApi, feedbackRecords, successfullyTrainedResult } = new TrainApiHelper();
+
+            async () => await executeHttpRequestToTrainApi();
+        
+            async function executeHttpRequestToTrainApi() {
```
stevengum, line 790: Please move the arrangement to before the act in the test.

libraries/botbuilder-ai/tests/qnaMaker.test.js
```diff
@@ -778,6 +780,105 @@ describe('QnAMaker', function () {
...
+
+                try {
+                    const httpUtils = new HttpRequestUtils();
+                    const qnaResult = await httpUtils.executeHttpRequest(
+                        endpoint.qnaHost + trainApi,
+                        payloadBody,
+                        endpoint
+                    );
+                    
+                    assert.deepStrictEqual(qnaResult, successfullyTrainedResult);
```
stevengum, line 807: If the `executeHttpRequestToTrainApi()` returns `httpUtils.executeHttpRequest()`, you can evaluate the `qnaResult` outside of the helper function and simplify this code.

libraries/botbuilder-ai/tests/qnaMaker.test.js
```diff
@@ -800,3 +901,30 @@ class overrideTwoEventsWithOverrideLogger extends QnAMaker
             });
     }
 }
+
+class TrainApiHelper
+{
+    constructor() {
```
stevengum, line 907: Lambda that returns an object? Do we need to have the class since the constructor doesn't take anything.

libraries/botbuilder-ai/tests/qnaMaker.test.js
```diff
@@ -778,6 +780,105 @@ describe('QnAMaker', function () {
...
+            };
+
+        });
+
+        it('executeHttpRequest() should throw when payload to QnA Service is malformed', function() {
+            const { endpoint, trainApi, feedbackRecords, successfullyTrainedResult } = new TrainApiHelper();
+
+            async () => await executeHttpRequestToTrainApi();
+        
+            async function executeHttpRequestToTrainApi() {
```
stevengum, line 821: Please see feedback above

libraries/botbuilder-ai/src/qnamaker-utils/httpRequestUtils.ts
Zerryth: Hmm, as far as I know, Train API was the only one that returns 204, but I say that with a grain of salt, as I think there may be more newer qna APIs that I haven't gotten a chance to play with quite yet.

However even if it is just the Train API for now that returns no-content, I think I see what you're getting at, and I'll rework this to verify that they're calling the Train API 1st, since maybe in the future some other new qna API might return no-content, and it would be wrong to return a qna result that said that training was successful, if it wasn't that very API we called

vipeketi:
@Zerryth  Could you test with sample 49, to verify that nothing is broken.

libraries/botbuilder-ai/src/qnamaker-utils/httpRequestUtils.ts
Zerryth: removed trycatch

libraries/botbuilder-ai/tests/qnaMaker.test.js
Zerryth: removed trycatch

libraries/botbuilder-ai/tests/qnaMaker.test.js
Zerryth: Sorry for the wonky test arrangement. I was running into the "await can only be used in an async function" error earlier, *completely* forgetting that the mocha test cases are encapsulated within a function already which I could easily make async 🤦 🤦 Was doing backflips trying to wrap things in random async helper functions 😅 

Made callback in `it()` `async`, so no helper function needed => cleaner, and tests are actually arranged properly now 👍

libraries/botbuilder-ai/tests/qnaMaker.test.js
Zerryth: answer: no, no class needed, indeed!

Just moved variables to the `describe('Active Learning'...)` level, so it's accessible to all test added here

Zerryth:
> @Zerryth Could you test with sample 49, to verify that nothing is broken.

@vipeketi done and done! I actually made sure to build this up testing off of sample `49.qnamaker-all-features` to make sure I didn't break any other qna APIs unexpectedly
___

@stevengum, I actually think we need to return either `null` or `undefined` unfortunately instead of a `QnAMakerResult`, which yes, this would be different from what you would expect back from the Generate Answer API 😞 

So the call stack goes like this: 
*`TrainUtils.callTrain()` --> `TrainUtils.queryTrain()` --> `HttpRequestUtils.executeHttpRequest()` --> `fetch(request)`*

* If you look at [`TrainUtils`](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/qnamaker-utils/trainUtils.ts#L53), you can see `callTrain` is already public and **returns nothing** (does not return result from `executeHttpRequest`), since `queryTrain` returns nothing. 

* As it is currently built `executeHttpRequest` , [expects to always return a json](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-ai/src/qnamaker-utils/httpRequestUtils.ts#L55), as indicated by trying to decode the response to JSON, however this is probably where the mistake is
    * responses from `fetch` can be all sorts of shapes, including those without content, so therefore it shouldn't always expect a json response

___
I propose: 
* Leave `callTrain` & `queryTrain` as they currently in the SDK, returning nothing
* Return *something* from `executeHttpRequest` in all cases. That *something* can either be:
   * At least the status code 204 by itself, so it can communicate the Train API's response to the caller
   * Inject a JSON that conforms to the shape of a `QnAMakerResult` (from perspective of always trying to make sure the function returns a consistent single type and have a predictable shape)

libraries/botbuilder-ai/src/qnamaker-utils/httpRequestUtils.ts
vipeketi: @stevengum delete, update, replace  APIs return 204 status codes.. Guess, GenerateAnswer and Train API are more relevant to the context .. Handling TrainAPI for 204 should suffice for current situation.

Zerryth:
Ok, I took the approach of:
* Leave `callTrain` & `queryTrain` as they currently in the SDK, returning nothing
and
* for `executeHttpRequest`: inject an object that conforms to the shape of a QnAMakerResult (from perspective of always trying to make sure the function returns a consistent single type and have a predictable shape)

@stevengum @vipeketi let me know if you would prefer something else otherwise

libraries/botbuilder-ai/tests/qnaMaker.test.js
Zerryth: removed helper function executeHttpRequestToTrainApi()

libraries/botbuilder-ai/src/qnamaker-utils/httpRequestUtils.ts
Zerryth: Updated code (again) to handle more than just specifically the Train API when a 204 response occurs.
This is better for future proofing as well, since who knows if in the future if QnA will include more APIs that respond with 204

...ok I'm going to bed 😅 
lmk if we want to go a diff direction

Issue #2512, opened 2020-07-09
[Tracking] Investigate/Fix LG test timeouts


mdrichardson:
Many of the LG tests frequently timeout. Investigate possible solutions to fix them.

[Analytics](https://fuselabs.visualstudio.com/SDK_v4/_pipeline/analytics/stageawareoutcome?definitionId=1008)

and [example failure](https://fuselabs.visualstudio.com/SDK_v4/_build/results?buildId=144952&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=d38544d5-e3b6-5628-9a8d-57f5f7ecd74f&l=3509)

mdrichardson:
So, I only have one run each to compare to, buuuuut:

* As-is LG tests took 2 seconds and maybe half show "slow" (>40ms?, anything under that doesn't show time taken): https://fuselabs.visualstudio.com/SDK_v4/_build/results?buildId=145436&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=d38544d5-e3b6-5628-9a8d-57f5f7ecd74f&l=249
* Making it use an async fs.readFile took 3 seconds and about half show "slow": https://fuselabs.visualstudio.com/SDK_v4/_build/results?buildId=145414&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=d38544d5-e3b6-5628-9a8d-57f5f7ecd74f&l=249
* Pre-loading the files took 541ms and like 10% show "slow" https://fuselabs.visualstudio.com/SDK_v4/_build/results?buildId=145479&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=d38544d5-e3b6-5628-9a8d-57f5f7ecd74f&l=249

I'll get these fixed up early next week.

PR #2588, opened 2020-07-27
Fix intermittent expressions test error due to hour change

mdrichardson:
## Description

Occasionally, one of the `adaptive-expressions` tests would fail with something like:

```cmd
 AssertionError [ERR_ASSERTION]: actual is: 07-27-20 17, expected is 07-27-20 16
      + expected - actual

      -07-27-20 17
      +07-27-20 16
```

I believe this is because:

* Test loads at say, 9:58am which [would give expected data of `07-27-20 09'](https://github.com/microsoft/botbuilder-js/blob/master/libraries/adaptive-expressions/tests/expressionParser.test.js#L467)
* Test actually runs two minutes later at 10:00, so the actual data would be `07-27-20 10'. 

## Specific Changes

Have the test refresh the expected time based off of the current time.

### DIFF
--- a/libraries/adaptive-expressions/tests/expressionParser.test.js
+++ b/libraries/adaptive-expressions/tests/expressionParser.test.js
@@ -464,7 +464,7 @@ const dataSource = [
     ['date(timestamp)', '3/15/2018'],//Default. TODO
     ['year(timestamp)', 2018],
     ['length(utcNow())', 24],
-    ['utcNow(\'MM-DD-YY HH\')', moment(new Date().toISOString()).utc().format('MM-DD-YY HH')],
+    ['utcNow(\'MM-DD-YY HH\')', 'getNowTime'],
     ['formatDateTime(notISOTimestamp)', '2018-03-15T13:00:00.000Z'],
     ['formatDateTime(notISOTimestamp, \'MM-dd-yy\')', '03-15-18'],
     ['formatDateTime(notISOTimestamp, \'ddd\')', 'Thu'],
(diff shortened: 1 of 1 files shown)


libraries/adaptive-expressions/tests/expressionParser.test.js
```diff
@@ -464,7 +464,7 @@ const dataSource = [
     ['date(timestamp)', '3/15/2018'],//Default. TODO
     ['year(timestamp)', 2018],
     ['length(utcNow())', 24],
-    ['utcNow(\'MM-DD-YY HH\')', moment(new Date().toISOString()).utc().format('MM-DD-YY HH')],
+    ['utcNow(\'MM-DD-YY HH\')', 'getNowTime'],
```
stevengum, line 467: Is this a case of "two new date objects that are supposed to have the same time but since they were instantiated at different times the test intermittently fails?"

stevengum: Because the tests in this file take a long time to run?

mdrichardson: I'm still not 100% on this, but:

> Is this a case of "two new date objects that are supposed to have the same time but since they were instantiated at different times the test intermittently fails?"

I believe so.

> Because the tests in this file take a long time to run?

[This is my pipeline test that broke](https://fuselabs.visualstudio.com/SDK_v4/_build/results?buildId=150473&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=d38544d5-e3b6-5628-9a8d-57f5f7ecd74f&l=3408). If you look at the [raw output](https://fuselabs.visualstudio.com/86659c66-c9df-418a-a371-7de7aed35064/_apis/build/builds/150473/logs/29), all tests began at 16:58, but this particular test file doesn't get hit until 17:01. So, I _think_ that mocha runs all of the test side effects before actually running the tests.

stevengum:
Another test related to https://github.com/microsoft/botbuilder-js/issues/2512.

PR #2596, opened 2020-07-28
Fix intermittent test error due to time change

mdrichardson:
Related: https://github.com/microsoft/botbuilder-js/pull/2588

## Description

PR linked above fixes when `new Date()` is called at the end of an hour for hour-sensitive tests. However, it was still possible to error, even though it was unlikely.

Instead, this PR uses `sinon` to freeze the system clock so `new Date()` always returns the same time.

## Notes

We discussed using `getPastTime()` and `getFutureTime()` to be used together to validate against an expected constant. For example, pseudo-code would be something like:

> assert `getPastTime(1day)` equals `getNow()` minus one day

However, `getPastTime()` and all other relevant expression functions, make their own `new Date()` calls, so we're still running into the same problem. Stubbing `Date` with `sinon` seemed the be the best solution.


libraries/adaptive-expressions/tests/expressionParser.test.js
```diff
@@ -837,35 +833,49 @@ const scope = {
...
+            ['utcNow(\'MM-DD-YY HH\')', moment(new Date().toISOString()).utc().format('MM-DD-YY HH')],
+            ['getPastTime(1, \'Year\', \'MM-dd-yy\')', moment(new Date().toISOString()).utc().subtract(1, 'years').format('MM-DD-YY')],
+            ['getPastTime(1, \'Month\', \'MM-dd-yy\')', moment(new Date().toISOString()).utc().subtract(1, 'months').format('MM-DD-YY')],
+            ['getPastTime(1, \'Week\', \'MM-dd-yy\')', moment(new Date().toISOString()).utc().subtract(7, 'days').format('MM-DD-YY')],
+            ['getPastTime(1, \'Day\', \'MM-dd-yy\')', moment(new Date().toISOString()).utc().subtract(1, 'days').format('MM-DD-YY')],
+            ['getFutureTime(1, \'Year\', \'MM-dd-yy\')', moment(new Date().toISOString()).utc().add(1, 'years').format('MM-DD-YY')],
+            ['getFutureTime(1, \'Month\', \'MM-dd-yy\')', moment(new Date().toISOString()).utc().add(1, 'months').format('MM-DD-YY')],
+            ['getFutureTime(1, \'Week\', \'MM-dd-yy\')', moment(new Date().toISOString()).utc().add(7, 'days').format('MM-DD-YY')],
+            ['getFutureTime(1, \'Day\', \'MM-dd-yy\')', moment(new Date().toISOString()).utc().add(1, 'days').format('MM-DD-YY')],
```
stevengum, line 875: Recommendation is to change these tests to instead compare these `'getPastTime()'` and `'getFutureTime()'` to constant times as opposed to creating a new Date each test run.

libraries/adaptive-expressions/tests/expressionParser.test.js
```diff
@@ -837,35 +833,49 @@ const scope = {
...
-            const newActual = newExpr.tryEvaluate(scope).value;
-            assertObjectEquals(actual, newActual);
         }
+
+        // Any tests related to getting the current time should be instantiated in the actual test.
+        // If `new Date()` is called near the end of a time change (end of an hour or end of a day),
+        // this can intermittently result in failures because the test may run at the start of an hour or day.
+        iterateAndCheckData(dataSource);
+        const timeSensitiveData = [
+            ['utcNow(\'MM-DD-YY HH\')', moment(new Date().toISOString()).utc().format('MM-DD-YY HH')],
```
stevengum, line 867: This change should be _mostly_ okay, as this expression is evaluated immediately after the creation...

libraries/adaptive-expressions/package.json
```diff
@@ -31,7 +31,8 @@
     "lodash": "^4.17.19",
     "lru-cache": "^5.1.1",
     "moment": "^2.25.1",
-    "moment-timezone": "^0.5.28"
+    "moment-timezone": "^0.5.28",
+    "sinon": "^7.5.0"
```
stevengum, line 35: These are devDependencies 😊

johnataylor:
This looks much better to me. Thanks for the extra effort Michael.

One thought, I wonder whether actually puting the constant string in the test case as the expected value would be better. I wonder whether that would be more explicit and more readable - thought that's pretty subjective. Anyhow, the logic is correct.

So do you have to restore the clock after the end of the test?

mdrichardson:
@johnataylor I like the idea of using constant strings as `expected` instead of more and more calls to `moment(new Date...`. I'll change that.

Yes, I believe the clock needs to be restored at the end of the test. I have it, but I believe it needs to be in a `try/finally`

Issue #2338, opened 2020-06-12
[Engineering] Lower time needed to complete running tests


stevengum:
Improve test duration by:

- Utilizing addons to the mocha test framework such as mocha-parallel-tests
- Removing duplicate tests
- Reusing previous tests' arrangements

The improvements in test duration should not be at the cost of reliability and code coverage.

PR #2611, opened 2020-07-29
[Engineering] Lower time for botbuilder-azure tests

denscollo:
Addresses #2338
## Description
Updates the warnings raised in some tests of the **BotBuilder-Azure** library to be shown once for the describe instead of each test inside it.

## Specific Changes
Update the tests in the files `blobStorage.test.js` and `cosmosDbStorage.test.js` to run the emulator check once in the describes to raise a warning instead of running it in each test of the describe. This change reduces the time for the tests a little bit and makes the test report more readable.

We also updated some eslint warnings in these files.
## Testing
Here you can compare the times for each test running when moving the warning to be shown once for the describe and how it is more readable in the report.
image
image
image

**Note:** the times vary in each run, but in general there is a slight reduction in the times.


stevengum:
@denscollo there are no changes in this PR?

denscollo:
@stevengum Hey Steven, sorry for this. We created a lot of PRs with a similar name (due to the library name), and I typed wrong the branch name. I'll close this and create a new one with the correct branch ASAP.

denscollo:
@stevengum created PR with the correct branch on #2629

PR #2661, opened 2020-08-06
[Port]Make Language fallback consistent with current LanguagePolicy

### LINKED ISSUE #2639: [PORT] Make Language fallback consistent with current LanguagePolicy
https://github.com/microsoft/botbuilder-dotnet/pull/4345

Fixes #4309

## Description
Original language policy:
en-us -> en-us, en, ''
fr-fr -> fr-fr, fr, ''
'' -> ''

Current default language policy:
en-us -> en-us, en
fr-fr -> fr-fr, fr
'' -> ''

So, it is necessary to update the `GroupByLocale` in `LGResourceLoader` to apply the lastest language policy.

## Specific Changes
Append all other resources those locale not match current entry locale to current locale.
For example：
there exist `a.en-us.lg`, `a.lg`, `b.lg`
At first, we should group all LG resources to make them locale awareness.
Suppose there exists the language policy:
en-us: en-us, en
'' : ''

The final grouped resources should like:
en-us -> a.en-us.lg, b.lg
'' -> a.lg, b.lg

# Changed projects

Danieladu:
Fixes #2639

## Description
Enable LG debuggability with event handler.
1. Add LG event and Expression event args in LG sdk
2. Register LG/Expression object and source range
3. Emit event during LG evaluation and Expression evaluation
4. Handler event in the adaptive language generator

## How to detect the debugger
1. Set up an Adaptive dialog bot, and startup the project
2. Install adaptive-tool extension for vscode. (Currently, adaptive-tool extension could be built from [source code](https://github.com/microsoft/BotBuilder-Samples/tree/master/experimental/adaptive-tool))
3. Goto the LG file that the adaptive-bot used, and set up the [launch file](https://github.com/microsoft/BotBuilder-Samples/tree/master/experimental/adaptive-tool#setting-up), press F5 to listen on the bot service runtime
4. Insert breaking points for Lg templates or Expressions
5. Go through the bot flow, if the bot has reached a certain template or a certain expression, the code will stop at the breakpoint.
Check the variables and status to get more details.

## Details for this PR
- Add four event args class: LGEventArgs, MessageArgs, BeginTemplateEvaluationArgs, and BeginExpressionEvaluationArgs.
They all inherit from LGEvent which is registered in dialogEvents.
- Add ExpressionRef class as the breakpoint matching object, since the debugger uses "reference equal" as the breaking point equivalence.
- Register the source map of templates and expressions during the paring stage.
- Add the event handler at the Adaptive generator layer. When there is LG/Expression hitting event, check the debugger to see if there is a breaking point at the current line. 
- Emit LG event when the code is hitting a certain template during the evaluation stage, and also emit the expression event when the code is evaluating a certain expression.
- Adaptive dialog would subscribe to these events and emit [dap](https://microsoft.github.io/debug-adapter-protocol/) messages to vscode.


libraries/botbuilder-dialogs-adaptive/src/languageResourceLoader.ts
```diff
@@ -30,34 +30,32 @@ export class LanguageResourceLoader {
...
-                            existNames.add(prefixName);
-                            if (!resourceMapping.has(locale)) {
-                                resourceMapping.set(locale, [u]);
-                            } else {
-                                resourceMapping.get(locale).push(u);
-                            }
+                const resourcesWithSuffix = allResouces.filter((u): boolean => this.parseLGFileName(u.id).language.toLocaleLowerCase() === suffix.toLocaleLowerCase());
+                resourcesWithSuffix.forEach((u): void => {
+                    const resourceName = u.id;
+                    const length = (!suffix)? 3 : 4;
```
chrimc62, line 36: >)? [](start = 43, length = 2)

space

Danieladu: fix. Just relying on the length of LG extension name("lg")

libraries/botbuilder-dialogs-adaptive/src/languageResourceLoader.ts
```diff
@@ -17,47 +17,49 @@ import { LanguagePolicy } from  './languagePolicy';
  * Load all LG resource and split them into different language groups.
  */
 export class LanguageResourceLoader {
+    private static readonly lgSuffix: string = 'lg';
```
chrimc62, line 20: >  private static readonly lgSuffix: string = 'lg'; [](start = 2, length = 50)

This is C#, not typescript right?  Does this actually compile?

PR #2666, opened 2020-08-06
Refactor SkillDialog to call ConversationFactory.createConversationId only once

### LINKED ISSUE #2653: [PORT] Refactored SkillDialog to call ConversationFacotry.CreateConversationId only once
https://github.com/microsoft/botbuilder-dotnet/pull/4382

Refactored SkillDialog to call the factory only during begin and store the generated conversationId in state to be used in subsequent calls to the skill.

Needs parity for python and JS

Fixes #4374

# Changed projects

mdrichardson:
Fixes #2653 

## Description

Refactored SkillDialog to call the factory only during begin and store the generated conversationId in state to be used in subsequent calls to the skill.


libraries/botbuilder-dialogs/tests/skillDialog.test.js
```diff
@@ -282,23 +289,31 @@ describe('SkillDialog', function() {
...
             context.turnState.set(adapter.OAuthScopeKey, DEFAULT_OAUTHSCOPE);
             const dialog = new SkillDialog({
                 botId: 'botId',
                 conversationIdFactory: { 
                     createSkillConversationIdWithOptions: async () => { throw new Error('Not Implemented'); },
-                    createSkillConversationId: async () => done()
+                    createSkillConversationId: async () => {}
                 },
                 conversationState: {
                     saveChanges: () => null
```
mdrichardson, line 319: These two tests were forced to change since `sendToSkill()` no longer calls `createSkillConversationId()`, which is what this tests. `beginDialog()` does call this and in order to switch to that, a lot of dialog state things needed to be created.

libraries/botbuilder-dialogs/tests/skillDialog.test.js
```diff
@@ -91,6 +91,8 @@ describe('SkillDialog', function() {
                 activityToSend.text = activityToSendText;
                 const client = new DialogTestClient(Channels.Test, dialog, { activity: activityToSend }, undefined, conversationState);
 
+                strictEqual(0, dialogOptions.conversationIdFactory.createCount);
```
stevengum, line 94:
```suggestion
                strictEqual(dialogOptions.conversationIdFactory.createCount, 0);
```

The value being tested is the first value in [`assert.strictEqual()`](https://nodejs.org/api/assert.html#assert_assert_strictequal_actual_expected_message)

PR #2697, opened 2020-08-13
Appropriately remove test env vars

### LINKED ISSUE #2696: Botbuilder\streaming tests fail when run alone
Lots of investigation needed, but if you run:

* `npm run test`, tests pass
* `mocha --recursive libraries/botbuilder/tests`, tests pass
* `mocha libraries/botbuilder/tests/streaming`, one test fails with:

image

Bizarrely, if you delete tests for **both** botFrameworkHttpClient AND channelServiceHandler, the test "appropriately" fails. If **either** file is present, the test passes.

Note: I say "appropriately" because the test fails when run alone, which is when I think it *should* pass, if the test were written to pass, which I don't believe it is.

Investigating...

mdrichardson:
Fixes #2696 

## Description

There was some weird behavior when running all tests (tests pass) vs only running tests in `botbuilder\tests\streaming` (one test fails) due to not appropriately deleting environment variables.

## Specific Changes

Change test cleanup from `<envVar> = undefined` to `delete <envVar>`.

It needs to be deleted because if you set it as null or undefined, the key still exists, but as the string `'null'` or `'undefined'`, respectively.

## Testing

I had to change one test to match the `err.message`. Added comment about it in "Files changed".

## Note

I considered looking at the SDK source code to add a check for `'undefined'` vs `undefined` in the environment variable, but decided against it because the user should really delete it instead.

I also searched to see if this occurred in other tests and this was the only place I found anything like this. Other tests that set env vars later delete them appropriately with `delete`.


libraries/botbuilder/tests/streaming/botFrameworkAdapterStreaming.test.js
```diff
@@ -124,7 +124,7 @@ describe('BotFrameworkAdapter Streaming tests', () => {
                 });
             } catch (err) {
                 expect(err.statusCode).to.equal(StatusCodes.UNAUTHORIZED);
-                expect(err.message).to.equal('Unauthorized. No valid identity.');
+                expect(err.message).to.equal('Unauthorized. Is not authenticated');
```
mdrichardson, line 127: Prior to this change, this test passes when test are run from `root` to `botbuilder\tests`. If they were run from `botbuilder\tests\streaming`, this test would fail because, when run solo, it lacked the environment variable for the ChannelService. When the tests above ran, they set the ChannelService to `'undefined'` which, because it is a string, is not falsy.

Therefore, I **believe** this change is appropriate, but am willing to concede if somebody has a strong argument, otherwise.

Zerryth: I agree with Michael. His change should be the correct error message, because the the test uses `'donttrustme'`, which eventually becomes `identity` once in the `*ChannelValidation` layer of authentication.

```ts

        if (!identity) {
            // No valid identity. Not Authorized.
            throw new AuthenticationError('Unauthorized. No valid identity.', StatusCodes.UNAUTHORIZED);
        }

        if (!identity.isAuthenticated) {
            // The token is in some way invalid. Not Authorized.
            throw new AuthenticationError('Unauthorized. Is not authenticated', StatusCodes.UNAUTHORIZED);
        }
```

`identity` *is* provided, it's just not validated to be an authenticated token, therefore should throw `Unauthorized. Is not authenticated` as the err message

stevengum:
:shipit:

PR #2698, opened 2020-08-13
Plug mocha test leaks

mdrichardson:
Addresses Point 1 of #2695 

## Description

I ran mocha tests with the `--check-leaks` flag to check for global variable scoping that can/does potentially cause tests to pollute each other.

The output gives something like:

```
1) expression parser functional test
       Test AccumulatePath:
     Error: global leaks detected: 'left', 'error'
      <stack trace>

  2) LuisRecognizer
       should call prepareErrorMessage when a non-200 status code is received.:
     Error: global leak detected: 'expectedError'
      <stack trace>

  3) LuisRecognizer
       null telemetryClient should work.:
     Error: global leaks detected: 'includeAllIntents', 'includeInstance', 'telemetryClient', 'telemetryProperties', 'logPersonalInformation'
      <stack trace>

  <etc>
```

This is caused mainly by variables being declared like:

`supposedToBeLocal = true`

instead of:

`let supposedToBeLocal = true`.

Not putting `let/var/const` adds it to the global scope like, `global.supposedToBeLocal`.

## Specific Changes

- Properly scope test variables
- Add `--check-leaks` flag to `test` (usually run for local dev) and `test-coverage` (run in CI pipeline) scripts.
- Some minor no-op changes as I came across them


libraries/adaptive-expressions/tests/expressionParser.test.js
```diff
@@ -948,22 +948,22 @@ describe('expression parser functional test', () => {
         // normal case, note, we doesn't append a " yet
         let exp = Expression.parse('a[f].b[n].z');
         let path = undefined;
-        ({path, left, error} = FunctionUtils.tryAccumulatePath(exp, memory, undefined));
```
mdrichardson, line 951: `left` and `error` are declared globally here, but unused, so I just deleted them.

libraries/botbuilder-applicationinsights/tests/telemetryInitializer.test.js
```diff
@@ -27,7 +27,6 @@ describe(`TelemetryInitializerMiddleware`, function () {
         var initializerMiddleware = new TestInitializerMiddleware(telemetryLoggerMiddleware, true, []);
         
         var adapter = new TestAdapter(async (context) => {
-            conversationId = context.activity.conversation.id;
```
mdrichardson, line 30: declared globally but not used

libraries/botbuilder-core/tests/telemetryMiddleware.test.js
```diff
@@ -3,7 +3,6 @@
 
 const assert = require('assert');
 const { TestAdapter, TelemetryLoggerMiddleware, ActivityTypes } = require('../');
-const { TeamsChannelData } = require('../../botframework-schema/');
```
mdrichardson, line 6: unused no-op

libraries/botbuilder/tests/inspectionMiddleware.test.js
```diff
@@ -10,18 +10,18 @@ const { TestAdapter, MemoryStorage, MessageFactory, UserState, ConversationState
 const { InspectionMiddleware, InspectionState } = require('../');
 const sinon = require('sinon');
 
-beforeEach(function(done) {
```
mdrichardson, line 13: putting these outside of `describe` adds them to the global mocha scope

package.json
```diff
@@ -7,11 +7,11 @@
     "clean": "lerna run clean",
     "functional-test": "lerna run build && nyc mocha \"libraries/functional-tests/tests/*.test.js\"",
     "browser-functional-test": "cd libraries/browser-functional-tests && node nightwatch.js -e",
-    "test": "lerna run build && nyc mocha \"libraries/@(adaptive*|bot*)/tests/**/*.test.js\" --exit",
+    "test": "lerna run build && nyc mocha \"libraries/@(adaptive*|bot*)/tests/**/*.test.js\" --exit --check-leaks",
```
mdrichardson, line 10: I think it would be good to have the `--check-leaks` flag in both `test` and `test-coverage` to prevent this in the future, but I'm willing to remove them if reviewers disagree. If a leak is detected, it makes the test fail. As far as I can tell, leaks are detected deterministically.

stevengum: I think this is worth trying out. If it causes pain to the dev team we can evaluate an alternative path for keeping track of test leaks.

stevengum (approved):
LGTM

stevengum:
@garypretty or @munozemilio can you review this PR? Thanks!

PR #2700, opened 2020-08-13
Make telemetryMiddleware tests actually fail (instead of just logging the error and passing)

mdrichardson:
## Description

I noticed when running tests that one of the TelemetryLoggerMiddleware tests fails ([noted here](https://github.com/microsoft/botbuilder-js/pull/2700/files#r470246739)), but it isn't caught by mocha. Instead, the failed assertion gets printed to the console and the test passes.

## Specific Changes
  - Move `done()` and `done(err)` calls into the `telemetryClient`'s `trackEvent` method.
  - A bunch of no-op stuff that was bothering me (sorry)

## Testing

In addition to running the test, I inserted some `assert(false)`'s to ensure that the tests would actually fail now, and they do.


libraries/botbuilder-core/tests/telemetryMiddleware.test.js
```diff
@@ -3,97 +3,164 @@
 
 const assert = require('assert');
 const { TestAdapter, TelemetryLoggerMiddleware, ActivityTypes } = require('../');
-const { TeamsChannelData } = require('../../botframework-schema/');
 
-describe(`TelemetryMiddleware`, function () {
+class OverrideReceiveLogger extends TelemetryLoggerMiddleware {
```
mdrichardson, line 7: green section is all no-op. Moved from bottom of file so get rid of "xyz is called before it's defined error"

libraries/botbuilder-core/tests/telemetryMiddleware.test.js
```diff
@@ -3,97 +3,164 @@
...
+}
+
+describe(`TelemetryMiddleware`, function() {
     this.timeout(5000);
-    it(`telemetry should log send and receive activities`, function (done) {
+    it(`telemetry should log send and receive activities`, function(done) {
         var callCount = 0;
         var telemetryClient = {
             trackEvent: (telemetry) => {
-                assert(telemetry, 'telemetry is null');
```
mdrichardson, line 14: These changes look big, but they're just wrapping all the assertions in a `try/catch` so that:

```js
try {
  // assertions
  done();
} catch(err) {
  done(err); // this is the only way I could get the test to fail
}
```

libraries/botbuilder-core/tests/telemetryMiddleware.test.js
```diff
@@ -3,97 +3,164 @@
...
+                            break;
+                    }
+                } catch (err) {
+                    done(err);
                 }
             }
-        }
+        };
         let myLogger = new TelemetryLoggerMiddleware(telemetryClient, true);
-        var conversationId = null;
```
mdrichardson, line 88: no-op. unused vars

libraries/botbuilder-core/tests/telemetryMiddleware.test.js
```diff
@@ -102,22 +169,18 @@ describe(`TelemetryMiddleware`, function () {
             .assertReply('echo:foo')
             .send('bar')
             .assertReply(activity => assert.equal(activity.type, ActivityTypes.Typing))
-            .assertReply('echo:bar')
-            .then(done);
```
mdrichardson, line 106: as part of the `try/catch` change above, I moved all of the `done()` calls into the `try/catch` wrapper

libraries/botbuilder-core/tests/telemetryMiddleware.test.js
```diff
@@ -559,76 +611,7 @@ describe(`TelemetryMiddleware`, function () {
...
         };
 
-        adapter
-            .send(activity)
-            .then(done);
+        adapter.send(activity);
     });
 
 });
-
```
mdrichardson, line 568: moved to top

libraries/botbuilder-core/tests/telemetryMiddleware.test.js
```diff
@@ -511,32 +559,36 @@ describe(`TelemetryMiddleware`, function () {
             .send('foo')
             .delay(100)
             .send('update')
-            .delay(100)
-            .then(done());
+            .delay(100);
     });
 
-    it(`telemetry should log channel specific properties`, function (done) {
+    it(`telemetry should log channel specific properties`, function(done) {
```
mdrichardson, line 565: This is the test that was actually failing, uncaught. The error was that it hit the `assert(false)` on removed line 536. This is because it gets called a 2nd time. I didn't account for the second call because, as far as I can tell, we're only trying to verify the assert information in `case 1:`

stevengum (approved):
LGTM

Edit: Thanks for providing the comments. It helped 😄

PR #2713, opened 2020-08-18
Increase AppInsights Coverage

### LINKED ISSUE #2433: Increase coverage for botbuilder-applicationinsights library
Increase unit test coverage of botbuilder-applicationinsights library to 80%

mdrichardson:
Fixes #2433 

## Description

I can see why coverage was so low; some of these tests are quite a stretch for usefulness. Maybe I just got carried away:

image

**I'm very willing to remove some tests that might seem unnecessary**, especially those from [this commit (100 percent statement, branch, function coverage)](https://github.com/microsoft/botbuilder-js/pull/2713/commits/382b8c41d0b57b6679a8d01ae48d4c5ba3782fe7).

## Specific Changes
  - Increase test coverage to 100% for `botbuilder-applicationinsights`
  - Some no-op changes


libraries/botbuilder-applicationinsights/tests/TelemetryWaterfallTest.test.js
```diff
@@ -1,31 +1,24 @@
-const { BotTelemetryClient, ConversationState, MemoryStorage, NullTelemetryClient, TestAdapter } = require('botbuilder-core');
```
mdrichardson, line 1: All changes in this file are no-op to quiet my linter

libraries/botbuilder-applicationinsights/tests/telemetryInitializer.test.js
```diff
@@ -4,6 +4,9 @@
 const assert = require('assert');
```
mdrichardson, line 4: New tests are at the end (line 123). Everything else is no-op, mostly changing `var` to `const` because `var` hoisting isn't intuitive.

libraries/botbuilder-applicationinsights/tests/telemetryInitializer.test.js
```diff
@@ -113,4 +120,125 @@ describe(`TelemetryInitializerMiddleware`, function () {
...
+        const logger = new TelemetryLoggerMiddleware(new TelemetryClient('fakeKey'));
+        let passedInFunctionWasCalled = false;
+        const middleware = new TelemetryInitializerMiddleware(logger, () => {
+            passedInFunctionWasCalled = true;
+            return true;
+        });
+
+        const result = middleware.logActivityTelemetry();
+        assert.strictEqual(result, true);
+        assert.strictEqual(passedInFunctionWasCalled, true);
```
stevengum, line 134: The expected second value for TelemetryInitializerMiddleware's constructor is a `boolean`, not `() => boolean`. ([link](https://github.com/microsoft/botbuilder-js/blob/ca9de94139f3aa57ff113eb5c87eb62de7873d54/libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts#L27))

The [`logActivityTelemetry` getter](https://github.com/microsoft/botbuilder-js/blob/ca9de94139f3aa57ff113eb5c87eb62de7873d54/libraries/botbuilder-applicationinsights/src/telemetryInitializerMiddleware.ts#L35) simply returns the `private readonly _logActivityTelemetry`.

This test has strayed into testing the language 😄 

Was there a reason you decided to test with a function?

```suggestion
        const middleware = new TelemetryInitializerMiddleware(logger, true);

        assert.strictEqual(middleware.logActivityTelemetry, true);
```

stevengum (changes requested):
One small bit of feedback, otherwise it looks good @mdrichardson. Thanks for taking increasing the code coverage!

libraries/botbuilder-applicationinsights/tests/telemetryClient.test.js
```diff
@@ -0,0 +1,238 @@
...
+            assert.strictEqual(envelope.data.baseData.properties.conversationId, activity.conversation.id);
+        });
+
+        it('addBotIdentifiers() correctly sets App Insights values, defaulting to empty strings', function() {
+            const client = new ApplicationInsightsTelemetryClient('fakeKey');
+            client.client.commonProperties = undefined;
+            const activity = {
+                from: undefined,
+                channelId: undefined,
+                conversation: undefined,
```
stevengum, line 153: Shouldn't these be unnecessary? Or is something doing an Object.keys()/hasOwnProperty?

mdrichardson: It's unnecessary. Since that test tests that those properties become empty string, I was just stating explicitly that the activity starts with those undefined. But yeah, functionally it's unnecessary, so I'll remove it.

libraries/botbuilder-applicationinsights/tests/telemetryInitializer.test.js
mdrichardson: This is one of those tests that I only really added for the 100% code coverage. Since `_logActivityTelemetry` can only be set by passing it into the constructor, there's no way of asserting that it is a particular value before it's instantiated. Meaning, I can't:

```js
assert.strictEqual(middleware.logActivityTelemetry, undefined);

joshgummersall (commented):
I'll defer to @stevengum for final approval. Looks good to me though.

libraries/botbuilder-applicationinsights/tests/telemetryInitializer.test.js
stevengum: That code coverage mechanism is unpleasant, thanks for providing context!

stevengum (approved):
:shipit:

munozemilio (approved):

PR #2720, opened 2020-08-19
Use resource id as dialog id only if the dialog doesn't have an id

### LINKED ISSUE #2579: [PORT] Only use resource.id as the default dialog id if the dialog doesn't have an id.
https://github.com/microsoft/botbuilder-dotnet/pull/4314

Fixes #4250 

## Description
only apply resource.id as default .id if there is no id in the incoming json payload.

## Specific Changes
Change conditional for auto assigning dialog.id

## Testing
added unit test to verify.

# Changed projects
*

chon219:
Fixes #2579 

## Description

If a dialog does not have an id in JSON file, resource id would be used as dialog id.

## Specific Changes

  - Updated loadType in resource explorer to assign dialog id.
  -
  -

## Testing

Updated test case to verify this feature.


stevengum:
https://github.com/microsoft/botbuilder-js/pull/2742 will fix the CI issues related to the release of TypeScript 4.x

PR #2744, opened 2020-08-25
Re-structure templates crud

### LINKED ISSUE #2679: [Composer] lg templates CRUD functions is not friendly to composer
In composer,  we use web worker to do lg full parse. But the parse result object contains template CRUD methods. We can't pass them from the worker thread  to main thread. If we want to get benefits from the template CRUD methods, we have to do a full parse in main thread and this may take a long time.

1. parse result is a pure object without any functions.
2. provide extra template CRUD utility methods

Danieladu:
Fixes #2679

## Description
Redo the template CRUD to meet the requirements from the Composer side.

The changes made:
1. Extract CRUD operators into a single file, and export functions directly
2. CRUD operators would not change the original templates
3. Add `reParse` optional parameter, if `reParse` is set to False, a mock parse would replace the real parse, and 
  the diagnostics would have no effect. Also, there is no `templateBodyParseTree` in the template result.


Danieladu:
Close it due to the Change of Composer side.

PR #2748, opened 2020-08-27
[Port] Fix error in parsing empty string in string interpolation

ShYuPe:
Port C# PR: https://github.com/microsoft/botbuilder-dotnet/pull/4504

In adaptive expression, currently, 
```JavaScript
var parsed = Expression.parse("``");
```
will throw an error.
In this PR, the parser rule is changed so that, empty string in string interpolation can be correctly processed.


package.json
```diff
@@ -1,56 +1,56 @@
 {
```
Danieladu, line 1: it is better to keep the original format

libraries/adaptive-expressions/package.json
```diff
@@ -24,7 +24,7 @@
     "@types/lru-cache": "^5.1.0",
     "@types/moment-timezone": "^0.5.13",
     "@types/xmldom": "^0.1.29",
-    "antlr4ts": "0.5.0-alpha.3",
```
Danieladu, line 27: it  is better to fix the version of antlr, because every small change is breaking change for the Lexer/Parser

libraries/botbuilder-lg/tests/lg.test.js
```diff
@@ -699,6 +699,9 @@ describe('LG', function() {
         var evaled = templates.evaluateText('hello');
         assert.strictEqual('hello', evaled);
 
+        evaled = templates.evaluateText("${''}");
```
cosmicshuai, line 702: evaled = templates.evaluateText("${``}"); 
should evalute an empty string interpolation

ShYuPe: fixed

package.json
ShYuPe: fixed

libraries/adaptive-expressions/package.json
ShYuPe: ok，fixed

cosmicshuai (approved):
LGTM.

boydc2014 (approved):

PR #2756, opened 2020-09-01
[Port] Support create object with multi lines in LG

cosmicshuai:
With the support of multi lines object creation, 
```
#ctobj
- ${{a: 1, c
:3,
b:2
}}
```
will return `{a:1, c:3, b: 2}`

```
#evalMultiLineObj
- ${string(
setProperty(
{a: 1, 
b: 2},
'c', 
{d: 4, 
e: 5})
)}
```
will return`" {'a':1,'b':2,'c':{'d':4,'e':5}}"`


chrimc62:
> 		"\x1Ck\x02\x1Dm\x02\x1Eo\x02\x1Fq\x02 s\x02!u\x02\"w\x02#\x07\x02\x03\x04" +

Should this file be check-in since it is generated?

---
Refers to: libraries/botbuilder-lg/src/generated/LGTemplateLexer.ts:611 in 9fc4bc5. [](commit_id = 9fc4bc5b9daed1b2b3f1fb929a47dbc38ac358f7, deletion_comment = False)

boydc2014:
> > ```
> >   "\x1Ck\x02\x1Dm\x02\x1Eo\x02\x1Fq\x02 s\x02!u\x02\"w\x02#\x07\x02\x03\x04" +
> > ```
> 
> Should this file be check-in since it is generated?
> 
> Refers to: libraries/botbuilder-lg/src/generated/LGTemplateLexer.ts:611 in [9fc4bc5](https://github.com/microsoft/botbuilder-js/commit/9fc4bc5b9daed1b2b3f1fb929a47dbc38ac358f7). [](commit_id = [9fc4bc5](https://github.com/microsoft/botbuilder-js/commit/9fc4bc5b9daed1b2b3f1fb929a47dbc38ac358f7), deletion_comment = False)

In C# we don't have to check in this file, because vs support an explicit step of "code generation", but in js, we don't have the native support for this, so it's recommended to check in this. We placed it under a "generated" folder to distinguish with other files.

PR #2760, opened 2020-09-03
SwitchCondition with number throws an exception when compared against a string

### LINKED ISSUE #2467: [PORT] SwitchCondition with number throws an exception when compared against a string
https://github.com/microsoft/botbuilder-dotnet/pull/4148

Fixes #4071 

## Description
SwitchCondition treated errors as exceptions when all that was happening was an comparison which was false. In addition, it was always assuming that a quoted number should be compared as a number. 

## Specific Changes
* Fix switch to not treat expression errors as exceptions but instead as false condition
* Fix switch to treat ambiguity in string case definition more robustly by comparing for both value type and string of value type. (23 OR "23")

## Testing
updated unit test cases to cover this.

# Changed projects

chon219:
Fixes #2467 

## Description

SwitchCondition treated errors as exceptions when all that was happening was an comparison which was false. In addition, it was always assuming that a quoted number should be compared as a number.

## Specific Changes

  - Fix switch to not treat expression errors as exceptions but instead as false condition
  - Fix switch to treat ambiguity in string case definition more robustly by comparing for both value type and string of value type. (23 OR "23")

## Testing

updated unit test cases to cover this.


libraries/botbuilder-dialogs-adaptive/src/actions/switchCondition.ts
```diff
@@ -66,8 +65,44 @@ export class SwitchCondition<O extends object = {}> extends Dialog<O> implements
...
+                }
+                if (caseScope.value === 'true' || caseScope.value === 'True') {
+                    // you don't have to put quotes around bools, "true" => true OR "true".
+                    this._caseExpresssions[caseScope.value] = Expression.orExpression(
+                        Expression.equalsExpression(this.condition, new Constant(true)),
+                        Expression.equalsExpression(this.condition, new Constant(caseScope.value))
+                    );
+                    continue;
+                }
+                if (caseScope.value === 'false' || caseScope.value === 'False') {
```
stevengum, line 94: This is odd code, should we just do `caseScope.value.toLowerCase() === 'true'`? Otherwise `'true'` and `'True'` are valid, but `'tRue'`, `'trUE'`, etc. are not. How strict should the SDK be on this user-provided value?

stevengum (changes requested):
Please see comment

chrimc62:
I'm confused.  The description only talks about switchCondition and there are a ton of other changes.  Either you should add them to the description or do a different PR.

build/AnalyzeDependencies/AnalyzeDepsJS.ps1
```diff
@@ -0,0 +1,221 @@
+#
+# Adapted from https://github.com/microsoft/botbuilder-dotnet/blob/main/build/AnalyzeDeps/AnalyzeDeps.ps1
+#
+
+<#
+  .SYNOPSIS
+  Generate a dependency report for a set of .tgz files
+  .PARAMETER PackagesPath
+  The path to the package(s) to analyze. Globs are supported.
```
chrimc62, line 9: Did you intend to include these changes?

libraries/adaptive-expressions/src/parser/expressionParser.ts
```diff
@@ -77,7 +77,9 @@ export class ExpressionParser implements ExpressionParserInterface {
                 result = new Constant(false);
             } else if (symbol === 'true') {
                 result = new Constant(true);
-            } else if (symbol === 'null' || symbol === 'undefined') {
+            } else if (symbol === 'null') {
+                result = new Constant(null);
+            } else if (symbol === 'undefined') {
```
chrimc62, line 82: Are you sure we should make this change?  This will break existing functions that check for null--now it will have to be changed to check for both null and undefined and I don't think that is a distinction we want to make.

libraries/botbuilder-dialogs-adaptive/etc/botbuilder-dialogs-adaptive.api.md
```diff
@@ -0,0 +1,1616 @@
...
+import { Activity } from 'botbuilder-core';
+import { ArrayExpression } from 'adaptive-expressions';
+import { BeginSkillDialogOptions } from 'botbuilder-dialogs';
+import { BoolExpression } from 'adaptive-expressions';
+import { BotFrameworkClient } from 'botbuilder-core';
+import { BotTelemetryClient } from 'botbuilder-core';
+import { BuilderRegistration } from 'botbuilder-dialogs-declarative';
+import { Choice } from 'botbuilder-dialogs';
+import { ChoiceFactoryOptions } from 'botbuilder-dialogs';
+import { ComponentRegistration } from 'botbuilder-dialogs-declarative';
```
chrimc62, line 16: This shouldn't be checked-in should it?

libraries/botbuilder-dialogs-adaptive/src/adaptiveDialog.ts
```diff
@@ -174,6 +174,12 @@ export class AdaptiveDialog<O extends object = {}> extends DialogContainer<O> {
         // Install dependencies on first access
         this.ensureDependenciesInstalled();
 
+        // Initialize dialog state
+        if (options) {
+            // Replace initial activeDialog.State with clone of options
+            dc.activeDialog.state = JSON.parse(JSON.stringify(options));
```
chrimc62, line 180: >JSON.parse(JSON.stringify(options)) [](start = 36, length = 35)

Is this how we do clone everywhere else in the SDK?

libraries/botbuilder-dialogs-adaptive/src/jsonExtensions.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+/**
+ * @module botbuilder-dialogs-adaptive
+ */
+
+import { DialogStateManager } from 'botbuilder-dialogs';
+import { ValueExpression } from 'adaptive-expressions';
+
+export function replaceJsonRecursively(state: DialogStateManager, unit: object): any {
+    if (typeof unit === 'string') {
```
chrimc62, line 13: This should be documented if exported--including why it is needed.

libraries/botbuilder-dialogs-adaptive/src/actions/baseInvokeDialog.ts
```diff
@@ -72,7 +73,12 @@ export class BaseInvokeDialog<O extends object = {}> extends Dialog<O> implement
 
         for (const key in bindingOptions) {
             const bindingValue = bindingOptions[key];
-            const value = new ValueExpression(bindingValue).getValue(dc.state);
+            let value = new ValueExpression(bindingValue).getValue(dc.state);
+
+            if (value) {
+                value = replaceJsonRecursively(dc.state, value);
+            }
```
chrimc62, line 80: This does not look safe to me at all--are you really sure that no child is a string that should not be replaced?

libraries/botbuilder-dialogs-adaptive/src/actions/switchCondition.ts
```diff
@@ -66,8 +65,44 @@ export class SwitchCondition<O extends object = {}> extends Dialog<O> implements
             this._caseExpresssions = {};
             for (let i = 0; i < this.cases.length; i++) {
                 const caseScope = this.cases[i];
-                const caseCondition = Expression.equalsExpression(this.condition, caseScope.createValueExpression());
-                this._caseExpresssions[caseScope.value] = caseCondition;
+                const intVal = parseInt(caseScope.value);
+                if (!isNaN(intVal)) {
+                    // you don't have to put quotes around numbers, "23" => 23 OR "23".
```
chrimc62, line 70: This whole thing is a bad idea.  I don't believe we should be loose in this way.  A number is a number and a string is a string.  What would I do if I wanted a case for each.

libraries/botbuilder-dialogs-declarative/etc/botbuilder-dialogs-declarative.api.md
```diff
@@ -0,0 +1,124 @@
...
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
+
+```ts
+
+import { FSWatcher } from 'chokidar';
+
+// @public (undocumented)
+export class BuilderRegistration {
+    constructor(name: string, builder: TypeBuilder);
+    // (undocumented)
```
chrimc62, line 12: Should this be checked in?

chrimc62:
Even worse--the issue being fixed should not be.  We do not let you add "21" and 2 why would we fold those together for cases?

---
In reply to: [697008094](https://github.com/microsoft/botbuilder-js/pull/2760#issuecomment-697008094) [](ancestors = 697008094)

chon219:
I would close this one and send out another.

Issue #2712, opened 2020-08-18
SkillConversationIdFactory memory leak on using DeliveryMode.ExpectReplies


alopix:
## Versions
What package version of the SDK are you using. 4.9.3
What nodejs version are you using 12.16.1
What browser version are you using –
What os are you using –

## Describe the bug
If the `DeliveryMode` is set to `ExpectReplies` when using the Skills SDK (not using `SkillHandler`), then the `SkillConversationIdFactory`'s `deleteConversationReference` will never be called. The more skills you call (which will respond directly in their response body), the mode conversation references are going to be saved and never cleaned up.
This will result in a potentially large memory leak when using in-memory storage for these references (as shown with the BotbuilderSamples) or a database that will fill up but also never be cleaned.

In our instance it was/is a memory leak.

## To Reproduce

1. Running the JS sample [81.skills-skilldialog](https://github.com/microsoft/BotBuilder-Samples/tree/master/samples/javascript_nodejs/81.skills-skilldialog)
2. Send an activity using `DeliveryMode` = `ExpectReplies` (which will NOT use `SkillHandler`), e.g. an `Invoke` activity.
3. Observe `SkillConversationIdFactory.deleteConversationReference` not ever being called

## Expected behavior
Add a call to `SkillConversationIdFactory.deleteConversationReference` when cleaning up these kinds of responses in e.g. after processing in [`SkillDialog.sendToSkill`](https://github.com/microsoft/botbuilder-js/blob/ad29ba286842813547093f000fed8699c6dd8ecf/libraries/botbuilder-dialogs/src/skillDialog.ts#L221).

## Screenshots
\-

## Additional context
\-

EricDahlvang:
Hi @alopix 

Thank you for raising this issue.  This does appear to be a gap in SkillDialog.  I'll discuss with the team how we want to address this, and get back to you.

gabog:
Thanks for the feedback @alopix, I created a root issue in our [SDK project](https://github.com/microsoft/botframework-sdk/issues/6019) and also created issues to solve this problem in the other SDKs.
We'll keep this issue open until is addressed.

PR #2787, opened 2020-09-22
Delete Conversation Reference from ConversationIdFactory when SkillDialog receives EndOfConversation while using ExpectReplies

EricDahlvang:
Fixes https://github.com/microsoft/botbuilder-js/issues/2712


stevengum (approved):
This looks good to me, but I would still hold on @gabog's approval

gabog (approved):

PR #2792, opened 2020-09-24
[PORT]Support several question mark operators in Expression

Danieladu:
Original issue: https://github.com/microsoft/botbuilder-dotnet/issues/4506
C# pr:https://github.com/microsoft/botbuilder-dotnet/pull/4689
https://github.com/microsoft/botbuilder-dotnet/pull/4686

### Nullish Coalescing
Nullish coalescing adds the ability to truly check nullish values instead of falsey values. 
For example:
`a ?? b ?? c`, if a is null/undefined return b, if b is also null or undefined return c

### Conditional Operator
Achieve a variable depending on a condition.
For example:
`a ? b : c `,  if 'a' is true, return b, otherwise return c.
which is equals to `if(a, b, c)`


chrimc62 (approved):
:shipit:

PR #2794, opened 2020-09-24
[PORT]Support null/undefined result in LG output

Danieladu:
issue link: https://github.com/microsoft/botbuilder-dotnet/issues/4620

## Description
Support `null` and `undefined` as the LG result.

## Specific Changes
1. Support `null` as the result
for example:
```
> Would return null but not "null"
# template1
- ${null}

> would return "the result is null"
# template2
-SWITCH: ${null}
  -CASE:${null}
     - the reuslt is null
  - DEFAULT:
      - the result is not null 
```

2. [Strict mode](https://docs.microsoft.com/en-us/azure/bot-service/file-format/bot-builder-lg-file-format?view=azure-bot-service-4.0#strict-option) keeps its original duty
3. [replaceNull option](https://docs.microsoft.com/en-us/azure/bot-service/file-format/bot-builder-lg-file-format?view=azure-bot-service-4.0#strict-option) keeps its original duty and only be applied in Expression library.


chrimc62 (approved):
:shipit:

PR #2812, opened 2020-09-30
Cherry-Pick: Change some utility functions in Adaptive-Expressions to internal

cosmicshuai:
Some utility functions in adaptive-expressions are exposed to public unexpected. In this PR, these public methods are changed to internal only by not export in index.ts.

Before this change:
When a project use adaptive-expressions with `"noImplicitAny": true`  is set in the tsconfig.json. It will throws an error:
image

After this change, this error will not happen again.

Related utility functions:

1. parseTimexProperty
2. toBinary
3. sortBy
4. accessIndex
5. verifyTimestamp
6. verifyISOTimestamp
7. parseTimestamp
8. ticks
9. accessProperty
10. wrapGetValue
11. parseStringOrUndefined
12. isLogicTrue
13. foreach
14. validateForeach
15. parseUri
16. timeUnitTransformer
17. returnFormattedTimeStampStr
18. isEqual


stevengum:
@cosmicshuai thanks for opening this PR. I'm not sure if we're going to do a patch release for this, I can see why we would and why we wouldn't. I'll raise the issue at the next SDK standup.

chrimc62 (approved):
:shipit:

PR #2859, opened 2020-10-01
Fixed some eslint issues in adaptive-expressions

GeoffCoxMSFT:
## Description
These are not auto-fixable eslint issues, so I fixed the ones that seemed straightforward and non-contentious.  Goes from 567 to 416 warnings = 151 warnings fixed.

## Specific Changes
- fixed prefer const issues
- removed unused imports
- removed unused variables
- underscore in front of contract unused variables

## Testing
N/A


libraries/adaptive-expressions/src/builtinFunctions/coalesce.ts
```diff
@@ -24,8 +24,8 @@ export class Coalesce extends ExpressionEvaluator {
         return FunctionUtils.apply((args: any[][]): any => Coalesce.evalCoalesce(args as any[]));
     }
 
-    private static evalCoalesce(objetcList: object[]): any {
-        for (const obj of objetcList) {
+    private static evalCoalesce(objectList: any[]): any {
```
GeoffCoxMSFT, line 27: I'm leaving avoiding using object as a type fixed for a later PR, except this one which was a private function.

joshgummersall (approved):
Looks good 👍

Issue #2234, opened 2020-05-13
Microsoft.QnAMakerDialog not registered


chon219:
## Versions
master branch

## Describe the bug
Microsoft.QnAMakerDialog component not correctly parsed in .dialog file.

## To Reproduce
Try to load a .dialog file with resource explorer which contains Microsoft.QnAMakerDialog components.

## Expected behavior
Correctly load .dialog files with no errors.

## Additional context

[bug]

PR #2873, opened 2020-10-07
Adaptive objects initialization refactoring

### LINKED ISSUE #2314: [Adaptive] Support inline object initialization via constructor for dialogs
Today, you need to use the configure method to initalize a dialog with properties. This is a bit cumbersome. 
```js
new SetProperty().configure({
        property: new StringExpression("dialog.itemTitle"),
        value: new ValueExpression("=@itemTitle")
    }),
```

Ask is to support object initialization of a dialog via constructor so the below is possible 

```js
new SetProperty({
    property: "dialog.itemTitle",
    value: "=@itemTitle"
}),
```

### LINKED ISSUE #2638: [PORT] Use Resource.Id as default Id for referenced objects
https://github.com/microsoft/botbuilder-dotnet/pull/4368

Fixes #4367

## Specific Changes
Changed resourceExplorer to Apply resource.id to referenced objects.

## Testing
Updated unit tests to have referenced objects too to validate.

# Changed projects

[R10]

chon219:
Fixes #2314 
Fixes #2638 

## Description

Refactored the way to initialize adaptive objects from configuration/dialog files.

## Specific Changes

  - Refactored component registration to register components for building objects.
  - Added `getConverter` method to each configurable class to convert raw configuration into property values.
  -


libraries/botbuilder-dialogs-adaptive/src/actions/setProperties.ts
```diff
@@ -5,28 +5,31 @@
...
  */
+import { ValueExpression, StringExpression, BoolExpression, BoolExpressionConverter } from 'adaptive-expressions';
 import { StringUtils } from 'botbuilder-core';
-import { Dialog, DialogContext, DialogTurnResult } from 'botbuilder-dialogs';
-import { Converter } from 'botbuilder-dialogs-declarative';
-import { ValueExpression, StringExpression, BoolExpression } from 'adaptive-expressions';
+import { Converter, Converters, Dialog, DialogContext, DialogTurnResult } from 'botbuilder-dialogs';
 import { replaceJsonRecursively } from '../jsonExtensions';
 
+class PropertyAssignmentsConverter implements Converter<any[], PropertyAssignment[]> {
```
joshgummersall, line 13:
```typescript
type Input<T> = { property: string, value: T };

class PropertyAssignmentsConverter<T = any> implements Converter<Input<T>[], PropertyAssignment[]> {
  public convert(assignments: Input<T>[]) PropertyAssignment[] {
    ...
  }
}
```

libraries/botbuilder-dialogs-adaptive/src/actions/switchCondition.ts
```diff
@@ -5,12 +5,28 @@
...
+    Expression,
+    ExpressionConverter,
+    ExpressionParser,
+    ValueExpression,
+} from 'adaptive-expressions';
+import { Converter, Converters, DialogTurnResult, DialogDependencies, Dialog, DialogContext } from 'botbuilder-dialogs';
 import { ActionScope } from './actionScope';
 import { Case } from './case';
 
+class CasesConverter implements Converter<any[], Case[]> {
```
joshgummersall, line 21:
```typescript
type Input<T> = { actions: Dialog[], value: T };

class CasesConverter<T = string> implements Converter<Input<T>[], Case[]> {
  public convert(value: Input<T>[]): Case[] {
    ...
  }
}
```

libraries/botbuilder-dialogs-adaptive/src/actions/telemetryTrackEventAction.ts
```diff
@@ -6,14 +6,18 @@
...
+    BoolExpressionConverter,
+    StringExpressionConverter,
+} from 'adaptive-expressions';
+import { Converter, Converters, Dialog, DialogContext, DialogTurnResult } from 'botbuilder-dialogs';
 
 /**
  * Converter to convert telemetry properties configuration.
  */
-export class TelemetryPropertiesConverter implements Converter {
+class TelemetryPropertiesConverter implements Converter<object, { [name: string]: StringExpression }> {
```
joshgummersall, line 20: I think the linter has been complaining about using `object`. `Record<string, unknown>` should work better.

libraries/botbuilder-dialogs-adaptive/src/adaptiveComponentRegistration.ts
```diff
@@ -0,0 +1,263 @@
...
+} from './recognizers';
+import { LuisAdaptiveRecognizer } from './luis';
+import { QnAMakerRecognizer } from './qnaMaker';
+import { ResourceMultiLanguageGenerator, TemplateEngineLanguageGenerator } from './generators';
+import { ConditionalSelector, FirstSelector, MostSpecificSelector, RandomSelector, TrueSelector } from './selectors';
+import { CustomDialogLoader } from './customDialogLoader';
+
+type Type = {
+    $kind: string;
+    new (): unknown;
```
joshgummersall, line 120: I think a better constructor type is `new (...args: unknown[]): unknown` in case the type does support constructor parameters.

libraries/botbuilder-dialogs-adaptive/src/converters/dialogExpressionConverter.ts
```diff
@@ -6,16 +6,12 @@
...
+import { ResourceExplorer } from 'botbuilder-dialogs-declarative';
 import { DialogExpression } from '../expressions';
 
-export class DialogExpressionConverter implements Converter {
-    private _resourceExplorer: ResourceExplorer;
-
-    public constructor(resouceExplorer: ResourceExplorer) {
-        this._resourceExplorer = resouceExplorer;
-    }
+export class DialogExpressionConverter implements Converter<string | object, DialogExpression> {
```
joshgummersall, line 13: Instead of `object` try `Record<string, unknown>`

libraries/botbuilder-dialogs-adaptive/src/converters/languageGeneratorConverter.ts
```diff
@@ -6,10 +6,11 @@
  * Licensed under the MIT License.
  */
 
-import { Converter } from 'botbuilder-dialogs-declarative';
+import { Converter } from 'botbuilder-dialogs';
 import { ResourceMultiLanguageGenerator } from '../generators';
 
-export class LanguageGeneratorConverter implements Converter {
+export class LanguageGeneratorConverter
+    implements Converter<string | ResourceMultiLanguageGenerator, ResourceMultiLanguageGenerator> {
```
joshgummersall, line 13: We could probably push down the `| T` piece to the underlying `Converter` type.

```typescript
type Converter<From, To> = (value: From | To) => To;
```

Then you wouldn't need to always include `| T` everywhere. Not sure if this will break backwards compat with any of the existing converters though.

libraries/botbuilder-dialogs-adaptive/src/converters/recognizerConverter.ts
```diff
@@ -24,3 +21,19 @@ export class RecognizerConverter implements Converter {
         return value as Recognizer;
     }
 }
+
+export class RecognizerListConverter implements Converter<any[], Recognizer[]> {
```
joshgummersall, line 25: Should be `Converter<string[] | object[], Recognizer[]>`

libraries/botbuilder-dialogs-adaptive/src/converters/multiLanguageRecognizerConverter.ts
```diff
@@ -6,11 +6,12 @@
...
  */
 
-import { Converter, ResourceExplorer } from 'botbuilder-dialogs-declarative';
+import { Converter } from 'botbuilder-dialogs';
+import { ResourceExplorer } from 'botbuilder-dialogs-declarative';
 import { Recognizer } from '../recognizers';
 import { RecognizerConverter } from './recognizerConverter';
 
-export class MultiLanguageRecognizerConverter implements Converter {
+export class MultiLanguageRecognizerConverter implements Converter<object, { [key: string]: Recognizer }> {
```
joshgummersall, line 14: `Record` over `object` if possible.

libraries/botbuilder-dialogs-adaptive/src/languagePolicy.ts
```diff
@@ -904,7 +904,7 @@ export class LanguagePolicy extends Map<string, string[]> {
     }
 }
 
-export class LanguagePolicyConverter implements Converter {
+export class LanguagePolicyConverter implements Converter<object, LanguagePolicy> {
```
joshgummersall, line 907: Try `Record<string, unknown>` if possible.

libraries/botbuilder-dialogs-adaptive/src/recognizers/crossTrainedRecognizerSet.ts
```diff
@@ -19,11 +21,19 @@ const deferPrefix = 'DeferToRecognizer_';
...
+    public static $kind = 'Microsoft.CrossTrainedRecognizerSet';
+
     /**
      * Gets or sets the input recognizers.
      */
     public recognizers: Recognizer[] = [];
 
+    public get converters(): Converters<CrossTrainedRecognizerSet> {
+        return {
+            recognizers: (resourceExplorer: ResourceExplorer) => new RecognizerListConverter(resourceExplorer),
```
joshgummersall, line 33: I've seen this same style of converter construction a few times. I wonder if we could express the value in the Converters type as:

```typescript
Converter<From, To> | new (...args: unknown[]): Converter<From, To>
```

and then these could all just be written as:

```typescript
return {
  recognizers: RecognizerListConverter
}
```

This would likely require some other code changes though so it's possible it would break backwards compat.

libraries/botbuilder-dialogs-adaptive/src/recognizers/regexRecognizer.ts
```diff
@@ -5,13 +5,21 @@
...
  */
-import { RecognizerResult, Entity, Activity } from 'botbuilder-core';
-import { DialogContext } from 'botbuilder-dialogs';
+import { Activity, Entity, RecognizerResult } from 'botbuilder-core';
+import { Converter, Converters, DialogContext } from 'botbuilder-dialogs';
 import { Recognizer } from './recognizer';
 import { IntentPattern } from './intentPattern';
 import { EntityRecognizer, TextEntity, EntityRecognizerSet } from './entityRecognizers';
 
+class IntentPatternsConverter implements Converter<any[], IntentPattern[]> {
```
joshgummersall, line 14:
```typescript
type Input = { intent: string, pattern: string }
class IntentPatternsConverter implements Converter<Input[], IntentPattern[]> {
  public convert(value: Input): IntentPattern[] {
    ...
  }
}
```

libraries/botbuilder-dialogs-declarative/src/customDeserializer.ts
```diff
@@ -6,6 +6,6 @@
  * Licensed under the MIT License.
  */
 
-export interface Converter {
-    convert(value: any): any;
+export interface CustomDeserializer {
```
joshgummersall, line 9: It would be great to parameterize this:

```typescript
export interface CustomerDeserializer<T, C> {
  load(config: C, type: new (...args: unknown[]): T): T;
}
```

libraries/botbuilder-dialogs-declarative/src/declarativeType.ts
```diff
@@ -0,0 +1,15 @@
...
+ * @module botbuilder-dialogs-declarative
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { CustomDeserializer } from './customDeserializer';
+
+export interface DeclarativeType {
```
joshgummersall, line 11:
```typescript
export interface DeclarativeType<T, C> {
  kind: string;
  type: new (...args: unknown[]): T;
  loader?: CustomDeserializer<T, C>;
}
```

libraries/botbuilder-dialogs-declarative/src/defaultLoader.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { Converter } from 'botbuilder-dialogs';
+import { CustomDeserializer } from './customDeserializer';
+import { ResourceExplorer } from './resources';
+
+export class DefaultLoader implements CustomDeserializer {
```
joshgummersall, line 13:
```typescript
export class DefaultLoader<T, C> implements CustomDeserializer<T, C> {
  ...

  public load(config: C, type: new (...args: unknown[]): T): T {
    ...
  }
}
```

libraries/botbuilder-dialogs-declarative/src/resources/resourceExplorer.ts
```diff
@@ -6,22 +6,27 @@
...
+import { DeclarativeType } from '../declarativeType';
+import { CustomDeserializer } from '../customDeserializer';
+import { DefaultLoader } from '../defaultLoader';
 
 /**
  * Class which gives standard access to content resources.
  */
 export class ResourceExplorer {
-    private _factory: TypeFactory = new TypeFactory();
+    private _kindToType: Map<string, new () => {}> = new Map();
```
joshgummersall, line 26: Instead of `{}` I think you want `unknown`.

libraries/botbuilder-dialogs-declarative/src/resources/resourceExplorer.ts
```diff
@@ -6,22 +6,27 @@
...
+    private _kindDeserializer: Map<string, CustomDeserializer> = new Map();
```
joshgummersall, line 27: This might require some care if type parameters are added.

libraries/botbuilder-dialogs/src/converter.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+/**
+ * The converter converts object from one type to another.
+ */
+export interface Converter<From = any, To = any> {
+    convert(value: From): To;
+}
+
+export type Converters<T> = {
```
joshgummersall, line 16: Left some notes about this via email.

joshgummersall (commented):
I like the approach, hopefully some of the type refactorings I suggested work well.

libraries/botbuilder-dialogs-declarative/src/resources/resourceExplorer.ts
```diff
@@ -240,9 +240,13 @@ export class ResourceExplorer {
         );
     }
 
-    private registerTypeInternal(kind: string, type: new () => {}, loader?: CustomDeserializer): void {
+    private registerTypeInternal(
```
joshgummersall, line 243: Maybe this would provide some more flexibility:

```typescript
private registerTypeInternal<T, C>(kind: string, type: new (...args: unknown[]): T, loader?: CustomerDeserializer<T, C> {
  ...
})
```

libraries/botbuilder-dialogs-adaptive/src/converters/recognizerConverter.ts
```diff
@@ -10,28 +10,28 @@ import { Converter } from 'botbuilder-dialogs';
 import { ResourceExplorer } from 'botbuilder-dialogs-declarative';
 import { Recognizer } from '../recognizers';
 
-export class RecognizerConverter implements Converter<any, Recognizer> {
+export class RecognizerConverter implements Converter<unknown, Recognizer> {
```
joshgummersall, line 13: This should be `Converter<string | unknown, Recognizer>` to match the `convert` method signature.

libraries/botbuilder-dialogs-adaptive/src/adaptiveComponentRegistration.ts
```diff
@@ -245,14 +245,14 @@ export class AdaptiveComponentRegistration extends ComponentRegistration impleme
...
-                    loader: new CustomDialogLoader(resourceExplorer),
+                    loader: new CustomDialogLoader<DynamicBeginDialog, Record<string, unknown>>(resourceExplorer),
                 });
             }
         });
         return declarativeTypes;
     }
 
-    private _addDeclarativeType(type: Type, loader?: CustomDeserializer): void {
+    private _addDeclarativeType(type: Type, loader?: CustomDeserializer<unknown, unknown>): void {
```
joshgummersall, line 255: Consider the following:

```typescript
type Type<T> = { $kind: String, new (...args: unknown[]): T }

...

private _addDeclarativeType<T, C>(type: Type<T>, loader?: CustomDeserializer<T, C>): void
```

libraries/botbuilder-dialogs-adaptive-testing/src/adaptiveTestComponentRegistration.ts
```diff
@@ -54,7 +54,7 @@ export class AdaptiveTestComponentRegistration extends ComponentRegistration imp
         return this._declarativeTypes;
     }
 
-    private _addDeclarativeType(type: Type, loader?: CustomDeserializer): void {
+    private _addDeclarativeType(type: Type, loader?: CustomDeserializer<unknown, unknown>): void {
```
joshgummersall, line 57: Consider

```typescript
private _addDeclarativeType<T, C>(type: Type<T>, loader?: CustomDeserializer<T, C>): void
```

libraries/botbuilder-dialogs-adaptive/src/actions/httpRequest.ts
```diff
@@ -8,21 +8,22 @@
...
+    ValueExpression,
     ValueExpressionConverter,
 } from 'adaptive-expressions';
 import { Activity } from 'botbuilder-core';
 import { Converter, Converters, DialogTurnResult, DialogContext, Dialog } from 'botbuilder-dialogs';
 import { replaceJsonRecursively } from '../jsonExtensions';
 
-class HttpHeadersConverter implements Converter<object, { [key: string]: StringExpression }> {
-    public convert(value: object): { [key: string]: StringExpression } {
+class HttpHeadersConverter implements Converter<Record<string, string | Expression>, Record<string, StringExpression>> {
```
joshgummersall, line 25: Since the `Converter` type parameters are slightly complex and duplicated in the `convert` signature, consider extracting them to type aliases such as:

```typescript
type Input = Record<string, string | Expression>
type Output = Record<string, StringExpression>

class HttpHeadersConverter implements Converter<Input, Output> {
  public convert(value: Input): Output { ... }
}
```

libraries/botbuilder-dialogs-adaptive/src/actions/telemetryTrackEventAction.ts
```diff
@@ -17,8 +17,8 @@ import { Converter, Converters, Dialog, DialogContext, DialogTurnResult } from '
 /**
  * Converter to convert telemetry properties configuration.
  */
-class TelemetryPropertiesConverter implements Converter<object, { [name: string]: StringExpression }> {
-    public convert(properties: { [name: string]: string }): { [name: string]: StringExpression } {
+class TelemetryPropertiesConverter implements Converter<Record<string, string>, Record<string, StringExpression>> {
+    public convert(properties: Record<string, string>): Record<string, StringExpression> {
```
joshgummersall, line 21: Same `Input`/`Output` type alias extraction suggestion.

libraries/botbuilder-dialogs-adaptive/src/converters/dialogExpressionConverter.ts
```diff
@@ -10,19 +10,19 @@ import { Converter, Dialog } from 'botbuilder-dialogs';
 import { ResourceExplorer } from 'botbuilder-dialogs-declarative';
 import { DialogExpression } from '../expressions';
 
-export class DialogExpressionConverter implements Converter<string | object, DialogExpression> {
+export class DialogExpressionConverter implements Converter<string | Record<string, unknown>, DialogExpression> {
```
joshgummersall, line 13: Same `Input`/`Output` type alias extraction suggestion.

libraries/botbuilder-dialogs-adaptive/src/converters/multiLanguageRecognizerConverter.ts
```diff
@@ -11,14 +11,14 @@ import { ResourceExplorer } from 'botbuilder-dialogs-declarative';
 import { Recognizer } from '../recognizers';
 import { RecognizerConverter } from './recognizerConverter';
 
-export class MultiLanguageRecognizerConverter implements Converter<object, { [key: string]: Recognizer }> {
+export class MultiLanguageRecognizerConverter implements Converter<Record<string, unknown>, Record<string, Recognizer>> {
```
joshgummersall, line 14: Same `Input`/`Output` type alias extraction suggestion.

libraries/botbuilder-dialogs/src/converter.ts
```diff
@@ -9,12 +9,12 @@
...
+export interface Converter<From = unknown, To = unknown> {
     convert(value: From): To;
 }
 
 export type Converters<T> = {
-    [K in keyof Partial<Omit<T, 'converters' | 'configure'>>]:
-        | Converter<any, T[K]>
-        | ((...args: any[]) => Converter<any, T[K]>);
-}
+    [K in keyof Partial<T>]: T[K] extends Function
```
joshgummersall, line 17: This is a step in the right direction, I think there are two types coupled here though. In the interested of separation of concerns they should be distinct.

The first type is `Properties<T>` that omits all functions from a type.

The second is the `Converters<T>` that expects the type parameter to contain the proper types. Users would then write `Converter<Properties<T>>`.

libraries/botbuilder-dialogs-adaptive/src/converters/languageGeneratorConverter.ts
chon219: We might need to update all existing converters to support "convert T to T".

libraries/botbuilder-dialogs-adaptive/src/converters/dialogListConverter.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Converter, Dialog } from 'botbuilder-dialogs';
+import { ResourceExplorer } from 'botbuilder-dialogs-declarative';
+
+export class DialogListConverter implements Converter<string[], Dialog[]> {
+    public constructor(private readonly _resourceExplorer: ResourceExplorer) {}
+
+    public convert(value: string[] | Dialog[]): Dialog[] {
+        const results: Dialog[] = [];
+
+        value.forEach((item: string | Dialog) => {
```
joshgummersall, line 18: Use `.map` when constructing an array of types.

```typescript
return value.map(item => {
  if (item instanceof Dialog) {
    return item;
  }
  ...
})
```

libraries/botbuilder-dialogs-adaptive/src/converters/multiLanguageRecognizerConverter.ts
```diff
@@ -6,21 +6,28 @@
...
+export class MultiLanguageRecognizerConverter implements Converter<Input, Output> {
     private _recognizerConverter: RecognizerConverter;
 
     public constructor(resouceExplorer: ResourceExplorer) {
         this._recognizerConverter = new RecognizerConverter(resouceExplorer);
     }
 
-    public convert(value: object): { [key: string]: Recognizer } {
+    public convert(value: Input | Output): Output {
         const recognizers = {};
```
joshgummersall, line 25: Use `reduce` when constructing an object like this.

```typescript
return Object.entries(value).reduce((recognizers, [key, value]) => {
  if (!Object.prototype...) {
    return recognizers;
  }
  return { ...recognizers, [key]: this._recognizerConverter.convert(value) };
}, {});
```

I actually think using `Object.entries` handles the `hasOwnProperty` filtering implicitly.

libraries/botbuilder-dialogs-adaptive/src/converters/recognizerConverter.ts
```diff
@@ -6,21 +6,34 @@
...
+export class RecognizerListConverter implements Converter<string[], Recognizer[]> {
+    private _recognizerConverter: RecognizerConverter;
+
+    public constructor(resourceExplorer: ResourceExplorer) {
+        this._recognizerConverter = new RecognizerConverter(resourceExplorer);
+    }
+
+    public convert(value: string[] | Recognizer[]): Recognizer[] {
+        const recognizers: Recognizer[] = [];
+        value.forEach((item: string | Recognizer) => {
```
joshgummersall, line 34: Use `map`.

libraries/botbuilder-dialogs-adaptive/src/recognizers/regexRecognizer.ts
```diff
@@ -5,13 +5,31 @@
...
+
+type Input = {
+    intent: string;
+    pattern: string;
+};
+
+class IntentPatternsConverter implements Converter<Input[], IntentPattern[]> {
+    public convert(items: Input[] | IntentPattern[]): IntentPattern[] {
+        const results: IntentPattern[] = [];
+        items.forEach((item) => {
```
joshgummersall, line 23: Use `map`.

libraries/botbuilder-dialogs-declarative/src/defaultLoader.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Configurable } from 'botbuilder-dialogs';
+import { CustomDeserializer } from './customDeserializer';
+import { ResourceExplorer } from './resources';
+
+export class DefaultLoader implements CustomDeserializer<Configurable, Record<string, unknown>> {
+    public constructor(private readonly _resourceExplorer: ResourceExplorer) {}
+
+    public load(config: Record<string, unknown>, type: { new (...args: unknown[]): Configurable }): Configurable {
+        const instance = new type();
+        Object.getOwnPropertyNames(config).forEach((k: string) => {
```
joshgummersall, line 18: I think `Object.entries(config).forEach(([key, value]) => 
  ...
{})` is more readable and clear.

libraries/botbuilder-dialogs-declarative/src/resources/resourceExplorer.ts
```diff
@@ -215,4 +222,42 @@ export class ResourceExplorer {
             this._eventEmitter.emit(event, resources);
         }
     }
+
+    private load<T extends object>(value: Record<string, any>): T {
```
joshgummersall, line 226: Prefer `unknown` over `any` as it's more semantically correct and requires proper type checking/casting at runtime.

libraries/botbuilder-dialogs/src/configurable.ts
```diff
@@ -14,28 +16,37 @@ export abstract class Configurable {
...
                     }
-                } else if (setting !== undefined) {
-                    this[key] = setting;
                 }
             }
         }
         return this;
     }
+
+    public getConverter(_property: string): Converter | ConverterFactory {
```
joshgummersall, line 49: Should this be `abstract` as well to force extension classes to implement the functionality?

libraries/botbuilder-dialogs/src/converter.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+/**
+ * The converter converts object from one type to another.
+ */
+export interface Converter<From = unknown, To = unknown> {
+    convert(value: From | To): To;
```
joshgummersall, line 13: Very nice!

joshgummersall (commented):
I left at least one comment about adjusting the `Input` type for a `Converter` to include the full `convert(from: From | To): To` type signature. Ignore those comments, I didn't read the adjusted `Converter` type signature before making those comments.

libraries/botbuilder-dialogs-adaptive/src/converters/dialogExpressionConverter.ts
```diff
@@ -6,27 +6,28 @@
...
 
-import { Converter, ResourceExplorer } from 'botbuilder-dialogs-declarative';
-import { Dialog } from 'botbuilder-dialogs';
+import { Converter, Dialog } from 'botbuilder-dialogs';
+import { ResourceExplorer } from 'botbuilder-dialogs-declarative';
 import { DialogExpression } from '../expressions';
 
-export class DialogExpressionConverter implements Converter {
-    private _resourceExplorer: ResourceExplorer;
+type Input = string | Record<string, unknown>;
```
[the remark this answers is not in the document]
joshgummersall: Ignore this 😄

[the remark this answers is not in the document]
chon219: Yes, we'll treat both Input | DialogExpression as valid input types.

libraries/botbuilder-dialogs/src/configurable.ts
chon219: I was considering make this method abstract, however if we do that, every non-abstract class inherited Dialog must implement this method, many unrelated classes would be touched.

libraries/botbuilder-dialogs-adaptive/src/converters/dialogListConverter.ts
chon219: There is an error when using `.map`
This expression is not callable.
  Each member of the union type '(<U>(callbackfn: (value: string, index: number, array: string[]) => U, thisArg?: any) => U[]) | (<U>(callbackfn: (value: Dialog<{}>, index: number, array: Dialog<{}>[]) => U, thisArg?: any) => U[])' has signatures, but none of those signatures are compatible with each other.ts(2349)

libraries/botbuilder-dialogs-adaptive-testing/src/actions/assertCondition.ts
```diff
@@ -6,13 +6,27 @@
...
+
+export interface AssertConditionConfiguration extends DialogConfiguration {
+    condition?: string | Expression;
+    description?: string | Expression | StringExpression;
+}
 
 /**
  * Dialog action which allows you to add assertions into your dialog flow.
  */
 export class AssertCondition<O extends object = {}> extends Dialog<O> {
```
joshgummersall, line 27: I think adding `implements AssertConditionConfiguration` makes sense as well.

libraries/botbuilder-dialogs-adaptive/src/actions/deleteProperties.ts
```diff
@@ -5,10 +5,34 @@
...
+    Dialog,
+    DialogConfiguration,
+    DialogContext,
+    DialogTurnResult,
+} from 'botbuilder-dialogs';
+
+class PropertiesConverter implements Converter<string[], StringExpression[]> {
+    public convert(value: string[] | StringExpression[]): StringExpression[] {
+        const results: StringExpression[] = [];
+        value.forEach((item: string | StringExpression) => {
```
joshgummersall, line 21: `.map`

joshgummersall (commented):
Couple quick comments, I think the interface implementation you came up with is great though. Nice work!

libraries/botbuilder-dialogs-adaptive/src/converters/dialogListConverter.ts
joshgummersall: You might need to add the same callback signature above `value.map((item: string | Dialog) => { ... })`

libraries/botbuilder-dialogs/src/configurable.ts
joshgummersall: Fair enough, good point.

joshgummersall:
Does this include a fix for https://github.com/microsoft/botbuilder-js/issues/2640?

libraries/botbuilder-dialogs-adaptive/src/converters/dialogListConverter.ts
chon219: I think the problem is: the input type of convert is `string[] | Dialog[]` not `(string | Dialog)[]`, so the child type can only be `string` or `Dialog` in a map iteration, not both. However, If I changed the input type to `(string | Dialog)[]`, it may violate the type defined in `Converter`, which is `convert(value: From | To): To`, any ideas?

chon219:
> Does this include a fix for #2640?

I'll try to fix cycle reference in this PR.

joshgummersall:
@chon219 This pr is already quite big. Let's save that for a separate pr to make it easier to review.

chon219:
> @chon219 This pr is already quite big. Let's save that for a separate pr to make it easier to review

OK, I'll fix that after this PR merged.

libraries/botbuilder-dialogs-declarative/src/componentDeclarativeTypes.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Licensed under the MIT License.
+ */
+
+import { DeclarativeType } from './declarativeType';
+import { ResourceExplorer } from './resources';
+
+/**
+ * Interface for registering declarative types.
+ */
+export interface ComponentDeclarativeTypes {
```
joshgummersall, line 15: Would it be helpful to add the type parameters from `DeclarativeType` to this type as well?

joshgummersall:
Does this include a fix for https://github.com/microsoft/botbuilder-js/issues/2234? It looked like there was another PR linked that might be superseded by this PR.

libraries/botbuilder-dialogs-adaptive/src/dynamicBeginDialogDeserializer.ts
```diff
@@ -0,0 +1,49 @@
...
+     * @param type The object type that the configuration will be deserialized to.
+     * @returns A `DynamicBeginDialog` object created from the configuration.
+     */
+    public load(
+        config: BeginDialogConfiguration,
+        type: { new (...args: unknown[]): DynamicBeginDialog }
+    ): DynamicBeginDialog {
+        config.dialog = this._resourceExplorer.loadType<Dialog>(this._resourceId);
+        return Object.entries(config).reduce((instance, [key, value]) => {
+            let converter = instance.getConverter(key as keyof BeginDialogConfiguration);
```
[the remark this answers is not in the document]
chon219: > Does this include a fix for #2234? It looked like there was a draft PR linked that might be superseded by this PR.
> 
> EDIT: Looks like #2142 was also linked to the draft PR.

No, I would prefer fixing these two issues in another PR after this.

chon219:
> Does this include a fix for #2234? It looked like there was a draft PR linked that might be superseded by this PR.
> 
> EDIT: Looks like #2142 was also linked to the draft PR.

No, I would prefer fixing these two issues in another PR after this.

libraries/botbuilder-dialogs-adaptive/src/dynamicBeginDialogDeserializer.ts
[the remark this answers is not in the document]
chon219: For a `DynamicBeginDialog`, there might be some user defined properties, we cannot limit the key of object to `BeginDialogConfiguration`, so I have to use `as keyof` to avoid build errors.

libraries/botbuilder-dialogs-declarative/src/componentDeclarativeTypes.ts
chon219: The `getDeclarativeTypes` method would return a bunch of different `DeclarativeType`s, how can we add the type to `ComponentDeclarativeTypes`?

joshgummersall (approved):
Nice work!

stevengum (approved):

Issue #2864, opened 2020-10-02
formatDateTime(<date>, 'dd/MM/yyyy) should produce "dd/MM/yyyy" and not "dd:MM:yyyy"


stevengum:
## Versions
adaptive-expressions ^4.9.2

## Describe the bug
Originally reported in https://github.com/microsoft/AdaptiveCards/issues/4842

> I agree that `formatDateTime(<date>, 'dd/MM/yyyy)` should produce "dd/MM/yyy" and not "dd:MM:yyyy". This would be an Adaptive Expression issue though.
>
> {{DATE}} uses the system's (browser's) locale settings to generate date/time strings in the user's selected locale.

_@dclaux via https://github.com/microsoft/AdaptiveCards/issues/4842#issuecomment-702856391_

`formatDateTime(<date>, 'dd/MM/yyyy)` isn't covered in [our unit tests](https://github.com/microsoft/botbuilder-js/blob/2fff3d8a62c44ba07be2a528672e7d9206a8b916/libraries/adaptive-expressions/tests/expressionParser.test.js#L496-L516) and needs to be added.

@boydc2014 can you size this issue and assign a priority and owner for R11? Thanks!

stevengum:
@boydc2014 The [original issue's author](https://github.com/microsoft/AdaptiveCards/issues/4842#issuecomment-703164981) indicated that `dd/MM/yy` is also a desirable format.

boydc2014:
Thanks @stevengum, @cosmicshuai will look at this soon

PR #2889, opened 2020-10-13
fix formatDateTime function with "dd/MM/yyy" produces wrong results

cosmicshuai:
closes: #2864 
Now `formatDateTime('2018-02-15T12:12:12', \'dd/MM/yyy\')` will return `02/15/2018`


Danieladu (approved):

libraries/adaptive-expressions/tests/expressionParser.test.js
```diff
@@ -508,6 +508,8 @@ const testCases = [
             ['date(timestamp)', '3/15/2018'],//Default. TODO
             ['year(timestamp)', 2018],
             ['length(utcNow())', 24],
+            ['formatDateTime(notISOTimestamp, \'dd/MM/yyy\')', '15/03/2018'],
```
stevengum, line 511: Do you think a test case for `'dd/MM/yy'` is necessary? The changes in datetimeFormatConverter are trivial so I don't believe another test case for that is required, but let's be explicit since this is a customer ask.

[Relevant comment](https://github.com/microsoft/botbuilder-js/issues/2864#issuecomment-704507726)

stevengum (approved):
Looks good, just one question to cover our bases. 👍

stevengum:
@cosmicshuai do we have test coverage for this in .NET and Python?

cosmicshuai:
> @cosmicshuai do we have test coverage for this in .NET and Python?

@stevengum 
I will check this in the python lib since it also have a formatConverter to keep the date time format align with the dotnet.

libraries/adaptive-expressions/tests/expressionParser.test.js
cosmicshuai: add a test cases in addDays function

PR #2906, opened 2020-10-16
make adaptive expression library compatible with IE

atishayv:
##Fixes:
Make adaptive expression library compatible with IE

## Description
There was no support of IE in the adaptive expression library. Using babel to transpile compiled code and add polyfills so that it works in IE also.

## Specific Changes
- Using babel to transpile adaptive-expression and other dependent libraries which are not compatible with IE.
- Adding core js to resolve all polyfills which are used in those dependent libraries.
- Updated templating library package json to create a webpack build in dist folder which is compatible with IE

## Testing
Tested manually by taking dependecy in adaptive card designer in IE to make sure that there are no errors


atishayv:
CC: @dclaux, @jaisanth

libraries/adaptive-expressions/webpack.config.js
```diff
@@ -0,0 +1,54 @@
+const path = require('path');
+const defaultTSConfig = require('./tsconfig.json');
+
+module.exports = (env, argv) => {
+    return {
+        mode: 'none',
+
+        entry: {
+            index: './src/index.ts',
```
stevengum, line 9: Should this be `index: path.resolve(__dirname, './src/index.ts')`?

libraries/adaptive-expressions/webpack.config.js
```diff
@@ -0,0 +1,54 @@
...
+                    test: /\.tsx?$/,
+                    loader: 'ts-loader',
+                    options: {
+                        compilerOptions: {
+                            ...defaultTSConfig.compilerOptions,
+                            outDir: './dist',
+                            target: 'es5',
+                        },
+                    },
+                    exclude: /node_modules/,
```
stevengum, line 32: I believe a safe-list would be better than instead of excluding `node_modules`. The entry is already `'./src/index.ts'` and a safe-list provides a better "locked-down" mechanism that ensures we only ship and webpack what we mean to.
```suggestion
                    include: [
                        path.resolve(__dirname, './src')
                    ],
```

libraries/adaptive-expressions/babel.config.js
```diff
@@ -0,0 +1,26 @@
...
+                },
+            },
+        ],
+    ];
+    const plugins = [
+        ['@babel/plugin-proposal-decorators', { decoratorsBeforeExport: true }],
+        ['@babel/plugin-proposal-class-properties'],
+        ['@babel/plugin-transform-modules-commonjs'],
+        ['@babel/plugin-transform-runtime', { useESModules: false }],
+    ];
```
stevengum, line 21: This is toolchain config and not consumed by end-users, but it would be great if you added some comments explaining what's going on here for future maintainers who may not be familiar with the current scenario and/or babel ecosystem.

atishayv: Added comments . Let me know if its okay

libraries/adaptive-expressions-ie11/webpack.config.js
```diff
@@ -0,0 +1,44 @@
+const path = require('path');
+
+module.exports = (env, argv) => {
+    return {
+        mode: 'none',
+        entry: ['core-js/modules/es.array.find', path.resolve(__dirname, './src/index.ts')],
```
joshgummersall, line 6: Now that this is a separate package let's put the polyfill in the entry point directly.

joshgummersall (changes requested):
One small comment, can you verify that this bundle still works for your use case as well?

libraries/adaptive-expressions-ie11/webpack.config.js
atishayv: Actualy we dont need that polyfill now as that was added because adaptive expression library was using Array.find before.However, now we are passing adaptive-expression also to babel which has core js built in as usage, so it will autmatically get added. 
Nice catch :-)

atishayv:
> One small comment, can you verify that this bundle still works for your use case as well?

Yes, I did verify it with adaptive-designer. I had to add aliases for the same but it worked. These are the 3 changes I did : 
1) package json, I removed the expression dependency and added dependency for ie 11 : 
image
2) Added alias in webpack:
image
3) Added alias in tsconfig:
image

CC : @dclaux, @jaisanth

atishayv:
Sorry, by mistake i closed the PR. Have reopened it

joshgummersall (approved):
Looks good to me! Thanks for your work on this 👍

atishayv:
> Looks good to me! Thanks for your work on this 👍

Can we merge this ?

libraries/adaptive-expressions-ie11/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from 'adaptive-expressions';
```
stevengum, line 1: I feel a bit silly for saying this, but can you add the appropriate copyright headers to the file?

```suggestion
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT License.

export * from 'adaptive-expressions';
```

libraries/adaptive-expressions-ie11/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "webpack": "^4.43.0",
+    "webpack-cli": "^3.3.12"
+  },
+  "scripts": {
+    "build": "webpack && copyfiles -u 3 node_modules/adaptive-expressions/lib/**/*.d.ts dist",
+    "clean": "erase /q /s .\\dist",
+    "set-version": "npm version --allow-same-version ${Version}"
+  },
+  "files": [
+    "src",
```
stevengum, line 47: Is there any benefit to us including `"src"`? If someone wanted to use the `"src"`, they should likely instead use `adaptive-expressions`, right?

stevengum, line 47:
```suggestion
```

stevengum (approved):

libraries/adaptive-expressions-ie11/package.json
atishayv: Yes, thanks for pushing those changes

PR #3077, opened 2020-11-19
Compute LG SourceRange based on content instead of parsed token

### LINKED ISSUE #3076: lg update template to the wrong position.
## Versions
What os are you using

## Describe the bug

## To Reproduce
1. Prepare a lg file with content：
```
# Exit
-Thanks for using todo bot.

# Greeting
-What's up bro
```

2. update `#Exit` with body
```
-Thanks for using todo bot.
- Bye.
                 // notice here is a new line
```

3. use step2 returned `resource`, continue do update on `#Greeting` with body
```
-Hi
``` 

4. See LG file will be updated to 
```
# Exit
-Thanks for using todo bot.
- Bye.

# Greeting
-Hi
-What's up bro   // this line should be replaced.
```

with more investigate, we find after step2, in returned `resource`, all templates below `#Exit`  source range is not be updated. it should have offset, cause step2 add a  new line, the template after it should also update it's source range.

## Expected behavior
Give a clear and concise description of what you expected to happen.

Danieladu:
Fixes #3076

### Root cause
Previously, the sourceRange(start and end line position) of a LG template is calculated based on the parsed token. But in certain scenarios that, when there are extra empty character (line breaks) at the end of a template, will cause this calculation not correct, and thus causing the further update on this lg file misplaced in the wrong position, causing potential data loss. 

### Fix
Use the real content received to calculate start\end line.  Content is the golden truth, more trustworthy than parsed tokens. 

### Impact
Composer

### DIFF
--- a/libraries/botbuilder-lg/src/templates.ts
+++ b/libraries/botbuilder-lg/src/templates.ts
@@ -409,6 +409,10 @@ export class Templates implements Iterable<Template> {
         if (updatedTemplates.toArray().length > 0) {
             const newTemplate = this.adjustSingleTemplateRange(updatedTemplates.toArray()[0], content);
             this.adjustRangeForAddTemplate(newTemplate, originalStartLine);
+
+            // adjust the previous range
+            this.items[this.items.length - 1].sourceRange.range.end.line = newTemplate.sourceRange.range.start.line - 1;
+
             this.items.push(newTemplate);
             new StaticChecker(this).check().forEach((u): number => this.diagnostics.push(u));
         }
--- a/libraries/botbuilder-lg/src/templates.ts
+++ b/libraries/botbuilder-lg/src/templates.ts
@@ -576,7 +576,7 @@ export class Templates implements Iterable<Template> {
 
     /**
      * @private
-     * Use content range replace the range from lexer.
+     * Compute LG SourceRange based on content instead of parsed token.
      * */
     private recomputeSourceRange(template: Template, content: string): Template {
         if (content != null) {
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-lg/src/templates.ts
```diff
@@ -567,6 +572,19 @@ export class Templates implements Iterable<Template> {
         return destList.join(this.newLine);
     }
 
+    /**
+     * @private
+     */
+    private adjustSingleTemplateRange(template: Template, content: string): Template {
```
boydc2014, line 578: Rename to recomputeSourceRange

libraries/botbuilder-lg/src/templates.ts
```diff
@@ -406,8 +407,12 @@ export class Templates implements Iterable<Template> {
         this.appendDiagnosticWithOffset(updatedTemplates.diagnostics, originalStartLine);
 
         if (updatedTemplates.toArray().length > 0) {
-            const newTemplate = updatedTemplates.toArray()[0];
+            const newTemplate = this.adjustSingleTemplateRange(updatedTemplates.toArray()[0], content);
             this.adjustRangeForAddTemplate(newTemplate, originalStartLine);
+
+            // adjust the previous range
```
boydc2014, line 413: adjust the last template's range when adding the template

libraries/botbuilder-lg/src/templates.ts
```diff
@@ -406,8 +407,12 @@ export class Templates implements Iterable<Template> {
...
+            this.items[this.items.length - 1].sourceRange.range.end.line = newTemplate.sourceRange.range.start.line - 1;
```
boydc2014, line 414: check index overflow?

boydc2014 (approved):

PR #3097, opened 2020-11-30
Make entity recognizers to be simple Recognizer

### LINKED ISSUE #3039: [PORT] Reduce concept count by making all entity recognizers to be simple Recognizer
https://github.com/microsoft/botbuilder-dotnet/pull/4848

EntityRecognizers are simple wrappers around TextAnalytics APIs, but currently are only usable from within RegExRecognizer via the EntityRecognizerSet.
This delta makes all EntityRecognizers full blown recognizers, allowing them to be composed like any other recognizer in the system.

This allows composition like this to feed external recgonizer results into LUIS recognizer:

```
{
    "$schema": "../schemas/app.schema",
    "$kind": "Microsoft.LuisRecognizer",
    "id": "LUIS_msmeeting-actions",
 ...
    "externalEntityRecognizer": {
        "$kind": "Microsoft.RecognizerSet",
        "recognizers": [
            { $kind": "Microsoft.PartNumberRecognizer" },
            { $kind": "Iciclecreek.QuotedTextEntityRecognizer" },
..
        ]
    }
}
```

* Changed EntityRecognizer base class
  * to derive from Recognizer, so it can be used as a recognizer.
  * implement standard RecgonizeAsync() method to return RecognizerResult data structure.
* Added ClientMentionEntityRecognizer to map client passed mention entities => RecognizerResult entity structure, allowing that data to be passed to LuisRecognizer.

### LINKED ISSUE #3044: [PORT] fix startIndex/endindex for external entities to match luis recognizer result of endIndex = start + length
https://github.com/microsoft/botbuilder-dotnet/pull/4884

Luis and Text Analytics libraries return entity endIndex as array index of last char.
SDK maps LUIS Recognizer Results to use EndIndex = start + length (aka the index of the first char to not keep
The Entity Recognizers are wrappers around TextAnalytics libraries and was not changing the endIndex to match the standard that LUIS was emitting, which meant that the text ranges were incorrect.

* Change mapping of text recognizer ModelResult => entity structure to make the end + 1;
* filter seed text entities (that was internal bookeeping which really shouldn't have been pushed out.

chon219:
Fixes #3039 
Fixes #3044 

## Description

EntityRecognizers are simple wrappers around TextAnalytics APIs, but currently are only usable from within RegExRecognizer via the EntityRecognizerSet.
This delta makes all EntityRecognizers full blown recognizers, allowing them to be composed like any other recognizer in the system.

## Specific Changes

  - Make `EntityRecognizer` class
  - Added `ChannelMentionEntityRecognizer`
  - Exposed `EntityRecognizerSet` in `AdaptiveComponentRegistration`
  - Updated `@microsoft/recognizers-text-suite` to latest version
  - Renamed protected method `recognize` in all EntityRecognizers to avoid conflicts with `recognize` method from parent EntityRecognizer
  - Change mapping of text recognizer ModelResult => entity structure to make the end + 1
  - Filter seed text entities

## Testing

Ported tests from C#.


libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/channelMentionEntityRecognizer.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        const entities = result.entities;
+
+        // prompt external mention entities from the activity into recognizer result
+        if (activity.entities) {
+            let iStart = 0;
+            activity.entities
+                .filter((e) => e.type === 'mention')
+                .forEach((entity) => {
+                    if (!entities.channelMention) {
```
joshgummersall, line 47: This `if` block can be written more succinctly as `entities.channelMention ??= [];`

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/channelMentionEntityRecognizer.ts
```diff
@@ -0,0 +1,81 @@
...
+
+                    entities.channelMention.push(entity.mentioned);
+
+                    let instance = entities['$instance'];
+                    if (!instance) {
+                        instance = {};
+                        entities['$instance'] = instance;
+                    }
+
+                    if (!instance.channelMention) {
```
joshgummersall, line 59: Same suggestion, `instance.channelMention ??= [];`

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
```diff
@@ -6,14 +6,122 @@
...
+     * @returns {Promise<RecognizerResult>} Analysis of utterance.
+     */
+    public async recognize(
+        dialogContext: DialogContext,
+        activity: Partial<Activity>,
+        telemetryProperties?: Record<string, string>,
+        telemetryMetrics?: Record<string, number>
+    ): Promise<RecognizerResult> {
+        // Identify matched intents
+        const text = activity.text || '';
```
joshgummersall, line 33: `const text = activity.text ?? '';`

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
```diff
@@ -6,14 +6,122 @@
...
+            intents: {},
+        };
+
+        if (!text) {
+            // nothing to recognize, return empty result
+            return recognizerResult;
+        }
+
+        const textEntity = new TextEntity(text);
+        textEntity['start'] = 0;
```
joshgummersall, line 46: We should be using dot access whenever possible, can this be written as `textEntity.start = 0;`?

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
```diff
@@ -6,14 +6,122 @@
...
+        // map entityPool of Entity objects => RecognizerResult entity format
+        const entities: Record<string, unknown> = {};
+        entityPool
+            .filter((e) => e !== textEntity)
+            .forEach((entityResult: Entity) => {
+                const entityType = entityResult.type.toLowerCase();
+
+                // add value
+                let values = [];
+                if (!Object.prototype.hasOwnProperty.call(entities, entityType)) {
```
joshgummersall, line 71: Is there a good reason to use `hasOwnProperty` from the Object prototype? Why can't this be `if (!entities[entityType]) { ... }`

joshgummersall: I think this code could be more succinctly written as:

```typescript
entities[entityType] ??= [];
entities[entityType].push(...);
```

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
```diff
@@ -6,14 +6,122 @@
...
+                } else {
+                    values = entities[`${entityType}`] as unknown[];
+                }
+
+                // the Entity type names are not consistent, map everything to camel case so we can process them cleaner.
+                values.push(entityResult.text);
+
+                // get/create $instance
+                let instanceRoot = {};
+                if (!Object.prototype.hasOwnProperty.call(entities, '$instance')) {
```
joshgummersall, line 82:
```typescript
entities['$instance'] ??= {};
const instanceRoot = entities['$instance'];
```

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
```diff
@@ -6,14 +6,122 @@
...
+                    entities['$instance'] = instanceRoot;
+                } else {
+                    instanceRoot = entities['$instance'];
+                }
+
+                // add instanceData
+                let instanceData = [];
+                if (!Object.prototype.hasOwnProperty.call(instanceRoot, entityType)) {
```
joshgummersall, line 90:
```typescript
instanceRoot[entityType] ??= [];
instanceRoot[entityType].push(...);
```

libraries/botbuilder-dialogs-adaptive/package.json
```diff
@@ -28,7 +28,7 @@
     }
   },
   "dependencies": {
-    "@microsoft/recognizers-text-suite": "1.1.4",
+    "@microsoft/recognizers-text-suite": "^1.3.0",
```
joshgummersall, line 31: I recall this library is used in more than one package - is it important that we use the same version across all our packages?

joshgummersall (changes requested):

libraries/botbuilder-dialogs-adaptive/package.json
chon219: There might be some bugs in the old version of this package, so I have to update the package or the test cases I ported from C# will fail.

chon219: After searching across the repo, I found this package was used by the following packages:
* adaptive-expressions
* botbuilder-ai
* botbuilder-dialogs
* botbuilder-dialogs-adaptive
* testbot

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
chon219: start/end/score are not builtin properties in `TextEntity`, we use it by extending the `[key: string]: any` part of `Entity` interface. How can we avoid lint errors if writing these properties using dot access?

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
chon219: Yes, you are right.

libraries/botbuilder-dialogs-adaptive/package.json
joshgummersall: Do we need it to be consistent across these packages?

chon219: I think we'd better change the version to `^1.3.0` over all packages.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
joshgummersall: Screen Shot 2020-11-30 at 7 04 22 PM

I don't see any lint errors in VS Code here - does `yarn lint` raise errors on this?

chon219: Maybe it is `TextEntity` (class) not `Entity` (interface)?
image

Cast to `Entity` would resolve the error:
image

joshgummersall: As long as there is no issue in typing `textEntity` as `Entity` then I would do that.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/channelMentionEntityRecognizer.ts
```diff
@@ -0,0 +1,76 @@
...
+        if (activity.entities) {
+            let iStart = 0;
+            activity.entities
+                .filter((e) => e.type === 'mention')
+                .forEach((entity) => {
+                    entities.channelMention ??= [];
+                    entities.channelMention.push(entity.mentioned);
+
+                    let instance = entities['$instance'];
+                    if (!instance) {
```
joshgummersall, line 51: This should be written with the `??=` style as well.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
```diff
@@ -6,14 +6,107 @@
...
+        // map entityPool of Entity objects => RecognizerResult entity format
+        const entities: Record<string, unknown> = {};
+        entityPool
+            .filter((e) => e !== textEntity)
+            .forEach((entityResult: Entity) => {
+                const entityType = entityResult.type;
+
+                // add value
+                entities[`${entityType}`] ??= [];
+                (entities[`${entityType}`] as unknown[]).push(entityResult.text);
```
joshgummersall, line 71: Type checking is safer than using `as`.

```typescript
entities[`${entityType}`] ??= [];
const entities = entities[`${entityType}`];
if (Array.isArray(entitities)) {
  entitites.push(entityResult.text);
} else {
  // throw?
}
```

libraries/botbuilder-dialogs-adaptive/src/recognizers/regexRecognizer.ts
```diff
@@ -94,81 +91,67 @@ export class RegexRecognizer extends Recognizer implements RegexRecognizerConfig
...
-                                entity['text'] = text;
-                                entity['start'] = match.index;
-                                entity['end'] = match.index + text.length;
-                                entityPool.push(entity);
-                            }
-                        }
-                    }
+            // TODO length weighted match and multiple intents
+            const intentKey = intentPattern.intent.split(' ').join('_');
+            (recognizerResult.intents[`${intentKey}`] as { score: number; pattern: string }) ??= {
```
joshgummersall, line 100: Type checking preferred to `as` casting.

joshgummersall (changes requested):
Couple of small changes, looks good though! Nice use of `reduce` and `??=`.

libraries/botbuilder-dialogs-adaptive/package.json
stevengum: We’ve had past difficulties with this collection of dependencies in JS and .NET. They tend to not follow semver... we should consult with @EricDahlvang and do a hand-off of this dependency’s validation process for uptake into the JS SDK.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/channelMentionEntityRecognizer.ts
```diff
@@ -0,0 +1,76 @@
...
+                    if (iStart >= 0) {
+                        instance.channelMention.push({
+                            startIndex: iStart,
+                            endIndex: iStart + mentionedText.length,
+                            text: mentionedText,
+                            score: 1.0,
+                        });
+
+                        // note, we increment so next pass through continues after the token we just processed.
+                        iStart++;
```
joshgummersall, line 69: Would it be more efficient to do `iStart += mentionedText.length` to move to the end of this token, rather than by just one char?

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
```diff
@@ -6,14 +6,107 @@
...
+            intents: {},
+        };
+
+        if (!text) {
+            // nothing to recognize, return empty result
+            return recognizerResult;
+        }
+
+        const textEntity: Entity = new TextEntity(text);
+        textEntity.start = 0;
```
joshgummersall, line 46: It's starting to seem like these should actually be properties on the `TextEntity` class itself, and probably computed inside the constructor. What do you think?

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/channelMentionEntityRecognizer.ts
chon219: what if the second entity text is a substring of the first one or vice versa?

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
chon219: Yes, every time we referred `TextEntity`, we referred start/end/score properties although they are not there in the implementation of C# version. Do you think we need to expand the `TextEntity` class to have these properties? I'm not sure because we don't have these in C#.

libraries/botbuilder-dialogs-adaptive/src/recognizers/regexRecognizer.ts
chon219: I ported `IntentScore` interface from C#, whose can be expanded to store properties other than `score: number`:
```
interface IntentScore {
  score?: number;
  [key: string]: unknown;
}
```
Changed the `intents` field of `RecognizerResult` from `{ [name: string]: { score: number } }` to `Record<string, IntentScore>` to reflect the change, I believe this would not break the backward compatibility.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/channelMentionEntityRecognizer.ts
joshgummersall: Got it, makes sense. I didn't consider the substring case.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
joshgummersall: Seems like a fix could be made in both places, but we don't need to block on that for now. I'll file a separate issue.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
```diff
@@ -6,14 +6,112 @@
...
+            .forEach((entityResult: Entity) => {
+                const { type: entityType, text: entityText } = entityResult;
+
+                // add value
+                entities[`${entityType}`] ??= [];
+                const value = entities[`${entityType}`];
+                if (Array.isArray(value)) {
+                    value.push(entityText);
+                } else {
+                    entities[`${entityType}`] = [value, entityText];
```
joshgummersall, line 75: This line should never really be hit, correct? Due to the `entities[`${entitityType}`] ??= [];` above. Perhaps the `else` block should be omitted.

libraries/botbuilder-dialogs-adaptive/src/recognizers/regexRecognizer.ts
```diff
@@ -94,81 +91,67 @@ export class RegexRecognizer extends Recognizer implements RegexRecognizerConfig
...
+            // TODO length weighted match and multiple intents
+            const intentKey = intentPattern.intent.split(' ').join('_');
+            recognizerResult.intents[`${intentKey}`] ??= {
+                score: 1.0,
+                pattern: intentPattern.pattern,
+            };
+
+            matches.forEach((match: RegExpExecArray) => {
+                if (match.groups) {
+                    Object.entries(pickBy(match.groups)).forEach(([name, text]) => {
```
joshgummersall, line 107: `pickBy` isn't really necessary is it?

libraries/botbuilder-dialogs-adaptive/package.json
EricDahlvang: @stevengum  https://github.com/microsoft/botbuilder-js/pull/1934

libraries/botbuilder-dialogs-adaptive/src/recognizers/regexRecognizer.ts
chon219: `pickBy` was used to filter out empty values from matched groups, it is necessary since values are always not empty.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
chon219: Removed `else` branch.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
chon219: Let's just leave it unchanged here and do it in another PR?

libraries/botbuilder-dialogs-adaptive/package.json
chon219: @joshgummersall @stevengum @EricDahlvang I would change the version back to `1.1.4` to align with other packages, and I have to made some changes on the test cases to let the tests pass.

libraries/botbuilder-dialogs-adaptive/src/recognizers/entityRecognizers/entityRecognizer.ts
joshgummersall: Yep, exactly.

joshgummersall (approved):

PR #3180, opened 2021-01-08
Fix MemoryAssertions and test

mdrichardson:
Related: https://github.com/microsoft/botbuilder-dotnet/pull/5084

## Description

The MemoryAssertions tests were silently failing. They'll now pass and also elicit errors if they fail. In the Node version, it also was passing the assertions when it shouldn't, due to [this](https://github.com/microsoft/botbuilder-js/compare/main...mdrichardson:main?expand=1#diff-934af35e8c122351d11ac7a4f294230a5fa2a3bf641d17c5d408af46061e48cfL50), which was correct in Dotnet, which [correctly uses `||`](https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive.Testing/TestActions/MemoryAssertions.cs#L62). You can see the issues in the original code if you throw a breakpoint [here](https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-dialogs-adaptive-testing/src/testActions/memoryAssertions.ts#L50)

## Specific Changes

- Added an `OnError` trigger for the PropertyMock test script. It would otherwise silently swallow errors.
- Adjusted what memory properties get written and overwritten.

### DIFF
--- a/libraries/botbuilder-dialogs-adaptive-testing/src/testActions/memoryAssertions.ts
+++ b/libraries/botbuilder-dialogs-adaptive-testing/src/testActions/memoryAssertions.ts
@@ -47,7 +47,7 @@ export class MemoryAssertions extends TestAction implements MemoryAssertionsConf
             await inspector((dc) => {
                 this.assertions.forEach((assertion) => {
                     const { value, error } = Expression.parse(assertion).tryEvaluate(dc.state);
-                    if (error && !value) {
+                    if (error || !value) {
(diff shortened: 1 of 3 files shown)


joshgummersall (approved):

mdrichardson:
@chon219 Mind reviewing this? I'd like somebody from the adaptive side to make sure I'm not crazy, if possible. Note: the CI test failure is unrelated.

chon219 (approved):

PR #3270, opened 2021-02-04
Updated dialogManager to send EoC code like dialogHelper

### LINKED ISSUE #3269: DialogManager does not return EoC code when a dialog ends (JS)
See [parent issue](https://github.com/microsoft/botframework-sdk/issues/6186)

gabog:
Fixes #3269

also added check to the related test

### DIFF
--- a/libraries/botbuilder-dialogs/src/dialogManager.ts
+++ b/libraries/botbuilder-dialogs/src/dialogManager.ts
@@ -10,6 +10,7 @@ import {
     TurnContext,
     BotState,
     ConversationState,
+    EndOfConversationCodes,
     UserState,
     ActivityTypes,
     BotStateSet,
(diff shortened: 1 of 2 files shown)


libraries/botbuilder-dialogs/src/dialogManager.ts
```diff
@@ -324,11 +325,16 @@ export class DialogManager extends Configurable {
...
+            const code =
+                turnResult.status == DialogTurnStatus.complete
+                    ? EndOfConversationCodes.CompletedSuccessfully
+                    : EndOfConversationCodes.UserCancelled;
             // Send End of conversation at the end.
             const activity: Partial<Activity> = {
                 type: ActivityTypes.EndOfConversation,
                 value: turnResult.result,
                 locale: turnContext.activity.locale,
+                code: code,
```
stevengum, line 337: This can be simplified to just `code`:

```suggestion
                code,
```

stevengum (approved):

PR #3361, opened 2021-03-05
Catch premature dialog endings

### LINKED ISSUE #3355: 1 step waterfall dialogs don't resolve back to parent dialogs
## Versions
What package version of the SDK are you using. 4.11
What nodejs version are you using 10.16.4
What browser version are you using n/a
What os are you using n/a

## Describe the bug
Waterfall dialogs that only have 1 step don't resolve back to parent dialogs

## To Reproduce
1. Spin up this [host](https://github.com/microsoft/BotFramework-FunctionalTests/tree/josh/jsTestBranch/Bots/JavaScript/Consumers/CodeFirst/WaterfallHostBot) and [skill](https://github.com/microsoft/BotFramework-FunctionalTests/tree/josh/jsTestBranch/Bots/JavaScript/Skills/CodeFirst/WaterfallSkillBot). I'll leave this branch so it's easier to repo.
Note: You will need app id and password for the host and skill.
2. In Emulator select "normal", "waterfallskill", "waterfallskillbotjs", "8"

## Expected behavior
After the "delete is not supported in emulator" message is received the host dialog doesn't restart.

## Additional context
If you uncomment out the lines in the `deleteDialog.js` file and follow the above steps, and select `no` when prompted it works as expected. After a lot of testing the problem strongly appears to be a bug with an assumption the SDK is making about waterfall dialogs always having more than 1 step. 

If you go through the flow and select the cards option instead of the delete option you can see the expected behavior.

mdrichardson:
Fixes #3355 

## Description

`runDialog()` works great for dialogs that take more than one turn to complete, but it fails otherwise. A single-step or immediately-cancelled dialog should be pretty rare, but this should fix things for them.

The general flow for the way things work currently:

1. `bot.js` file calls `runDialog()` (usually in `onTurn()` or `onMessage()`
2. [`runDialog` sets `result` to `continueDialog()`](https://github.com/mdrichardson/botbuilder-js/blob/dialogHelperCompletions/libraries/botbuilder-dialogs/src/dialogHelper.ts#L87)
3. If no dialogs have started (normal), the `result` will be `DialogTurnStatus.Empty`, so `runDialog` then calls `beginDialog()`.
4. Normally, the turn completes and when the user messages the bot to respond to a dialog prompt, `runDialog()` is called again, so `continueDialog()` is called again. 
    * However, if--after first calling `beginDialog()`--the dialog is cancelled or ended before the turn ends, `result` remains `DialogTurnStatus.Empty` (instead of getting updated to `Completed` or `Cancelled`), so [EoC is never sent back to the parent](https://github.com/mdrichardson/botbuilder-js/blob/dialogHelperCompletions/libraries/botbuilder-dialogs/src/dialogHelper.ts#L92)

## Specific Changes

Allow `result` to be updated with the `beginDialog()` call.

## Testing
Tested using the bots in the linked issue and it worked.


libraries/botbuilder-dialogs/tests/dialogHelper.test.js
```diff
@@ -10,15 +10,9 @@ const {
...
-    ComponentDialog,
-    DialogReason,
-    runDialog,
-    TextPrompt,
-    WaterfallDialog
-} = require('../');
+const { ComponentDialog, DialogReason, runDialog, TextPrompt, WaterfallDialog } = require('../');
 const { AuthConstants } = require('../lib/prompts/skillsHelpers');
-const { EndOfConversationCodes } = require('../../botbuilder-core')
+const { EndOfConversationCodes } = require('../../botbuilder-core');
```
stevengum, line 15: `botbuilder-core` is a dependency

```suggestion
const { EndOfConversationCodes } = require('botbuilder-core');
```

libraries/botbuilder-dialogs/src/dialogHelper.ts
```diff
@@ -84,9 +84,9 @@ export async function runDialog(
     }
 
     // Continue or start the dialog.
-    const result = await dialogContext.continueDialog();
+    let result = await dialogContext.continueDialog();
     if (result.status === DialogTurnStatus.empty) {
-        await dialogContext.beginDialog(dialog.id);
+        result = await dialogContext.beginDialog(dialog.id);
```
stevengum, line 89: Between what you [explained](https://github.com/microsoft/botbuilder-js/pull/3361#issue-585226650) and the [.NET implementation](https://github.com/microsoft/botbuilder-dotnet/blob/046e160829764c4653503b2ccf31acb8b46f2299/libraries/Microsoft.Bot.Builder.Dialogs/DialogExtensions.cs#L73-L76), this fix is spot on. 👍🏼

libraries/botbuilder-dialogs/tests/dialogHelper.test.js
```diff
@@ -131,18 +128,55 @@ class SimpleComponentDialog extends ComponentDialog {
...
         return step.endDialog(step.result);
     }
+}
+
+class ComponentDialogWithPrematureEnd extends ComponentDialog {
+    constructor() {
+        super('ComponentDialogWithPrematureEnd');
+        this.WaterfallDialog = 'WaterfallDialog';
+        this.addDialog(new WaterfallDialog(this.WaterfallDialog, [this.finalStep.bind(this)]));
+        this.initialDialogId = this.WaterfallDialog;
```
stevengum, line 146: Using a locally-scope variable for `'WaterfallDialog'` will suffice for registering the dialog and the `ComponentDialog.initialDialogId`. There are no references to `.WaterfallDialog` outside of the constructors in this file.

libraries/botbuilder-dialogs/tests/dialogHelper.test.js
```diff
@@ -131,18 +128,55 @@ class SimpleComponentDialog extends ComponentDialog {
...
+        return step.endDialog();
+    }
+}
+
+class ComponentDialogWithCancellation extends ComponentDialog {
+    constructor() {
+        super('ComponentDialogWithCancellation');
+        this.WaterfallDialog = 'WaterfallDialog';
+        this.addDialog(new WaterfallDialog(this.WaterfallDialog, [this.finalStep.bind(this)]));
+        this.initialDialogId = this.WaterfallDialog;
```
stevengum, line 165: Same feedback regarding `this.WaterfallDialog` as earlier.

libraries/botbuilder-dialogs/tests/dialogHelper.test.js
```diff
@@ -219,4 +259,27 @@ describe('runDialog()', function() {
...
+    describe('HandlesEarlyDialogEndings', () => {
+        it('handles premature dialog ending', async () => {
+            const dialog = new ComponentDialogWithPrematureEnd();
+            const testFlow = createTestFlow(dialog, FlowTestCase.MiddleSkill);
+            await testFlow.send('Hi').startTest();
+
+            strictEqual(dialog.endReason, DialogReason.endCalled);
+            ok(_eocSent, 'Skills should send EndConversation to channel');
+            strictEqual(_eocSent.type, ActivityTypes.EndOfConversation);
+            strictEqual(_eocSent.code, EndOfConversationCodes.CompletedSuccessfully);
```
stevengum, line 272: These assertions are unnecessary as the scenario is reliant on a Skills scenario and this behavior is tested in the `'HandlesBotAndSkillsTestCases'` collection of tests.

The cancellation test should cover two tests and is fine as is. Thanks!

stevengum (changes requested):
Small changes and cleanup requested, otherwise it looks great! Thanks for the quick turnaround 👍🏼

libraries/botbuilder-dialogs/src/dialogHelper.ts
mdrichardson: Ah, yeah, I probably should've looked there.

libraries/botbuilder-dialogs/tests/dialogHelper.test.js
stevengum: @mdrichardson and I synced offline, we've decided to keep the test as is to ensure the linked issue is fully covered.

stevengum (approved):
🚀

PR #3427, opened 2021-03-23
Replace done w/ async part 4

mdrichardson:

## Covers

* `botbuilder-dialogs`

## Excludes

None

## Description

## Specific Changes

  - `done` => `async/await`

## Notes (mostly for me)


libraries/botbuilder-dialogs/tests/dialogContext.test.js
```diff
@@ -227,203 +243,215 @@ describe('DialogContext', function() {
...
                 assert(err instanceof DialogContextError, 'err should be a DialogContextError');
                 assert(err.dialogContext, 'err should include dialogContext');
-
-                return done();
-            }
-            if (results.status === DialogTurnStatus.empty) {
-                await dc.beginDialog('a');
             }
-            await convoState.saveChanges(turnContext);
         });
```
mdrichardson, line 285: I couldn't think of a way to do `assert.rejects()` here since `dc.continueDialog()` is called multiple times. I debated about whether or not I should add a `errorCaught = false => true` boolean to ensure the `catch` block gets hit. Thoughts?

libraries/botbuilder-dialogs/tests/choicePrompt.test.js
```diff
@@ -405,78 +477,38 @@ describe('ChoicePrompt', function () {
...
+                            inlineSeparator: PromptCultureModels.English.separator,
+                        }).text;
+                        assert.strictEqual(activity.text, `Please choose a color.${expectedChoices}`);
                     })
                     .startTest();
-            }));
-        }));      
-    });
-
-    it('should default to english locale', async function () {
```
mdrichardson, line 434: Note: This test wasn't deleted; it's the one above. The diff just looks weird.

libraries/botbuilder-dialogs/tests/dialogContext.test.js
mdrichardson: [resolved](https://github.com/microsoft/botbuilder-js/pull/3427/commits/8d5fc3a5c6557127a383269a4681c80b69bc66d0)

joshgummersall (approved):

PR #3436, opened 2021-03-24
Replace done w/ async part 5 (final PR for this)

### LINKED ISSUE #3226: Use async tests instead of done(), where possible

mdrichardson:
Note: Even though this basically only covers `botbuilder-connector`, this is the last PR on this issue and everything that can be easily converted now has been.

Fixes #3226

## Covers

* `botbuilder-dialogs-adaptive` through end, inclusive

## Excludes

* mocha before/after hooks in `connector.test.js`
* All tests in `botframework-streaming`. Many rely on callbacks and don't work when converted.

## Description

## Specific Changes

  - `done` => `async/await`


libraries/botbuilder-testing/src/dialogTestClient.ts
```diff
@@ -134,7 +134,7 @@ export class DialogTestClient {
      * Send an activity into the dialog.
      *
      * ```javascript
-     * DialogTest.send('hello').assertReply('hello yourself').then(done);
+     * await DialogTest.send('hello').assertReply('hello yourself').startTest();
```
mdrichardson, line 137: Note: Figured I may as well make the comments show the "right" way.

libraries/botframework-connector/tests/connector.test.js
```diff
@@ -44,551 +47,459 @@ const BotConnector = require('../lib');
...
-                    })
-                    .then(done, done);
+
+                const result = await client.conversations.createConversation(params);
+                const members = await client.conversations.getActivityMembers(result.id, result.activityId);
+                members.should.matchAny((member) => member.id === user.id);
             });
-            it('should fail with invalid conversatoinId', function(done) {
-                var params = createConversation();
+            it('should fail with invalid conversationId', async function () {
```
mdrichardson, line 262: Note: Renaming this test required this file to be renamed, below: `Bot_Framework_Connector_SDK_Conversations_GetActivityMembers_should_fail_with_invalid_conversationId.nock.js`

joshgummersall (approved):

PR #3532, opened 2021-04-06
 Reduce usage of the 'any' type in adaptive-expressions [Part 1]

Danieladu:
## Description
`any/object` type in typescript is abused in some situations.
This is the first part of issue #3203. Reduce the usage except "BuiltinFunctions" folder

## Specific Changes
Replace `any` with other alternatives.


Danieladu:
Will re-open it while some conflicts are solved.

PR #3653, opened 2021-05-05
chore: Replace `catch` with `assert.rejects|throws` in tests (part 2)

mdrichardson:
Partially addresses: #3426
#minor

## Description

We has some awkward usage of

```js
try {
  methodThatShouldFail()
  assert.fail('not whatever')
} catch (err) {
}
```

That I converted to

```js
assert.throws(
	() => methodThatShouldFail(),
	Error('whatever')
);
```

# Note

There's a commit with a bunch of no-op linting fixes. I recommend mostly reviewing the other one.


libraries/botbuilder-ai/tests/luisRecognizer.test.js
```diff
@@ -490,119 +490,90 @@ describe('LuisRecognizer', function () {
...
-        try {
-            await recognizer.recognize(context);
-            assert.fail('should have thrown');
-        } catch (error) {
-            const expectedError = `Response 400: The request's body or parameters are incorrect, meaning they are missing, malformed, or too large.`;
-            assert(error.message === expectedError, `unexpected error message thrown.`);
-        }
+        await assert.rejects(
+            recognizer.recognize(context),
+            Error(
```
joshgummersall, line 495: How is the diff when this doesn't match? I've noticed using `{message: "Expected message here"}` produces really useful diffs, just want to make sure that's the case with this as well. also, should really be `new Error`

libraries/botbuilder-core/tests/storageBaseTests.js
```diff
@@ -171,31 +175,16 @@ class StorageBaseTests {
...
-            assert.notStrictEqual(reloadedStoreItem4, null);
-
-            reloadedStoreItem4.eTag = '';
-            const dict2 = { pocoStoreItem: reloadedStoreItem4 };
-
-            await storage.write(dict2);
-
-            assert.fail('Should have thrown exception on write with storeItem because of empty eTag');
-        } catch (err) { }
-
```
mdrichardson, line 188: Note: I removed this test because:

1. If you read it closely, you can see it will always pass. This is my fault, as I wrote the test, originally.
2. This test merges many *Storage tests into one, ensuring they all pass. As it turns out, [the old tests](https://github.com/microsoft/botbuilder-js/blob/4c01e2fd7bd8febba8a40028bc78dd862d50ec87/libraries/botbuilder-core/tests/memoryStorage.test.js) didn't have a test for this and [would've failed anyway](https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-core/src/memoryStorage.ts#L80). I believe the test in this file (that this PR removed) actually comes from one I found in .NET, orginally.

libraries/botbuilder-core/tests/browserStorage.test.js
```diff
@@ -1,127 +1,112 @@
...
-                });
-            });
+    it('invalid eTag', async function () {
+        await storage.write({ keyUpdate2: { count: 1 } });
+
+        const result = await storage.read(['keyUpdate2']);
+        result.keyUpdate2.count = 2;
+        result.keyUpdate2.eTag = -1;
+
+        await assert.rejects(storage.write(result), Error('Storage: error writing "keyUpdate2" due to eTag conflict.'));
```
mdrichardson, line 39: Note: This is another test that would always pass when it shouldn't. As you can see, I changed `result.keyUpdate2.eTag` intentionally to something that was different than it's previous value. [eTags of the same value are NOT invalid](https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-core/src/memoryStorage.ts#L84). However, an new eTag < the old eTag IS invalid. [This is handled the same way in .NET.](https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Builder/MemoryStorage.cs#L155)

libraries/botbuilder-ai/tests/luisRecognizer.test.js
mdrichardson: It's good!

image

> also, should really be new Error

I actually looked this one up, first because I wasn't completely sure. Per [the spec](https://tc39.es/ecma262/#sec-error-constructor) they're the same. However, as far as Bot Framework "standard practices" go, I'll go ahead and change it, if you prefer.

joshgummersall: Nice! Useful diff is all I desire, neither syntax is decidedly better IMO. So this is good!

joshgummersall (approved):

PR #3661, opened 2021-05-06
chore: Replace `catch` with `assert.rejects|throws` in tests (part 3 of 3)

mdrichardson:
Fixes: #3426

Note: This is part 3 of 3, so the issue is done after this merges. There's some tests that weren't converted, for various reasons, detailed in each PR.

#minor

## Description

We has some awkward usage of

```js
try {
  methodThatShouldFail()
  assert.fail('not whatever')
} catch (err) {
}
```

That I converted to

```js
assert.throws(
	() => methodThatShouldFail(),
	Error('whatever')
);
```

# Note

There's a commit with a bunch of no-op linting fixes. I recommend mostly reviewing the other one. However, I accidentally committed some no-op changes to `skillDialog.test.js`. sorry.

I skipped a LOT of tests in `botframework-streaming` because:

1. They rely on `chai`, which doesn't handle `rejects()` without `chai-as-promised`
2. The nature of the promises in this library don't work well with `rejects()` and are handled better/easier with the existing try/catch


libraries/botbuilder-dialogs/tests/skillDialog.test.js
```diff
@@ -18,125 +18,143 @@ const { spy, stub } = require('sinon');
...
-        if (BOTBUILDER !== null && BOTBUILDER_TESTING !== null) {
-            const { DialogTestClient } = BOTBUILDER_TESTING;
-
-            /**
-             * @remarks
-             * Port of BeginDialogShouldCallSkill from C#
-             * https://github.com/microsoft/botbuilder-dotnet/blob/41120728b22c709ec2d9247393505fdc778b2de1/tests/Microsoft.Bot.Builder.Dialogs.Tests/SkillDialogTests.cs#L47-L49
-             * @param {string} deliveryMode 
-             */
-            async function beginDialogShouldCallSkill(deliveryMode, useCreateSkillConversationId = false) {
```
mdrichardson, line 59: Note: This is difficult to read in the diff, but I moved this function definition outside of the `if` statement.

libraries/botbuilder-dialogs/tests/skillDialog.test.js
```diff
@@ -194,202 +212,197 @@ describe('SkillDialog', function() {
...
-                dialog.validateBeginDialogArgs({});
-            } catch (e) {
-                typeErrorValidator(e, '"activity" is undefined or null in options.');
-            }
+        it('should fail if options is falsy', function () {
+            const dialog = new SkillDialog({}, 'SkillDialog');
+            assert.throws(() => dialog.validateBeginDialogArgs(), new TypeError('Missing options parameter'));
         });
 
-        it('should fail if dialogArgs.activity.type is not "message" or "event"', () => {
```
mdrichardson, line 295: Note: This method was removed. When "fixed" so that it throws when expected, it doesn't actually throw. This is because [it doesn't validate within the activity](https://github.com/microsoft/botbuilder-js/blob/77d8fb7893881b3072b3c148cb256409955e8b96/libraries/botbuilder-dialogs/src/skillDialog.ts#L231) and [neither does .NET](https://github.com/microsoft/botbuilder-dotnet/blob/395c8bd1365dc042600ac8fe8546afc6f6eafaf4/libraries/Microsoft.Bot.Builder.Dialogs/SkillDialog.cs#L218)

libraries/botframework-config/tests/loadAndSave.test.js
```diff
@@ -1,212 +1,204 @@
...
         await config.saveAs(saveBotPath, secret);
 
-        try {
-            await bf.BotConfiguration.load(saveBotPath);
-            assert.fail("Load should have thrown due to no secret");
-        }
-        catch (Error) { }
-    });
-
-    if ("CantSaveWithoutSecret", async () => {
```
mdrichardson, line 105: Note: This was deleted because:

1. This whole package/library is functionally deprecated
2. [It won't throw anyway](https://github.com/microsoft/botbuilder-js/blob/main/libraries/botframework-config/src/botConfiguration.ts#L145
)

libraries/botframework-streaming/tests/NodeWebSocket.test.js
```diff
@@ -3,74 +3,68 @@ const { expect } = require('chai');
...
-    it('creates a new NodeWebSocket', () => {
-        const socket = new NodeWebSocket(new FauxSock);
+describe('NodeWebSocket', function () {
+    it('creates a new NodeWebSocket', function () {
+        const socket = new NodeWebSocket(new FauxSock());
         expect(socket).to.be.instanceOf(NodeWebSocket);
         expect(socket.close()).to.not.throw;
     });
 
-    it('requires a valid URL', () => {
```
mdrichardson, line 13: Note: This test was deleted because once "fixed" so that it waits for an expected error, the test doesn't actually throw one. Per @stevengum, and based on my own search, this error doesn't actually exist so the test is fine to delete.

libraries/botbuilder-dialogs/tests/skillDialog.test.js
stevengum: Thanks for adding the note!

stevengum (approved):

PR #3778, opened 2021-06-18
fix: return full response w/ adaptiveCard invoke action

### LINKED ISSUE #3776: Adaptive Card Invoke Activity should return the response instead of response value
## Versions
What package version of the SDK are you using. 4.13.5
What nodejs version are you using - 14.7
What browser version are you using - Chrome
What os are you using - Linuxmint

## Describe the bug
When trying to respond to an adaptiveCard invoke activity using `onAdaptiveCardInvoke` method, returning a string response using the `value` field results in Teams bot responding with a generic _Something went wrong_ error message.

After working with Azure Support and Teams engineering teams over the past two weeks, we found that the bot should respond with `AdaptiveCardInvokeResponse` in the response body.

There is an implementation mismatch between the Node.js and C# versions. As referenced from these two [lines](https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-core/src/activityHandler.ts#L501-L502), the current implementation returns the invoke response value instead of the invokeResponse. This leads to the `statusCode` and `type` values being repeated in order to satisfy the `AdaptiveCardInvokeResponse` interface.

The unit tests described at [Activity Handler Unit Test](https://github.com/microsoft/botbuilder-js/blob/2bc15c67beabd841e02f5f08915f7cc27467f314/libraries/botbuilder-core/tests/ActivityHandler.test.js#L295) and [Teams Activity Handler Unit Test](https://github.com/microsoft/botbuilder-js/blob/2bc15c67beabd841e02f5f08915f7cc27467f314/libraries/botbuilder/tests/teamsActivityHandler.test.js#L276) seems to use the string value instead of `AdaptiveCardInvokeResponse` which fails in actual implementation.

## To Reproduce
1. Send AC invoke card with `Action.Execute` to the user.
2. Implement `onAdaptiveCardInvoke` in your bot which extends `ActivityHandler` or `TeamsActivityHandler`.
3. Respond with a text/AC response. The AC described in the next step/image fails.
4. image

## Expected behavior
The `onAdaptiveCardInvoke` method should be coded as described below which repeats `statusCode` and `type` values multiple times. The return value at [onInvokeActivity](https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-core/src/activityHandler.ts#L502) should be updated to the following.

```diff
- return { status: response.statusCode, body: response.value };
+ return { status: response.statusCode, body: response };
```

image

mdrichardson:
Fixes #3776

### DIFF
--- a/libraries/botbuilder-core/src/activityHandler.ts
+++ b/libraries/botbuilder-core/src/activityHandler.ts
@@ -499,7 +499,7 @@ export class ActivityHandler extends ActivityHandlerBase {
                 case 'adaptiveCard/action': {
                     const invokeValue = this.getAdaptiveCardInvokeValue(context.activity);
                     const response = await this.onAdaptiveCardInvoke(context, invokeValue);
-                    return { status: response.statusCode, body: response.value };
+                    return { status: response.statusCode, body: response };
                 }
 
                 case verifyStateOperationName:
@@ -694,12 +694,22 @@ export class ActivityHandler extends ActivityHandlerBase {
     }
 
     private getAdaptiveCardInvokeValue(activity: Activity): AdaptiveCardInvokeValue {
-        const { value } = activity;
+        const { value }: { value?: AdaptiveCardInvokeValue } = activity;
         if (!value) {
             const response = this.createAdaptiveCardInvokeErrorResponse(
                 StatusCodes.BAD_REQUEST,
                 'BadRequest',
                 'Missing value property'
+            );
+
+            throw new InvokeException(StatusCodes.BAD_REQUEST, response);
+        }
+
+        if (value.action.type !== 'Action.Execute') {
+            const response = this.createAdaptiveCardInvokeErrorResponse(
+                StatusCodes.BAD_REQUEST,
+                'NotSupported',
+                `The action '${value.action.type}' is not supported.`
             );
 
             throw new InvokeException(StatusCodes.BAD_REQUEST, response);
--- a/libraries/botbuilder-core/tests/ActivityHandler.test.js
+++ b/libraries/botbuilder-core/tests/ActivityHandler.test.js
@@ -298,7 +298,15 @@ describe('ActivityHandler', function () {
         };
 
         await processActivity(
-            { type: ActivityTypes.Invoke, name: 'adaptiveCard/action', value: 'Action.Execute' },
+            {
+                type: ActivityTypes.Invoke,
+                name: 'adaptiveCard/action',
+                value: {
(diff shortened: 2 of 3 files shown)


libraries/botbuilder-core/src/activityHandler.ts
```diff
@@ -499,7 +499,7 @@ export class ActivityHandler extends ActivityHandlerBase {
                 case 'adaptiveCard/action': {
                     const invokeValue = this.getAdaptiveCardInvokeValue(context.activity);
                     const response = await this.onAdaptiveCardInvoke(context, invokeValue);
-                    return { status: response.statusCode, body: response.value };
+                    return { status: response.statusCode, body: response };
```
mdrichardson, line 502: Note: This is part that actually addresses the issue.

libraries/botbuilder-core/src/activityHandler.ts
```diff
@@ -705,6 +705,16 @@ export class ActivityHandler extends ActivityHandlerBase {
...
+        if (value.action.type !== 'Action.Execute') {
+            const response = this.createAdaptiveCardInvokeErrorResponse(
+                StatusCodes.BAD_REQUEST,
+                'NotSupported',
+                `The action '${value.action.type}' is not supported.`
+            );
+
+            throw new InvokeException(StatusCodes.BAD_REQUEST, response);
+        }
+
```
mdrichardson, line 717: Note: This was just to sync w/ .NET

libraries/botbuilder-core/src/activityHandler.ts
```diff
@@ -694,7 +694,7 @@ export class ActivityHandler extends ActivityHandlerBase {
     }
 
     private getAdaptiveCardInvokeValue(activity: Activity): AdaptiveCardInvokeValue {
-        const { value } = activity;
+        const { value }: { value?: AdaptiveCardInvokeValue } = activity;
```
joshgummersall, line 697: Is this explicit type necessary? I would have thought it was inferred?

joshgummersall (approved):
One small question, otherwise LGTM.

libraries/botbuilder-core/src/activityHandler.ts
mdrichardson: Necessary I think because `activity.value` can be anything.

Note: Had to use optional `value?` because otherwise eslint complains that it can't cast because `activity.value` is optional.

image
image

joshgummersall (approved):

PR #3779, opened 2021-06-18
fix: first activity from bot to user has fake replyToId value

### LINKED ISSUE #3732: port: Fix first activity from bot to user has fake replyToId value (#5313)
The changes in [Fix first activity from bot to user has fake replyToId value (#5313)](https://github.com/microsoft/botbuilder-dotnet/pull/5313) may need to be ported to maintain parity with `microsoft/botbuilder-dotnet`.

<blockquote>
Fixes #5310 

## Description
When the activity is the first one of the conversation and it's sent from bot to user, the ReplyToId was assigned to the bogus ConversationUpdate activity's id. The fix is to not assign the bogus ReplyToId if previous activity is the ConversationUpdate activity

## Specific Changes
  - Do not assign ReplyToId if conversantionReference is from ConversationUpdate activity
  - Add test tests to test the changes

## Testing
Tests are added
</blockquote>

mdrichardson:
Fixes #3732

## Description
When the activity is the first one of the conversation and it's sent from bot to user, the ReplyToId was assigned to the bogus ConversationUpdate activity's id. The fix is to not assign the bogus ReplyToId if previous activity is the ConversationUpdate activity

## Specific Changes
  - Do not assign ReplyToId if conversantionReference is from ConversationUpdate activity
  - Add test tests to test the changes

## Testing
Tests are added

### DIFF
--- a/libraries/botbuilder-core/src/turnContext.ts
+++ b/libraries/botbuilder-core/src/turnContext.ts
@@ -10,6 +10,7 @@ import {
     InputHints,
     ResourceResponse,
     Mention,
+    Channels,
 } from 'botframework-schema';
 import { INVOKE_RESPONSE_KEY } from '.';
 import { BotAdapter } from './botAdapter';
@@ -118,6 +119,17 @@ export type DeleteActivityHandler = (
 ) => Promise<void>;
 
 export const BotCallbackHandlerKey = 'botCallbackHandler';
+
+function getAppropriateReplyToId(source: Partial<Activity>): string | undefined {
+    if (
+        source.type !== ActivityTypes.ConversationUpdate ||
+        (source.channelId !== Channels.Directline && source.channelId !== Channels.Webchat)
+    ) {
+        return source.id;
+    }
+
+    return undefined;
+}
 
 // tslint:disable-next-line:no-empty-interface
 export interface TurnContext {}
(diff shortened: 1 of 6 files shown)


libraries/botbuilder-core/src/turnContext.ts
```diff
@@ -119,6 +120,17 @@ export type DeleteActivityHandler = (
...
+    if (
+        source.type !== ActivityTypes.ConversationUpdate ||
+        (source.channelId !== Channels.Directline && source.channelId !== Channels.Webchat)
+    ) {
+        return source.id;
+    }
+
+    return undefined;
+}
+
```
mdrichardson, line 133: Note: I debated about what to do with this function, which is also located in `ActivityEx`. Since `TurnContext` and `ActivityEx` are in two separate libraries, I copy/pasted the function to keep exposed surface area lower. I wanted a function just to avoid the harder-to-read ternary (used in the .NET PR). So, this isn't DRY, but is DRY-er than copy/pasting the ternary around, I think.

I think this *could* be removed from here, entirely, though. In .NET, TurnContext doesn't have a `getConversationReference` and instead relies on the one in `ActivityEx`. I *think* it might be worth removing here as "Find all References" only returns it being used in 6 different places. However, this is core enough to the SDK and a little beyond the scope of this PR, so I was more hesitant to mess with it.

yarn.lock
```diff
@@ -10876,7 +10876,7 @@ run-queue@^1.0.0, run-queue@^1.0.3:
   dependencies:
     aproba "^1.1.1"
 
-runtypes@6.3.0, runtypes@~6.3.0:
+runtypes@~6.3.0:
```
mdrichardson, line 10879: Note: This is due to accidentally forgetting to commit in previously-merged PR.

joshgummersall (approved):
Fair points regarding the duplicated function. For now, I'd say it's fine.

PR #3796, opened 2021-06-23
fix: handle reactions added/removed separately

### LINKED ISSUE #3795: JS Activity Handler doesn't handle messageReactions for teams correctly
## Versions
Prod
14.16.0
n/a
What os are you using
windows

## Describe the bug
There isn't parity between JS and C# for how messageReaction events are handled. When a user swaps 1 message reaction for another reaction Teams will send 1 activity and populate the reactionAdded and reactionRemoved properties on the activity. JS doesn't handle the dispatching of this scenario. 

## To Reproduce
1. Add a JS bot that overrides reactionAdded and reactionRemoved events for teams
2. send the bot a message
3. add a reaction to the response from the bot
4. swap the reaction for a different reaction

## Expected behavior
The reactionAdded and reactionRemoved events should both be fired

## Additional context
This issue was raised [here](https://github.com/microsoft/botframework-sdk/issues/5972), but JS hasn't implemented that fix. Here's a [link](https://github.com/microsoft/botbuilder-dotnet/blob/ee4f5f074302af9c24d498025c45a6ae809ef1eb/libraries/Microsoft.Bot.Builder/ActivityHandler.cs#L273) to how .NET implements it.

mdrichardson:
Fixes #3795 

## Description
Simple change for .NET parity. We just needed to allow reactions `added` and `removed` to _both_ be handled.

## Testing
image

### DIFF
--- a/libraries/botbuilder-core/src/activityHandlerBase.ts
+++ b/libraries/botbuilder-core/src/activityHandlerBase.ts
@@ -155,9 +155,11 @@ export class ActivityHandlerBase {
      *   call [onReactionsRemovedActivity](xref:botbuilder-core.ActivityHandlerBase.onReactionsRemovedActivity).
      */
     protected async onMessageReactionActivity(context: TurnContext): Promise<void> {
-        if (context.activity.reactionsAdded && context.activity.reactionsAdded.length > 0) {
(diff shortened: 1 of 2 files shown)


joshgummersall (approved):
Nice! Thank you.

PR #3868, opened 2021-07-15
feat: allow sending trace activities to clients

### LINKED ISSUE #3865: Feature request: Allow sending trace messages to client
Currently, both the old `BotFrameworkAdapter` and the new `CloudAdapter` prevent trace activities from being sent to the client. There are some use-cases where this might be useful though (e.g., certain test scenarios).
Would it be possible to add an option to the `CloudAdapterBase` class that would allow trace messages to be sent?
It could be a simple boolean or a function that determines per request if traces should be forwarded.

Happy to suggest a PR myself but wanted to know if it would even have a chance getting merged.

https://github.com/microsoft/botbuilder-js/blob/7afa288668abd45b680bc0cd8bc7a58f3e276f34/libraries/botbuilder-core/src/cloudAdapterBase.ts#L75

alopix:
Fixes #3865

## Description
Allows sending trace activities to the client even if it is not the official botframework emulator.

## Specific Changes
  - CloudAdapter(Base) has a new flag that allows enabling trace activities being sent to the client all the time.

## Testing
Added tests to check the flag actually being used.

### DIFF
--- a/libraries/botbuilder-core/etc/botbuilder-core.api.md
+++ b/libraries/botbuilder-core/etc/botbuilder-core.api.md
@@ -292,7 +292,9 @@ export class CardFactory {
 
 // @public (undocumented)
 export abstract class CloudAdapterBase extends BotAdapter {
-    constructor(botFrameworkAuthentication: BotFrameworkAuthentication);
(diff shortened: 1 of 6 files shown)


gabog:
Thanks for the PR @alopix, but trace activities are only supported on the emulator channel, we can't take this PR

PR #3920, opened 2021-09-03
chore: switch to new xmldom and fast-xml-parser

### LINKED ISSUE #3916: CVE-2021-32796 Package: xmldom
adaptive-expressions uses **xmldom** 0.5.0 which has CVE-2021-32796. It should be migrated to @xmldom/xmldom

vasicvuk:
Fixes #3916

## Description

Since **x2js** did not respond to PR I have changed the library to **fast-xml-parser** which is far more maintained and then I have updated xmldom to latest @xmldom/xmldom

## Specific Changes

  - In package.json I have changed xmldom to latest @xmldom/xmldom which has newer versions
  - In package.json I have changed **x2js** with **fast-xml-parser** which is more maintained project for conversion. 
  - In xml.ts i have changed x2js to xml-js
  - In xpath.ts i have changed xmldom to @xmldom/xmldom

## Testing
All existing tests are passing as no functionality is changed


vasicvuk:
I also have removed the xml2js as fast-xml-parser works both with Node and Browser

joshgummersall:
For posterity, [`fast-xml-parser`](https://www.npmjs.com/package/fast-xml-parser) gets roughly 90x the weekly downloads as [`x2js`](https://www.npmjs.com/package/x2js) on NPM.

@vasicvuk can you confirm that this package works well in the browser? Running `yarn build` inside the `adaptive-expressions` package should produce a file `lib/browser.js` which _should_ load and run in a browser.

joshgummersall:
Looks like `@types/xmldom` can be removed (as per `depcheck` results).

vasicvuk:
Sure, i will remove the typings, and check in the browser ones i get to PC later

vasicvuk:
Hi @joshgummersall,

I removed typings and also tested if the changed worked in browser.

I used following HTML for testing:
```html
<html>
<head>
	<script src="lib/browser.js"></script>
</head>
<body>
	<script>
		const scope = {};
		var expression = AEL.Expression.parse('xml(\'{"person": {"name": "Sophia Owen", "city": "Seattle"}}\')'.toString());
		const value = expression.tryEvaluate(scope);
		console.log(value);
	</script>
</body>

</html>
```

As a result i got this:
image

which confirms that it works in browser too.

joshgummersall:
@vasicvuk, please remove `package-lock.json` and re-run `yarn` within the repo to update `yarn.lock`.

vasicvuk:
I can remove the package lock but there is no yarn.lock in libraries/adaptive-expressions. Root folder yarn.lock will remain same.

joshgummersall:
@vasicvuk, yep that's expected! One `yarn.lock` covers the whole repo.

libraries/adaptive-expressions/src/builtinFunctions/xml.ts
```diff
@@ -28,48 +28,24 @@ export class XML extends ExpressionEvaluator {
...
-            } catch (err) {
-                error = `${args[0]} is not a valid json`;
-            }
-
-            return { value: result, error: error };
+            const parser = new j2xParser({
+                indentBy: '  ',
+                format: true,
+            });
+            result = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>\n' + parser.parse(obj).trim('\n');
```
joshgummersall, line 44: We try to use string interpolation when possible (works with multiline strings). Try the following:

```typescript
result = `<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
  ${parser.parse(obj)}`.trim('\n');
```

vasicvuk: Sure I will change it. Also argument \n for trim is not needed. As trim automatically removes new lines at the end.

vasicvuk:
It seems that build of whole project cannot pass without  **@types/xmldom**

joshgummersall:
It looks like `@xmldom/xmldom` provides types internally: https://github.com/xmldom/xmldom/blob/master/package.json#L21, so `@types/xmldom` should be unnecessary. There must be something else going on. I'll clone your fork later today and take a closer look.

vasicvuk:
Okey. I just noticed that build passes when i return it and i get same error when i dont have it

joshgummersall:
Huh - well, I suppose you can go ahead and add the types back. You'll have to add `@types/xmldom` to the `depcheck` ignore list in `adaptive-expressions` (see [example here](https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-dialogs-adaptive-runtime-core/package.json#L40)).

.depcheckrc
```diff
@@ -1,2 +1,2 @@
-ignores: ["mocha", "rimraf", "sinon"]
+ignores: ["mocha", "rimraf", "sinon", "@types/xmldom"]
```
joshgummersall, line 1: This should only go in the `adaptive-expressions` `package.json` file:
```json
{
  ...,
  "scripts": {
    ...,
    "depcheck": "depcheck ... --ignores @types/xmldom",
  },
  ...
}
```

vasicvuk:
I have added it again and included it in **.depcheckrc** file since **--ignores** flag seems to be ignored if **.depcheckrc** file provided. Also I have run depcheck to verify it and run build on clean repository and everything passes.

.depcheckrc
vasicvuk: It gives an error if the depcheckrc file is provided? I can do it but then I must remove calling of depcheckrc file from the call

vasicvuk: Okey my error but now i get
```
> depcheck --config ../../.depcheckrc --ignores @types/xmldom

Missing dependencies
* sinon: .\tests\expressionParser.test.js
```

vasicvuk: I added sinon and remove @types/xmldoc from depcheckrc. Now it should be fine

vasicvuk:
Fixed it now, had to add also sinon as dependency as it is used within tests. I had this error:
```
npm run depcheck

> adaptive-expressions@4.1.6 depcheck D:\Projects\botbuilder-js\libraries\adaptive-expressions
> depcheck --config ../../.depcheckrc --ignores @types/xmldom

Missing dependencies
* sinon: .\tests\expressionParser.test.js
```

joshgummersall:
Please remove the added `sinon` dependency and add `--ignores sinon,@types/xmldom` for `adaptive-expressions`. It's inherited from the root `package.json` file in the repo. `depcheck` isn't perfect and doesn't understand hoisting very well. Thanks!

vasicvuk:
@joshgummersall  done

joshgummersall:
@cosmicshuai can you take a look at this as well? Just want a second set of eyes and you've worked on `adaptive-expressions` a lot in the past.

vasicvuk:
@joshgummersall New version of x2js is also released. So I can also now provide a PR with less changes, but I think it is a better option to change it to **fast-xml-parser** as it has more Contributors, Stars on GitHub, Downloads, it is used in other Microsoft OSS and other large companies too.

joshgummersall:
Agreed, I was hoping to get another set of eyes on the change but I’m inclined to move forward either way. I’ll land this today either way.

joshgummersall (approved):

mogadanez:
fast-xml-parser have issue now too
https://security.snyk.io/vuln/SNYK-JS-FASTXMLPARSER-3325616

PR #3923, opened 2021-09-07
port:[#3906] Port Managed Identity (MSI) + Single Tenant from DotNet

### LINKED ISSUE #3906: port: Managed Identity (MSI) + Single Tenant support for Bot apps (#5829)
The changes in [Managed Identity (MSI) + Single Tenant support for Bot apps (#5829)](https://github.com/microsoft/botbuilder-dotnet/pull/5829) may need to be ported to maintain parity with `microsoft/botbuilder-dotnet`.

<blockquote>
Fixes #5830

## Description
- Enable Azure's Managed Identity (MSI) for bot apps, so that bot authors don't have to manage app secrets. 
- Support for hosting bots as single tenant apps, which enables acquiring AAD access tokens from the bot's host tenant, as opposed to, `botframework.com` tenant.

Below are the 3 ways a bot app can now be hosted:
#### 1. Multi-Tenant (Default):
```json
{
  "MicrosoftAppId": "",
  "MicrosoftAppPassword": ""
}
```
#### 2. Single Tenant:
```json
{
  "MicrosoftAppType": "SingleTenant",
  "MicrosoftAppId": "",
  "MicrosoftAppPassword": "",
  "MicrosoftAppTenantId": ""
}
```
#### 3. Managed Identity (MSI):
```json
{
  "MicrosoftAppType": "UserAssignedMSI",
  "MicrosoftAppId": "",
  "MicrosoftAppTenantId": ""
}
```

## Specific Changes
  - Added classes for Managed Identity credentials management and token acquisition.
  - Updated `ConfigurationServiceClientCredentialFactory` to instantiate the appropriate credential mechanism (MultiTenant/SingleTenant/MSI) based on appsettings.
  - Added the bot's host tenant as a valid JWT token issuer since the tokens will be issued from the host tenant when using SingleTenant/MSI hosting model. This was done by adding a property to the `AuthenticationConfiguration` class which would be used in the skills scenarios.
  - Added an overload in `PasswordServiceClientCredentialsFactory` to accept TenantId which can be used to acquire token using the bot's host tenant for SingleTenant scenario.

## Testing
Following scenarios were verified:
#### 1. Bot to Channel scenarios:
- MultiTenant
- SingleTenant
- MSI
#### 2. Bot to bot (skills) scenarios:
<table>
    <tr><td></td><td>MultiTenant Skill</td><td>SingleTenant Skill</td><td>MSI Skill</td></tr>
    <tr><td>MultiTenant Host</td><td>Y</td><td>-</td><td>-</td></tr>
    <tr><td>SingleTenant Host</td><td>-</td><td>Y</td><td>Y</td></tr>
    <tr><td>MSI Host</td><td>-</td><td>Y</td><td>Y</td></tr>
</table>
NOTE: When one of the bots is MultiTenant, the other cannot be SingleTenant/MSI since the token acquisition by the MultiTenant bot will fail due to not being able to find the SingleTenant/MSI bot resource registered in the `botframework.com` tenant. So, those scenarios are not supported.
</blockquote>

sw-joelmut:
Fixes #3906

## Description
This PR ports the Managed Identity (MSI) and Single Tenant functionalities from DotNet.

> **Note:** While testing the different combinations of the following table, we found that the case where a SingleTenant host connects a MultiTenant skill through `expectReplies` delivery mode works, the skill answered to the host properly, whereas using `normal` it didn't, due to not finding the resource in `botframework.com` (which is not supported, more information can be found in the related issue).
<table>
    <tr><td></td><td>MultiTenant Skill</td><td>SingleTenant Skill</td><td>MSI Skill</td></tr>
    <tr><td>MultiTenant Host</td><td>Y</td><td>-</td><td>-</td></tr>
    <tr><td>SingleTenant Host</td><td>normal (-), expectReplies (Y)</td><td>Y</td><td>Y</td></tr>
    <tr><td>MSI Host</td><td>-</td><td>Y</td><td>Y</td></tr>
</table>

> NOTE: When one of the bots is MultiTenant, the other cannot be SingleTenant/MSI since the token acquisition by the MultiTenant bot will fail due to not being able to find the SingleTenant/MSI bot resource registered in the `botframework.com` tenant. So, those scenarios are not supported. 

## Specific Changes
- Updated `ConfigurationServiceClientCredentialFactory` to support the three MultiTenant, SingleTenant and MSI functionalities.
- Added valid token issuers into the `botbuilder-dialogs-adaptive-runtime` and `botframework-connector` libraries.
- Added new `JwtTokenProviderFactory` class that will be used to instantiate the `DefaultAzureCredential` class from `@azure/identity`.
- Added `ManagedIdentityAppCredentials` class by extending from `AppCredentials` and be able to get the jwt token.
- Added `ManagedIdentityAuthenticator` that will be in charge of creating the token provider and gather the token from it.
- Added `ManagedIdentityServiceClientCredentialsFactory` class Factory in charge of creating the `ManagedIdentityAppCredentials`.
- Added new unit tests for previous specified classes.
- Added new/updated dependencies (package.json)
  - `botframework-connector`
    - `@azure/identity@2.0.0-beta.6`
  - `root`
    - `@azure/logger@1.0.2`
    - `@azure/ms-rest-js@1.9.1`

## Testing
The following image shows some new added unit tests passing successfully.
image

## Troubleshooting
- `@azure/identity`
  - When using the latest available version in NPM (`1.5.2`) we started getting `Unexpected character...` error on the `keytar` dependency when running `yarn build`.
  - We found an issue related to `keytar` which will not be included in the base package in the `2.0.0` release.
    - Issue: `https://github.com/Azure/azure-sdk-for-js/issues/14852#issuecomment-818915410`
  - **Solution:** Instead of installing the version `1.5.2` we use `2.0.0-beta.6`  and no issues were shown, and the build was successful.
  image

- `@azure/logger`
  - Package dependency of the `@azure/identity` library.
  - Browserify fails to build the project due to not finding the `@azure/logger` library in the `@azure/identity`. This is caused by the browser configuration property set in the  `@azure/logger` library.
  - This issues was already solved in the newest versions, from which `@azure/identity` wasn't installing.
  - **Solution:** By including the `@azure/logger` into the root `package.json` and setting the minimum version of `^1.0.2` solved the issue when building with browserify.
   image


package.json
```diff
@@ -40,7 +40,8 @@
     "update-versions": "yarn workspace botbuilder-repo-utils update-versions"
   },
   "devDependencies": {
-    "@azure/ms-rest-js": "1.9.0",
+    "@azure/logger": "^1.0.2",
```
joshgummersall, line 43: Where is this used? This should likely go in a `package.json` file within the package (i.e. `libraries/X`) it's used in.

package.json
```diff
@@ -40,7 +40,8 @@
...
+    "@azure/ms-rest-js": "1.9.1",
```
joshgummersall, line 44: This also should likely be reconciled with whatever package(s) it's used in. Why was this updated?

libraries/botbuilder-core/src/configurationServiceClientCredentialFactory.ts
```diff
@@ -28,14 +62,93 @@ export type ConfigurationServiceClientCredentialFactoryOptions = z.infer<typeof
...
+            MicrosoftAppType = null,
+            MicrosoftAppTenantId = null,
+        } = TypedConfig.nonstrict().parse(factoryOptions);
+        super(MicrosoftAppId, MicrosoftAppPassword, MicrosoftAppTenantId);
+
+        const appType = MicrosoftAppTypes[MicrosoftAppType] ?? MicrosoftAppTypes.MultiTenant;
+
+        switch (appType) {
+            case MicrosoftAppTypes.UserAssignedMsi:
+                if (!MicrosoftAppId || MicrosoftAppId.trim() === '') {
```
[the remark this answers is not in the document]
MartinLuccanera: Hey Josh, I have 2 questions.
1) This is not a test, I guess that copying the assert part was a mistake, right?
2) `MicrosoftAppId?.trim()` is not equivalent to `!MicrosoftAppId || MicrosoftAppId.trim() === ''`. The first option would only be true `if trim() === ''`, whereas the second case would be true if `trim() === ''` OR if `MicrosoftAppId is false(ish)`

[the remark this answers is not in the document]
joshgummersall: `assert` is a node built-in that just makes it easier to throw errors, it's a convention we use throughout the JS SDK and is more common in Typescript projects.

package.json
MartinLuccanera: Regarding this issue, here's the error we get when building without this dependency
image

Identity doesn't have a newer version that makes use of logger 1.0.2 and version 1.0.0 is the problem.

We also tried to use nohoist, but botFramework-connector is not private, so that config was ignored.

Do you have an alternative we can use for this?

package.json
MartinLuccanera: For consistency with other uses of the same dependency. We can revert it if you deem it so

package.json
MartinLuccanera: I also forgot to mention that adding logger 1.0.2 to connector's package.json had no effect

libraries/botframework-connector/src/auth/parameterizedBotFrameworkAuthentication.ts
```diff
@@ -227,6 +227,13 @@ export class ParameterizedBotFrameworkAuthentication extends BotFrameworkAuthent
         authHeader: string,
         channelId: string
     ): Promise<ClaimsIdentity> {
+        // Add allowed token issuers from configuration (if present)
+        if (this.authConfiguration.validTokenIssuers?.length > 0) {
+            const validIssuers = Array.from(ToBotFromBotOrEmulatorTokenValidationParameters.issuer);
```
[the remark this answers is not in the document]
MartinLuccanera: We restructured this portion of code and removed Array.from

libraries/botframework-connector/src/auth/jwtTokenProviderFactory.ts
```diff
@@ -8,6 +8,7 @@
 
 import { DefaultAzureCredential } from '@azure/identity';
 import { JwtTokenProviderFactoryInterface } from './jwtTokenProviderFactoryInterface';
+import * as assert from 'assert';
```
joshgummersall, line 11: `import {ok} from 'assert'` and then call `ok()` which is the same as `import * as assert from 'assert'; assert(...)`

libraries/botframework-connector/src/auth/passwordServiceClientCredentialFactory.ts
```diff
@@ -68,45 +68,22 @@ export class PasswordServiceClientCredentialFactory implements ServiceClientCred
...
-                    : new PrivateCloudAppCredentials(
-                          appId,
-                          this.password,
-                          this.tenantId,
-                          audience,
-                          normalizedEndpoint,
-                          validateAuthority
-                      );
+            credentials = new PrivateCloudAppCredentials(
+                appId,
```
joshgummersall, line 79: If `appId` is `undefined`, are `password` and `tenantId` also undefined? Along with `audience`? I'm wary of this change.

libraries/botbuilder-core/package.json
```diff
@@ -27,6 +27,7 @@
     }
   },
   "dependencies": {
+    "@azure/ms-rest-js": "1.9.1",
```
joshgummersall, line 30: I don't think this package actually imports anything from `@azure/ms-rest-js` right?

MartinLuccanera: If I remove that, I get this error while building
image

libraries/botframework-connector/src/auth/passwordServiceClientCredentialFactory.ts
MartinLuccanera: This change comes straight from porting the dotnet version, you can check it [here ](https://github.com/microsoft/botbuilder-dotnet/pull/5829/files?file-filters%5B%5D=.cs#diff-3bff06c07579e50f9e2f978a5b221eafc7f1f5ac5dff2af5eac6302a2637ed30). Besides that, this portion of code would not be reached because of the `if` on line 62, right?

libraries/botbuilder-core/etc/botbuilder-core.api.md
```diff
@@ -33,6 +33,7 @@ import { O365ConnectorCard } from 'botframework-schema';
 import { PasswordServiceClientCredentialFactory } from 'botframework-connector';
 import { ReceiptCard } from 'botframework-schema';
 import { ResourceResponse } from 'botframework-schema';
+import type { ServiceClientCredentials } from '@azure/ms-rest-js';
```
joshgummersall, line 36: This type should be exported from `botframework-connector` so we don't have to introduce `@azure/ms-rest-js` as a dependency in `botbuilder-core`.

libraries/botframework-connector/src/auth/managedIdentityAppCredentials.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        this.tokenProviderFactory = tokenProviderFactory;
+        super.appId = appId;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected async refreshToken(): Promise<TokenResponse> {
+        this.authenticator ??= new ManagedIdentityAuthenticator(this.appId, this.oAuthScope, this.tokenProviderFactory);
```
joshgummersall, line 43: Why not create this in the `constructor`?

libraries/botframework-connector/src/auth/passwordServiceClientCredentialFactory.ts
```diff
@@ -31,9 +39,10 @@ export class PasswordServiceClientCredentialFactory implements ServiceClientCred
         };
     }
 
-    constructor(appId: string, password: string) {
+    constructor(appId: string, password: string, tenantId: string) {
```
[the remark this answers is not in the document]
joshgummersall: How was this resolved?

libraries/botframework-connector/src/auth/passwordServiceClientCredentialFactory.ts
joshgummersall: Ah fair - didn't realize the .NET piece was refactored.

libraries/botframework-connector/src/auth/skillValidation.ts
```diff
@@ -136,6 +136,13 @@ export namespace SkillValidation {
             ? GovernmentConstants.ToBotFromEmulatorOpenIdMetadataUrl
             : AuthenticationConstants.ToBotFromEmulatorOpenIdMetadataUrl;
 
+        // Add allowed token issuers from configuration (if present)
+        if (authConfig.validTokenIssuers?.length > 0) {
+            ToBotFromBotOrEmulatorTokenValidationParameters.issuer = [
+                ...ToBotFromBotOrEmulatorTokenValidationParameters.issuer,
+                ...this.authConfiguration.validTokenIssuers,
```
joshgummersall, line 143: It seems like this list will continue to grow as long as a bot runs since we're pushing entries into a static array. Is that a concern?

libraries/botframework-connector/src/auth/parameterizedBotFrameworkAuthentication.ts
```diff
@@ -304,6 +312,14 @@ export class ParameterizedBotFrameworkAuthentication extends BotFrameworkAuthent
         authHeader: string,
         channelId: string
     ): Promise<ClaimsIdentity> {
+        // Add allowed token issuers from configuration (if present)
+        if (this.authConfiguration.validTokenIssuers?.length > 0) {
+            ToBotFromBotOrEmulatorTokenValidationParameters.issuer = [
```
joshgummersall, line 317: It seems like this list will continue to grow as long as a bot runs since we're pushing entries into a static array. Is that a concern?

libraries/botframework-connector/src/auth/managedIdentityServiceClientCredentialsFactory.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Initializes a new instance of the ManagedIdentityServiceClientCredentialsFactory class.
+     *
+     * @param appId Client ID for the managed identity assigned to the bot.
+     * @param tokenProviderFactory The JWT token provider factory to use.
+     */
+    constructor(appId: string, tokenProviderFactory: JwtTokenProviderFactoryInterface) {
+        super();
+        if (!appId || appId.trim() === '') {
```
joshgummersall, line 29: Use the `ok()` assert pattern here too.

libraries/botbuilder-core/src/configurationServiceClientCredentialFactory.ts
```diff
@@ -28,14 +50,87 @@ export type ConfigurationServiceClientCredentialFactoryOptions = z.infer<typeof
...
      *
      * @param factoryOptions A [ConfigurationServiceClientCredentialFactoryOptions](xref:botbuilder-core.ConfigurationServiceClientCredentialFactoryOptions) object.
      */
     constructor(factoryOptions: ConfigurationServiceClientCredentialFactoryOptions = {}) {
-        const { MicrosoftAppId = null, MicrosoftAppPassword = null } = TypedConfig.nonstrict().parse(factoryOptions);
-        super(MicrosoftAppId, MicrosoftAppPassword);
+        // Exclude MicrosoftAppType from Zod to ignore-casing comparison.
+        // .NET code:
+        // https://github.com/microsoft/botbuilder-dotnet/blob/d84c6a1f76a56dbee0d18a16adf5d678e5b30035/libraries/integration/Microsoft.Bot.Builder.Integration.AspNet.Core/ConfigurationServiceClientCredentialFactory.cs#L53-L55
+        const { MicrosoftAppType, ...options } = factoryOptions;
```
joshgummersall, line 63: If we aren't enforcing case-sensitivity, there's no reason to make this an `enum` IMO. Sorry for the bad advice before! Just make the type above a `z.string()` and then compare against a set of `const`s. The point of `zod` is that it provides good error messages by default, and if we aren't going to force users to use the correct casing in the JSON file there's no point.

libraries/botbuilder-core/package.json
joshgummersall: I'll push a quick commit fixing the type export and dependencies, it's a little subtle so easier to show than explain :)

libraries/botframework-connector/src/auth/managedIdentityAuthenticator.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Acquires the security token.
+     *
+     * @returns {Promise<AccessToken>} An [AccessToken](xref:botframework-connector.AccessToken).
+     */
+    public async getToken(): Promise<AccessToken> {
+        // Retry gradually, starting from 10 ms up to 5 times.
+        return retry(() => this.tokenProvider.getToken(this.resource), 5, 10);
```
joshgummersall, line 44: Nice use of `retry`!

libraries/botframework-connector/tests/auth/managedIdentityAuthenticator.test.js
```diff
@@ -0,0 +1,117 @@
...
+                getToken() {
+                    callsToAcquireToken++;
+                    throw new Error('AuthError');
+                },
+            }));
+
+            const sut = new ManagedIdentityAuthenticator(testAppId, testAudience, tokenProvider);
+
+            try {
+                await sut.getToken();
```
joshgummersall, line 109: I think these tests are better written with `await assert.rejects()`, like so:

```typescript
await assert.rejects(sut.getToken(), {
  message: 'AuthError'
});
assert.strictEqual(...);
```

This code won't actually fail if no error is thrown, which it seems like it should.

joshgummersall (commented):
Little more feedback here, it's looking solid though. I pulled the branch and understand the root `package.json` changes better now, thanks for your patience while I convinced myself those changes were necessary.

package.json
joshgummersall: I saw your issue, some problem with `browserify`. Easier to take this change than figure out why the bundle process fails.

libraries/botframework-connector/package.json
```diff
@@ -27,6 +27,7 @@
     }
   },
   "dependencies": {
+    "@azure/identity": "2.0.0-beta.5",
```
joshgummersall, line 30: We try to not use `beta` packages - can we use the latest non-beta version on NPM in the `1.x.x` version?

sw-joelmut: Our intention was to use the `1.5.2` version, but we encountered an error with `keytar` when building the project. We found an issue saying that `keytar` will be an optional dependency in `2.0.0`. Therefore, we updated the `@azure/identity` to the `2.0.0-beta` version to be able to build the project.
More information can be found in the **Troubleshooting** section on the PR description.

libraries/botframework-connector/src/auth/parameterizedBotFrameworkAuthentication.ts
sw-joelmut: We updated how the issuers were assigned, we defined a shallow clone object assigning new `issuers` without altering the `ToBotFromBotOrEmulatorTokenValidationParameters` constant object.

libraries/botframework-connector/src/auth/skillValidation.ts
sw-joelmut: We updated how the issuers were assigned, we defined a shallow clone object assigning new `issuers` without altering the `ToBotFromBotOrEmulatorTokenValidationParameters` constant object.

sw-joelmut:
@joshgummersall Thank you for your feedback, we just pushed some changes, let us know if we can improve anything else 🤗(hug).

libraries/botframework-connector/src/auth/passwordServiceClientCredentialFactory.ts
stevengum: Yes, this was code I didn't have time to 100% test and clean up when we shipped the JS, Python and Java implementations of CloudAdapter. Without an AppId, it's impossible to fetch a token from AAD which is why the check on line 62 exists and an [issue was filed in .NET](https://github.com/microsoft/botbuilder-dotnet/issues/5591#issuecomment-864167043). Giving the bot a `MicrosoftAppCredentials.Empty` instance is functionally equivalent to giving it an empty `PrivateCloudAppCredentials` or a [`MicrosoftGovernmentAppCredentials.Empty`](https://github.com/microsoft/botbuilder-dotnet/pull/5730/files#diff-3bff06c07579e50f9e2f978a5b221eafc7f1f5ac5dff2af5eac6302a2637ed30L90) in .NET.

stevengum (commented):
The Identity team released [`@azure/identity@2.0.0-beta.6`](https://www.npmjs.com/package/@azure/identity/v/2.0.0-beta.6) which contains a few breaking changes that I don't believe affects the changes in this PR nor the rest of the `botframework-connector` package. That said, would you bump @azure/identity to this current `next` version and vet the contents of this PR again?

https://github.com/Azure/azure-sdk-for-js/blob/4868986e7f311d64478492c1571e59903ccbd3a1/sdk/identity/identity/CHANGELOG.md

libraries/botbuilder-core/src/configurationServiceClientCredentialFactory.ts
```diff
@@ -28,14 +49,77 @@ export type ConfigurationServiceClientCredentialFactoryOptions = z.infer<typeof
  * ServiceClientCredentialsFactory that uses a [ConfigurationServiceClientCredentialFactoryOptions](xref:botbuilder-core.ConfigurationServiceClientCredentialFactoryOptions) or a [Configuration](xref:botbuilder-dialogs-adaptive-runtime-core.Configuration) instance to build ServiceClientCredentials with an AppId and App Password.
  */
 export class ConfigurationServiceClientCredentialFactory extends PasswordServiceClientCredentialFactory {
+    private readonly inner: ServiceClientCredentialsFactory;
```
stevengum, line 52: Ouch, this is unfortunate but it happens 😞

sw-joelmut:
> 
> 
> The Identity team released [`@azure/identity@2.0.0-beta.6`](https://www.npmjs.com/package/@azure/identity/v/2.0.0-beta.6) which contains a few breaking changes that I don't believe affects the changes in this PR nor the rest of the `botframework-connector` package. That said, would you bump @azure/identity to this current `next` version and vet the contents of this PR again?
> 
> https://github.com/Azure/azure-sdk-for-js/blob/4868986e7f311d64478492c1571e59903ccbd3a1/sdk/identity/identity/CHANGELOG.md

Updated the `@azure/identity` to the latest `2.0.0-beta.6` and the functionality continues to work.
image

joshgummersall (commented):
One final set of feedback

libraries/botframework-connector/src/auth/authenticationConstants.ts
```diff
@@ -83,14 +83,34 @@ export namespace AuthenticationConstants {
...
+
+    /**
+     * The Government V1 Azure AD token issuer URL template that will contain the tenant id where the token was issued from.
+     */
+    export const ValidGovernmentTokenIssuerUrlTemplateV1 = 'https://login.microsoftonline.us/';
+
+    /**
+     * The Government V2 Azure AD token issuer URL template that will contain the tenant id where the token was issued from.
+     */
+    export const ValidGovernmentTokenIssuerUrlTemplateV2 = 'https://login.microsoftonline.us/';
```
stevengum, line 104: Shouldn't this be in the [`GovernmentConstants`](https://github.com/microsoft/botbuilder-js/blob/eb58d15b67405fe11a9bcdf958ff5a241feced97/libraries/botframework-connector/src/auth/governmentConstants.ts) instead? @msomanathan 

It's not a big deal, but historically all of the US Gov Cloud related constants are stored in a separate file. My main concern is the v4-to-v5 work and making sure our ducks are in a row.

I did note that this is a verbatim port from [.NET](https://github.com/microsoft/botbuilder-dotnet/pull/5829/files#diff-7f25611536bab084126e2ba4ea408532cc78ac0b18534e36365ab3c7c7cd71b1R88-R96).

libraries/botframework-connector/src/auth/jwtTokenProviderFactoryInterface.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { DefaultAzureCredential } from '@azure/identity';
+
+/*
+ * A factory that can create OAuth token providers for generating JWT auth tokens.
+ */
+export interface JwtTokenProviderFactoryInterface {
```
stevengum, line 14: @joshgummersall should we instead export a same-name interface from jwtTokenProviderFactory.ts instead of having this `JwtTokenProviderFactoryInterface` and have TypeScript [merge](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces) it for consumers instead?

I realize my aversion is primarily due to the name, but still...

Sub-optimal example in our code base: [TurnContext](https://github.com/microsoft/botbuilder-js/blob/eb58d15b67405fe11a9bcdf958ff5a241feced97/libraries/botbuilder-core/src/turnContext.ts#L137-L146)

stevengum: Also, the access level for this interface does not match [.NET](https://github.com/microsoft/botbuilder-dotnet/pull/5829/files#diff-49d5a695d7c466c9b4a44b89aae393b51fec3905e77be373ae5f99f7944177c9R12), which is public. To match .NET, the interface needs to be exported from `src/auth/index.ts`.

libraries/botframework-connector/src/auth/jwtTokenProviderFactoryInterface.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @module botframework-connector
+ */
+/**
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import { DefaultAzureCredential } from '@azure/identity';
```
stevengum, line 9:
```suggestion
import type { DefaultAzureCredential } from '@azure/identity';

```

libraries/botframework-connector/src/auth/managedIdentityAppCredentials.ts
```diff
@@ -0,0 +1,53 @@
...
+import { TokenResponse } from 'adal-node';
+import { AppCredentials } from './appCredentials';
+import { JwtTokenProviderFactoryInterface } from './jwtTokenProviderFactoryInterface';
+import { ManagedIdentityAuthenticator } from './managedIdentityAuthenticator';
+import { ok } from 'assert';
+
+/**
+ * Managed Service Identity auth implementation.
+ */
+export class ManagedIdentityAppCredentials extends AppCredentials {
```
stevengum, line 18: The access level for this class doesn't match [.NET](https://github.com/microsoft/botbuilder-dotnet/pull/5829/files#diff-5b38e347048b007edd6393cc510e83ce6fc883d48c38b1ff6262edc62a21985eR14); to do so, the class needs to be exported from `src/auth/index.ts`.

libraries/botframework-connector/src/auth/managedIdentityAuthenticator.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Initializes a new instance of the ManagedIdentityAuthenticator class.
+     *
+     * @param appId Client id for the managed identity to be used for acquiring tokens.
+     * @param resource Resource for which to acquire the token.
+     * @param tokenProviderFactory The JWT token provider factory to use.
+     */
+    constructor(appId: string, resource: string, tokenProviderFactory: JwtTokenProviderFactoryInterface) {
+        ok(appId?.trim(), 'ManagedIdentityAuthenticator.constructor(): missing appid.');
```
stevengum, line 29:
```suggestion
        ok(appId?.trim(), 'ManagedIdentityAuthenticator.constructor(): missing appId.');

```

libraries/botframework-connector/src/auth/managedIdentityAuthenticator.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export class ManagedIdentityAuthenticator {
+    private readonly tokenProvider: DefaultAzureCredential;
+    private readonly resource: string;
+
+    /**
+     * Initializes a new instance of the ManagedIdentityAuthenticator class.
+     *
+     * @param appId Client id for the managed identity to be used for acquiring tokens.
+     * @param resource Resource for which to acquire the token.
```
stevengum, line 25: @msomanathan is there a more descriptive docstring we can use here? Or include `<remarks/>` that indicate what a `resource` might be?

libraries/botframework-connector/tests/auth/jwtTokenProviderFactory.test.js
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+const assert = require('assert');
+const { JwtTokenProviderFactory } = require('../../lib/auth/jwtTokenProviderFactory');
```
stevengum, line 5:
```suggestion
const { JwtTokenProviderFactory } = require('../../lib');

```

libraries/botframework-connector/tests/auth/managedIdentityAppCredentials.test.js
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+const assert = require('assert');
+const sinon = require('sinon');
+const { JwtTokenProviderFactory } = require('../../lib/auth/jwtTokenProviderFactory');
+const { ManagedIdentityAppCredentials } = require('../../lib/auth/managedIdentityAppCredentials');
+const { AppCredentials } = require('../../lib/auth/appCredentials');
+const { AuthenticationConstants } = require('../../lib/auth/authenticationConstants');
```
stevengum, line 9:
```suggestion
const { AppCredentials, AuthenticationConstants, JwtTokenProviderFactory, ManagedIdentityAppCredentials } = require('../../lib');
```

Note: this will only work once `ManagedIdentityAppCredentials` are publicly accessible.

libraries/botframework-connector/src/auth/managedIdentityAuthenticator.ts
```diff
@@ -0,0 +1,46 @@
...
+        ok(tokenProviderFactory, 'ManagedIdentityAuthenticator.constructor(): missing tokenProviderFactory.');
+
+        this.resource = resource;
+        this.tokenProvider = tokenProviderFactory.createAzureServiceTokenProvider(appId);
+    }
+
+    /**
+     * Acquires the security token.
+     *
+     * @returns {Promise<AccessToken>} An [AccessToken](xref:botframework-connector.AccessToken).
```
stevengum, line 40: This is incorrect, the `xref` is for `@azure/identity`, though I'm not sure if the `xref` would work in our official docs. @JonathanFingold do you know if we could cross link this way and have a working link?

libraries/botframework-connector/src/auth/managedIdentityServiceClientCredentialsFactory.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Initializes a new instance of the ManagedIdentityServiceClientCredentialsFactory class.
+     *
+     * @param appId Client ID for the managed identity assigned to the bot.
+     * @param tokenProviderFactory The JWT token provider factory to use.
+     */
+    constructor(appId: string, tokenProviderFactory: JwtTokenProviderFactoryInterface) {
+        super();
+        ok(appId?.trim(), 'ManagedIdentityServiceClientCredentialsFactory.constructor(): missing appid.');
```
stevengum, line 30:
```suggestion
        ok(appId?.trim(), 'ManagedIdentityServiceClientCredentialsFactory.constructor(): missing appId.');

```

libraries/botframework-connector/src/auth/managedIdentityAuthenticator.ts
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public async getToken(): Promise<AccessToken> {
```
stevengum, line 42: The default access modifier is `public`, thus we've stopped being explicit by using `public`. Please remove from `public` this method and the other applicable methods in this PR.

libraries/botframework-connector/src/auth/managedIdentityServiceClientCredentialsFactory.ts
```diff
@@ -0,0 +1,66 @@
...
+        );
+
+        this.appId = appId;
+        this.tokenProviderFactory = tokenProviderFactory;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public async isValidAppId(appId: string): Promise<boolean> {
```
stevengum, line 43: Please remove the unnecessary `public` access modifier.

libraries/botframework-connector/src/auth/passwordServiceClientCredentialFactory.ts
```diff
@@ -59,43 +88,23 @@ export class PasswordServiceClientCredentialFactory implements ServiceClientCred
...
         let credentials: MicrosoftAppCredentials;
-        let normalizedEndpoint = loginEndpoint?.toLowerCase();
+        const normalizedEndpoint = loginEndpoint?.toLowerCase();
+
         if (normalizedEndpoint?.startsWith(AuthenticationConstants.ToChannelFromBotLoginUrlPrefix)) {
-            // TODO: Unpack necessity of these empty credentials based on the loginEndpoint as no tokens are fetched when auth is disabled.
-            credentials =
-                appId == null
-                    ? MicrosoftAppCredentials.Empty
-                    : new MicrosoftAppCredentials(appId, this.password, undefined, audience);
```
stevengum, line 68: Thanks @sw-joelmut for cleaning this up! ❤️

libraries/botframework-connector/tests/auth/managedIdentityAuthenticator.test.js
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+const assert = require('assert');
+const sinon = require('sinon');
+const { ManagedIdentityAuthenticator } = require('../../lib/auth/managedIdentityAuthenticator');
+const { JwtTokenProviderFactory } = require('../../lib/auth/jwtTokenProviderFactory');
```
stevengum, line 7: Please clean up these `require`.

libraries/botframework-connector/tests/auth/managedIdentityAuthenticator.test.js
```diff
@@ -0,0 +1,114 @@
...
+        it('should throw with null tokenProviderFactory', function () {
+            assert.throws(() => new ManagedIdentityAuthenticator(testAppId, testAudience, null), {
+                name: 'AssertionError',
+                message: errorMsgToken,
+            });
+        });
+    });
+
+    describe('getToken', function () {
+        it('should get token', async function () {
```
stevengum, line 66: I think simplifying this test suite by collapsing the `describe('getToken')` and `it('...')` would create a better dev experience for the maintainers. Example:

```ts
it('getToken() should get token', async function () {
```

Please apply this where easily applicable and use [simpleCredentialProvider.test.js](https://github.com/microsoft/botbuilder-js/blob/9c9f8d95927ea2f035af97ba53ddaa73e0231289/libraries/botframework-connector/tests/simpleCredentialProvider.test.js#L15-L21) as a reference point.

libraries/botframework-connector/tests/auth/managedIdentityServiceClientCredentialsFactory.test.js
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+const assert = require('assert');
+const {
+    ManagedIdentityServiceClientCredentialsFactory,
+} = require('../../lib/auth/managedIdentityServiceClientCredentialsFactory');
+const { JwtTokenProviderFactory } = require('../../lib/auth/jwtTokenProviderFactory');
```
stevengum, line 8: Please clean up these `require`.

libraries/botframework-connector/tests/auth/managedIdentityServiceClientCredentialsFactory.test.js
```diff
@@ -0,0 +1,81 @@
...
+            assert.strictEqual(false, await sut.isValidAppId('invalidAppId'));
+        });
+    });
+
+    describe('IsAuthenticationDisabled', function () {
+        it('should not be disabled', async function () {
+            const sut = new ManagedIdentityServiceClientCredentialsFactory(testAppId, testTokenProvider);
+            assert.strictEqual(false, await sut.isAuthenticationDisabled());
+        });
+    });
```
stevengum, line 64: Please see my earlier remarks for collapsing the `describe()` and `it()` together. [passwordServiceClientCredentialFactory.test.js](https://github.com/microsoft/botbuilder-js/blob/9c9f8d95927ea2f035af97ba53ddaa73e0231289/libraries/botframework-connector/tests/passwordServiceClientCredentialFactory.test.js#L17-L28) is a good reference point.

libraries/botframework-connector/tests/auth/managedIdentityServiceClientCredentialsFactory.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+    describe('CreateCredentials', function () {
+        it('should create credentials', async function () {
+            const sut = new ManagedIdentityServiceClientCredentialsFactory(testAppId, testTokenProvider);
+            const credentials = await sut.createCredentials(testAppId, testAudience);
+            assert.notStrictEqual(null, credentials);
```
stevengum, line 70:
```suggestion
            assert(credentials instanceof ManagedIdentityAppCredentials);
```

stevengum (changes requested):

sw-joelmut:
@stevengum Thank you for the feedback, we just pushed the changes.
There are some additional changes applied based on the provided feedback, let us know if there is anything else we could improve on the implementation. Thanks.

libraries/botframework-connector/src/auth/jwtTokenProviderFactoryInterface.ts
joshgummersall: I’m not a fan of merging identifiers, it’s not obvious or explicit enough in my opinion. If anything, I think the name should be `IJwtTokenProviderFactory` as we have used that pattern elsewhere in the codebase.

libraries/botframework-connector/src/auth/managedIdentityAuthenticator.ts
joshgummersall: While not ideal, this comment is ported from .NET so it must be acceptable 😄

joshgummersall (approved):
Approved, pending test pass!

libraries/botframework-connector/src/auth/jwtTokenProviderFactoryInterface.ts
joshgummersall: Fixed in last two commits I pushed.

libraries/botframework-connector/src/auth/authenticationConstants.ts
joshgummersall: Filed https://github.com/microsoft/botframework-sdk/issues/6384 for tracking.

libraries/botframework-connector/src/auth/jwtTokenProviderFactory.ts
```diff
@@ -8,12 +8,23 @@
 
 import { DefaultAzureCredential } from '@azure/identity';
 import { ok } from 'assert';
-import type { JwtTokenProviderFactoryInterface } from './jwtTokenProviderFactoryInterface';
+
+/*
+ * A factory that can create OAuth token providers for generating JWT auth tokens.
+ */
+export interface IJwtTokenProviderFactory {
```
stevengum, line 15: Thanks @joshgummersall!

stevengum (approved):

PR #3929, opened 2021-09-17
chore: bump adal-node for xmldom CVE

joshgummersall:
`adal-node` published `0.2.3` which moves to `@xmldom/xmldom` as a CVE resolution.

#minor


mdrichardson:
I can't seem to start a Review, but LGTM, unless we also want to bump it here: https://github.com/microsoft/botbuilder-js/blob/6fab548c8ade620a7eb3d8905729ff6cd0649237/tools/framework/suite-base.js

But that's only used for tests, so might not matter.

joshgummersall:
Yeah, not a concern IMO as we don't publish that package anywhere.

stevengum (approved):

PR #3964, opened 2021-10-27
feat: Merge recognizerOptions from object and context.prompt call

### LINKED ISSUE #3945: Allow recognizerOptions when calling DialogContext.prompt
**Is your feature request related to a problem?**

I need to set `recognizeNumbers` and `recognizeOrdinals` in options when calling `stepContext.prompt(id, options)` in a step of a `WaterfallDialog`.

When calling `DialogContext.prompt(id: string, options: PromptOptions)` method, it ends calling `ChoicePrompt.onRecognize` with the `options` param, but then it only uses `options.choices`, and ignore the rest of the properties when calling `const results = recognizeChoices(utterance, choices, opt);` [here](https://github.com/microsoft/botbuilder-js/blob/4.14/libraries/botbuilder-dialogs/src/prompts/choicePrompt.ts#L153), as it uses `const opt = this.recognizerOptions || {};` [here](https://github.com/microsoft/botbuilder-js/blob/4.14/libraries/botbuilder-dialogs/src/prompts/choicePrompt.ts#L151).

When calling `ChoicePrompt.onRecognize(context: TurnContext, state: any, options: PromptOptions)`, those `options` could be merged to `this.recognizerOptions` when assigning to `opt`, something like:

```typescript
const opt = Object.assign({}, options.recognizerOptions, this.recognizerOptions);
```

Being `options.recognizerOptions` a new property in `PromptOptions` type.

On my dialog with a choice prompt, I could:

```typescript
const myPrompt = new ChoicePrompt(id);
myPrompt.recognizerOptions = {
  recognizeNumbers: false,
  recognizeOrdinals: false
};
super.addDialog(myPrompt);
```
But this is not as flexible, as the object will have the same values in every call to `stepContext.prompt(id, options)`, as those options cannot contain recognizer options.
On the other hand, those options are not stored in dialog state (within dialog stack) to be accessed if required, as it is our case.

**Additional context**

Using botbuilder 4.13.3

greuze:
Fixes #3945

## Description
Allow to merge `recognizerOptions` (such as `recognizeNumbers`, `recognizeOrdinals`, etc.) from the prompt object with the options in `context.prompt` call, as the latter were only used to get the choices, but didn't allow to set options for the recognizer.

## Specific Changes
  - Allow to merge `recognizerOptions` in the prompt object with `recognizerOptions` in the `context.prompt` call.
  - Added `recognizerOptions` property to `PromptOptions` interface, as it is required by previous point.

## Testing
I added two choice prompt tests, one with default recognizerOptions (with `recognizeNumbers: true`) and other overriding this recognizerOptions in `context.prompt` call, the first understand numbers, and the second doesn't, as expected.


joshgummersall:
Hey @greuze, thank you for the contribution here. We try to avoid introducing features like this in one language only. If you are open to introducing the feature in github.com/microsoft/botbuilder-dotnet as well it may gain traction more quickly. Either way, the team will have to review the change to ensure it makes sense in all the SDKs, not just the Javascript SDK. Apologies for the delay!

greuze:
Hi @joshgummersall, thank you for your response, we are working on the dotnet version.

I don't get the problem with the unsuccessfull check `test-compat.yml / test:compat`, could anyone help us with this?

mdrichardson:
@greuze, `test:compat` ensures that our API docs are updated with the latest changes to the code. The error message says,

> botbuilder-dialogs
 | Warning: You have changed the public API signature for this project. Please copy the file "temp/botbuilder-dialogs.api.md" to "etc/botbuilder-dialogs.api.md", or perform a local build (which does this automatically). See the Git repo documentation for more info.
 
 So, you need to go into `../botbuilder-dialogs/temp` and copy the `botbuilder-dialogs.api.md` file to `../botbuilder-dialogs/etc`, then re-run `test:compat`, ensure it passes, and commit the file.

greuze:
Telefonica `main` branch got messed after updating with latest Microsoft `main` branch, and the PR was automatically closed after force-pushing the latest from Microsoft remote to Telefonica remote, as they both were equals.

After this, I cherry-picked the original commit of the modification (`9a27e2ad6c1e3e620fccd51d858ecc117ebbd44a`) and now the PR is correct again.

I run `test:compat` locally, and it worked properly, without modifying `botbuilder-dialogs/etc/botbuilder-dialogs.api.md` (so I don't have anything to commit, as @mdrichardson explained). That file is the same as the one in `botbuilder-dialogs/temp/botbuilder-dialogs.api.md` (but changing the end-of-line sequence from LF to CRLF). In any case, I think I didn't change the public API signature, could the problem be related to those LF vs CRLF?

I would appreciate that anyone could approve the workflows again, to check that everything goes OK this time.

greuze:
Please, I need help with this, `test:compat` keeps failing, but I don't know why, in local it completes successfully, and as far as I know, I didn't change the public API signature.

EricDahlvang:
Hi @greuze 

I've run test:compat on your fork https://github.com/Telefonica/botbuilder-js and it is failing for me also. If you take the file from libraries\botbuilder-dialogs\temp 
image
and move it to libraries\botbuilder-dialogs\etc
image

it will pass. Then commit this file with the one line change:
image

greuze:
Thank you @EricDahlvang, I did what you suggested, and it worked, I commited the new file.

joshgummersall (approved):

greuze:
Hello, is there any news about this? I got an approval, but I need a code owner review, and every time I merge with master, I require also a new workflow approval :confused:

gabog:
Hi @greuze, thank you for the PR. 

We reviewed this with the SDK team and we can't take the change as is. `PromptOptions` applies to all `Prompt<T>` classes:  `AttachmentPrompt`, `ChoicePrompt`, `ConfirmPrompt`, `DateTimePrompt` and `NumberPrompt<T>`.  Each one of these types have different properties that are relevant to the type of prompt being used.

image

The PR suggests adding a `FindChoicesOptions` property to `PromptOptions` that is only relevant to `ChoicePrompt` and that wouldn't be correct since this property would be just noise for the other prompts.

In the current implementation, Prompt properties can only be set on the prompt instance that is shared by all the conversations that use the prompt. The only thing that can currently be overridden is PromptOptions. 

Adding support for changing the other properties on `stepContext.Prompt()` calls would require changes on all the prompts (not just `ChoicePrompt`) and it would be hard to do without introducing breaking changes or adding global properties that are not relevant to all the prompts (there may be other options but they are not obvious or trivial). 

As an alternative consider creating your own ChoicePrompt that implements the desired behavior or create several ChoicePrompt instances with different FindChoiceOptions settings and call the one you need when invoking stepContext.Prompt(desiredPromptId) in your waterfall step.

Closing this PR for now, I have some follow up questions that I will post in the main issue so we can understand your need a bit better and consider this feature for a future version.

PR #3972, opened 2021-11-04
fix: [#3969] useLanguagePolicy setting turn state, instead of languagePolicy

### LINKED ISSUE #3969: MissingProperties custom function has introduced logic that breaks Core Language bot scenarios
## Versions
`4.15.0-rc0.preview`

## Describe the bug
Core Language bot is hung when making LUIS calls and is unable to proceed. Something is up with consultation mode when it's invoking child dialogs.

## To Reproduce
1. Create a Core Language (JS) bot in Composer
2. Update the `botbuilder-dialogs-adaptive-runtime-integration-express` reference to `4.15.0-rc0.preview`

## Expected behavior
When saying "help" or "cancel" but should reply as expected, but instead it doesn't respond.

## Additional context
Reverting https://github.com/microsoft/botbuilder-js/commit/3e2beae24e616c89075532a6385e2399f2bc8500 appears to resolve the problem.

sw-joelmut:
Reverts #3970

Fixes #3969

Fix: [commit](https://github.com/microsoft/botbuilder-js/commit/52d282246ff903f81fcf551c6a2082f0652c6a3d)

## Description
This PR fixes an issue where the bot wasn't able to respond to messages after the Welcome message is sent.
When the `AdaptiveDialog.beginDialog` method is executed, the `context.turnState` for the second message (`e.g.: help`) was the same as the first message (`WelcomeMessage`), thus, having the `activityProcessed` property set to `true`, the bot didn't process the new Activity and ended the Dialog right away.

## Specific Changes
- Revert PR #3970
- Updates the `LanguageGeneratorExtensions.useLanguagePolicy` method to set the correct LanguagePolicy like in DotNet ([link](https://github.com/microsoft/botbuilder-dotnet/blob/406d58042e49737b8738416d35bc7fbb47deccc9/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive/LanguageGeneratorExtensions.cs#L83-L84)) instead of setting the `turnState` object with the `languagePolicy`.
  - Change: `dialogManager.initialTurnState.set(TurnPath.languagePolicy, policy);`

## Testing
The following images show what was causing the issue, the DialogContext using the same turnState as the message sent before, and the bot working after the fix.
image

### DIFF
--- a/libraries/botbuilder-dialogs-adaptive-testing/tests/action.test.js
+++ b/libraries/botbuilder-dialogs-adaptive-testing/tests/action.test.js
@@ -390,6 +390,10 @@ describe('ActionTests', function () {
         await TestUtils.runTestScript(resourceExplorer, 'Action_SetProperties');
     });
 
+    it('MissingProperty', async () => {
+        await TestUtils.runTestScript(resourceExplorer, 'Action_MissingProperty');
+    });
+
     it('SetProperty', async () => {
         await TestUtils.runTestScript(resourceExplorer, 'Action_SetProperty');
     });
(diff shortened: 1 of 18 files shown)


joshgummersall (approved):
Validated by @ryanlengel - thanks for the fix!

PR #3981, opened 2021-11-16
Suggestion for on-responded on prompts

### LINKED ISSUE #3947: [botbuilder-dialogs] Dialog does not prompt on continue dialog if any activity is sent on the turn
## Versions
What package version of the SDK are you using: 4.14.1
What nodejs version are you using: node14
What browser version are you using: NA
What os are you using: MacOS

## Describe the bug

On continueDialog a prompt checks first if the context.responded is false before showing the prompt. This behavior is incorrect in most cases. I show here an example with a typing activity (Which is very common)

```ts 
/* eslint-disable no-console */
import { ActivityTypes, ConversationState, MemoryStorage, TestAdapter } from 'botbuilder';
import { ChoicePrompt, DialogSet, DialogTurnStatus, ListStyle, WaterfallDialog } from 'botbuilder-dialogs';

const STATEMENT = 'Hello, I will ask you a question.';
const QUESTION = 'Do you like me?';
const prompt = new ChoicePrompt('choices', async (p) => {
  if (!p.recognized && p.attemptCount < 2) {
    return false;
  }

  return true;
});
const dialog = new WaterfallDialog('testDialog', [
  async (step) => {
    await step.context.sendActivity(STATEMENT);
    return step.prompt('choices', { prompt: QUESTION, choices: ['Yes', 'Of course'], style: ListStyle.list });
  },
]);

const memory = new MemoryStorage();

const dialogMemory = new ConversationState(memory);

const dialogSet = new DialogSet(dialogMemory.createProperty('dialogs'));

dialogSet.add(prompt);
dialogSet.add(dialog);

const testAdapter = new TestAdapter(async (ctx) => {
  const dc = await dialogSet.createContext(ctx);
  // Comment out the typing activity to see how it changes the behavior of the script.
  await ctx.sendActivity({ type: ActivityTypes.Typing });
  const result = await dc.continueDialog();
  if (result.status === DialogTurnStatus.empty) {
    await dc.beginDialog('testDialog');
  }
  await dialogMemory.saveChanges(ctx);
});

(async () => {
  await testAdapter.send('hi');
  testAdapter.activeQueue.splice(0, Infinity);
  await testAdapter.send('sorry, what?');
  const texts = testAdapter.activeQueue.map((a) => a.text);
  if (!texts.some((t) => t?.includes(QUESTION))) {
    throw new Error('Question not included in replies');
  }
})()
  .then(() => console.log('Well done'))
  .catch((err) => console.error(err));

// (async () => {
//   await testAdapter
//     .send('hi')
//     .assertReply({ type: ActivityTypes.Typing })
//     .assertReply(STATEMENT)
//     .assertReply((r) => {
//       r.text?.includes(QUESTION);
//     });
//   await testAdapter
//     .send('sorry, what')
//     .assertReply({ type: ActivityTypes.Typing })
//     .assertReply((r) => {
//       r.text?.includes(QUESTION);
//     });
// })()
//   .then(() => console.log('Well done'))
//   .catch((err) => console.error(err));
```
## Expected behavior
On continueDialog, the question should be shown, regardless of whether the application sends message before the prompt continueDialog is called

If backwards compatibility for this is necessary, I would recommend to add an option parameter that specifies whether to reply on continue or not. And if this parameter is empty, continue doing the current behavior

## PS: TestAdapter is broken

Also, note that if you run the test with the assertReply, the test adapter throws a unhandledpromiserejection. I believe currently the test adapter throws unhandled exceptions whenever

1. there is a failed assertion, 
2. If the test adapter handler itself throws an error.

That makes it really hard to use the TestAdapter on tests to test error throwing

alexrecuenco:
Fixes #3947

## Description
On issue #3947, relying on the responded flag todecide whether to show the prompt interferes with logic that could be happening either

a. Concurrently
b. On some middleware
c. As part of how we reached the prompt

I believe the *intention* is to *not* send a prompt if the validator has already sent a message. This is in fact not explained anywhere and it leads to confusion, and restricts the ability to write concise explanations to users along side a retry.

## Specific Changes

  - The default behavior is kept, *except* if a message was sent *before* the validator (Those are not considered as part of the prompt logic). This still runs into concurrency issues. For that,
  - A static method in Prompt has been added, setRepromptStatus. This allows the user to choose explicitly if the prompt should re-prompt or not.

## Testing

- Added extra tests for the new functionality *only* for the text prompt, since all other prompts reuse that same logic.

### DIFF
--- a/libraries/botbuilder-dialogs/src/prompts/prompt.ts
+++ b/libraries/botbuilder-dialogs/src/prompts/prompt.ts
@@ -172,6 +172,30 @@ export abstract class Prompt<T> extends Dialog {
      */
     protected constructor(dialogId: string, private validator?: PromptValidator<T>) {
         super(dialogId);
+
+        if (!validator) validator = async (prompt) => prompt.recognized.succeeded;
+
+        this.validator = async (prompt) => {
+            const context = prompt.context;
+            
+            const hasRespondedBeforeValidator = context.responded;
+            // Making the validator *only* consider as "responded" if responded has been set during a turn
+            // @ts-expect-error 
+            context._respondedRef.responded = false;
+            try {
+                const result = await validator(prompt);
+                const shouldReprompt = context.turnState.get(Prompt.repromptOnRetry);
+
+                // Set reprompt retry status based on whether the bot responded or not while on a validator
+                Prompt.setRepromptRetry(context, this.shouldRepromptOnRetry(context));
+                return result;
+            } finally {
+                if (hasRespondedBeforeValidator) prompt.context.responded = true;
+            }
+        }
+        
+
+
     }
 
     /**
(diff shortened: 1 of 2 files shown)


alexrecuenco:
Notice how I don't expect this to be merged *as-is*, but I hope to instead bring a conversation on how we can address #3947 while keeping as many of the design decisions made as possible.

PR #3983, opened 2021-11-17
fix: add accept header

### LINKED ISSUE #3962: port: Add */* accept header to ConnectorClient httpclient (#5926)
The changes in [Add */* accept header to ConnectorClient httpclient (#5926)](https://github.com/microsoft/botbuilder-dotnet/pull/5926) may need to be ported to maintain parity with `microsoft/botbuilder-dotnet`.

<blockquote>
partly fixes #5659
</blockquote>

EricDahlvang:
Fixes #3962


joshgummersall:
I think we probably need this code here as well: https://github.com/microsoft/botbuilder-js/blob/main/libraries/botframework-connector/src/auth/connectorFactoryImpl.ts

joshgummersall (approved):

PR #3990, opened 2021-11-30
fix: [botbuilder-core] TestAdapter throws unhandled promise rejection on error

alexrecuenco:
## Description
TestAdapter throws unhalded promise rejections on errors

- The testadapter should handle errors
- For that, TestFlow must implement the thenable implementation correctly

## Specific Changes

  - Implemented further the tenable on TestFlow to allow handling catch clauses correctly

## Testing

A test is added to verify the behavior change (If you run that test on main before this PR is merged, it fails)

#minor

### DIFF
--- a/libraries/botbuilder-core/tests/testAdapter.test.js
+++ b/libraries/botbuilder-core/tests/testAdapter.test.js
@@ -25,6 +25,41 @@ describe(`TestAdapter`, function () {
         await adapter.receiveActivity('test');
     });
 
+    it(`should reject on unhandled error.`, async function () {
+        const err = new Error('Intentional');
+        const adapter = new TestAdapter(() => {
+            throw err;
+        });
+        await assert.rejects(async () => {
+            await adapter.send('Message');
+        }, err);
+    });
+
+    it(`should call finally when no error happens`, async function () {
+        const adapter = new TestAdapter(async () => undefined);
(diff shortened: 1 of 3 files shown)


alexrecuenco:
Working around this bug on the testing framework is a bit painful currently, since we can't test that the application recovers correctly after unexpected errors.

libraries/botbuilder-core/tests/testAdapter.test.js
```diff
@@ -25,6 +25,41 @@ describe(`TestAdapter`, function () {
         await adapter.receiveActivity('test');
     });
 
+    it(`should reject on unhandled error.`, async function () {
+        const adapter = new TestAdapter((context) => {
```
joshgummersall, line 29: Slightly more strict assertion here, to ensure the proper error is propagated:

```typescript
const err = new Error('intentional');

const adapter = new TestAdapter(() => {
  throw err;
});

await assert.rejects(adapter.send(), err);
```

mrivera-ms (approved):
:shipit:

alexrecuenco:
Rebased again and pushed the api change as obtained running `yarn test:compat --local`

alexrecuenco:
Hmm. It seems that the api-extractor is changing the type of end-of-line. Any idea how to prevent that?

PR #4045, opened 2022-01-07
fix: CG security alert for axios 0.19.0, 0.21.1

BruceHaley:
Fixes #minor

## Description
axios is vulnerable to Inefficient Regular Expression Complexity
Severity high.

https://fuselabs.visualstudio.com/SDK_v4/_componentGovernance/112352/alert/5622360?typeId=10220548
https://fuselabs.visualstudio.com/SDK_v4/_componentGovernance/112352/alert/4224550?typeId=10220548
https://fuselabs.visualstudio.com/SDK_v4/_componentGovernance/112352/alert/5846780?typeId=10220548

## Specific Changes
Update @azure/ms-rest-js devDependency to the latest, 2.6.0. This eliminates the axios dependency altogether.
Remove unused dependencies.


joshgummersall:
Changing axios versions is a breaking change as we have code that exposes axios options to consumers of the SDK, so keep that in mind (cc @mrivera-ms).

libraries/adaptive-expressions/package.json
```diff
@@ -21,27 +21,9 @@
   "typings": "./lib/index.d.ts",
   "dependencies": {
     "@microsoft/recognizers-text-data-types-timex-expression": "1.3.0",
-    "@types/atob-lite": "^2.0.0",
```
joshgummersall, line 24: There's no way this is allowed, the build will fail. We use these dependencies all over the `adaptive-expressions` codebase.

libraries/botbuilder-ai-luis/package.json
```diff
@@ -30,7 +30,6 @@
     }
   },
   "dependencies": {
-    "botbuilder-ai": "4.1.6"
```
joshgummersall, line 33: You can't do this - `botbuilder-ai-luis` depends on this package: https://github.com/microsoft/botbuilder-js/blob/main/libraries/botbuilder-ai-luis/src/index.ts#L4

BruceHaley:
Rechecking for the axios vulnerability shows it to no longer be present:
C:\src\botbuilder-js>npm ls axios
botbuilder-js@4.13.0 C:\src\botbuilder-js
`-- @azure/ms-rest-js@1.9.1
  `-- axios@0.21.4

PR #4048, opened 2022-01-10
chore: Update botbuilder-js-daily.yml

munozemilio:
#minor

### DIFF
--- a/build/yaml/botbuilder-js-daily.yml
+++ b/build/yaml/botbuilder-js-daily.yml
@@ -106,18 +106,10 @@ steps:
 
 - template: templates/list-files.yml
 
-- task: DeleteFiles@1
-  displayName: 'Delete files from '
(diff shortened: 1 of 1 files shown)


build/yaml/botbuilder-js-daily.yml
```diff
@@ -106,18 +106,10 @@ steps:
...
-     ./node_modules/*
-  enabled: false
-
 - task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
   displayName: 'Component Detection'
   inputs:
-    failOnAlert: false
+    sourceScanPath: './libraries'
+    ignoreDirectories: './testing,./tools'
+    failOnAlert: true
```
mrivera-ms, line 114: Let's set this to false as there may be some issues that will remain.

PR #4069, opened 2022-01-20
fix: [#2745] Fix current eslint warnings - botbuilder-dialogs-adaptive library - src & conditions

sw-joelmut:
Adresses #2745
#minor

## Description
When reviewing the botbuilder-js project, it was found that eslint is showing several warnings for the botbuilder-dialogs-adaptive library.

## Specific Changes

- Replaced strings quotes with the correct one depending on the case
- Replaced hasOwnProperty with Object.prototype.hasOwnProperty.call
- Adjusted formatting
- Limited importing scopes
- Added missing jsdocs documentation for parameters and returns for functions
- Removed unused variables

## Testing
Eslint with no warnings and tests passed
image


libraries/botbuilder-dialogs-adaptive/package.json
```diff
@@ -28,6 +28,12 @@
     }
   },
   "dependencies": {
+    "@microsoft/recognizers-text-sequence": "~1.1.4",
+    "@microsoft/recognizers-text": "~1.1.4",
+    "@microsoft/recognizers-text-number-with-unit": "~1.1.4",
+    "@microsoft/recognizers-text-number": "~1.1.4",
+    "@microsoft/recognizers-text-choice": "~1.1.4",
+    "@microsoft/recognizers-text-date-time": "~1.1.4",
```
mrivera-ms, line 36: Why are we moving these?

mrivera-ms (changes requested):
@sw-joelmut, I don't see a lint rule recommending the change to dependencies. Please see my question above.

libraries/botbuilder-dialogs-adaptive/package.json
sw-joelmut: @mrivera-ms, moving these libraries from `devDependencies` to `dependencies` was required due to the rule [import/no-extraneous-dependencies](https://github.com/import-js/eslint-plugin-import/blob/v2.22.1/docs/rules/no-extraneous-dependencies.md).
The following image shows an example.
imagen

mrivera-ms (approved):

Issue #4094, opened 2022-02-03
Upgrade ms-rest-js to latest version (2.6.0)


ceciliaavila:
## Versions
What package version of the SDK are you using - 4.15.0
What nodejs version are you using 16.13.1

## Description
As part of supporting Node.js 16 (#3756) we should update the **@azure/ms-rest-js** package in the SDK libraries as its current version (1.9.1) is using uuid 3.4.0 which is deprecated.

## Additional Context
These are the warnings shown when installing the packages on a bot using botbuilder 4.15.0
![image](https://user-images.githubusercontent.com/44245136/152367205-d618acc9-fe62-4273-bf86-d4a78e45bc33.png)

PR #4161, opened 2022-03-16
fix: [#4123][botframework-connector] upgrade and clean up dependencies

### LINKED ISSUE #4123: [botframework-connector] upgrade and clean up dependencies
1. `@azure/identity` uses a beta version,  @azure/ms-rest-js is also very outdated. This could also have security implications. Suggest to upgrade.
2. Is it necessary to pin `jsonwebtoken` to a specific patch version? I suggest to use `^8.0.1` of `~8.0.1` instead of `8.0.1`. This would make it more compatible. 
3. There are a few type dependencies, which should be defined as dev dependencies `@types/jsonwebtoken` and `@types/node`.

current package.json:
```
"dependencies": {
    "@azure/identity": "2.0.0-beta.6",
    "@azure/ms-rest-js": "1.9.1",
    "@types/jsonwebtoken": "7.2.8",
    "@types/node": "^10.17.27",
    "adal-node": "0.2.3",
    "axios": "^0.25.0",
    "base64url": "^3.0.0",
    "botbuilder-stdlib": "4.1.6",
    "botframework-schema": "4.1.6",
    "cross-fetch": "^3.0.5",
    "jsonwebtoken": "8.0.1",
    "rsa-pem-from-mod-exp": "^0.8.4",
    "zod": "~1.11.17"
  },
```

ceciliaavila:
Fixes #4123 #4094

## Description
This PR upgrades de dependencies in **_botframework-connector_** library to use the latest version of `@azure/ms-rest-js`, `@azure/identity`, and `jsonwebtoken`.

**_Note: The consumer-tests check was updated to test against the TypeScript versions from 3.5 to 4.3. 
Versions 3.3 and 3.4 were removed from the check, otherwise, azure/ms-rest-js version should be fixed in 2.0.0 to avoid failures._**

## Specific Changes
  - Updated `botframework-connector`'s dependencies.
     - @azure/ms-rest-js and @azure/identity to latest version.
     - jsonwebtoken to version ^8.5.1 to match the one resolved for other projects.
     - @types/jsonwebtoken to version 8.3.5 (higher versions introduce breaking changes in _jwtTokenExtractor_)
  - Upgraded `@azure/ms-rest-js` in botbuilder and botbuilder-ai
  - Updated `botFrameworkAdapter.test`. unit tests to access the headers as an array due to a change between ms-rest-js and nock.
  - Updated `yarn.lock` with the new versions.
  - Updated `run.ts` to test against TypeScript versions from 3.5 to 4.3.

## Testing
This image shows the unit tests passing after the upgrades.
image


libraries/botbuilder-ai/package.json
```diff
@@ -28,7 +28,7 @@
   },
   "dependencies": {
     "@azure/cognitiveservices-luis-runtime": "2.0.0",
-    "@azure/ms-rest-js": "1.9.1",
+    "@azure/ms-rest-js": "^2.6.1",
```
joshgummersall, line 31: IIRC this is might be a breaking change due to the underlying library using `fetch` instead of `axios`. Would not upgrade this without deciding it's okay to take a breaking change./

tracyboehrer: @ceciliaavila Can we confirm if this is a breaking change and if there are alternatives?

ceciliaavila: Hi @tracyboehrer,

We reviewed the code where this library is used and the [changes](https://github.com/Azure/ms-rest-js/blob/master/Changelog.md#200---2019-06-21) that its new versions introduced and consider that they are not breaking changes for us (we are not using the interfaces and methods that were updated).

We also tested the libraries with the _azure/ms-rest-js_ upgraded version:

- We published the packages to a testing feed and ran the [BotFramework-FunctionalTests](https://github.com/microsoft/BotFramework-FunctionalTests/tree/main/Tests) pipelines. All tests passed (cards, proactive, authentication (single-tenant, multi-tenant, MSI), among others).
- We tested a sample using _botframework-streaming_ and _webchat_ and it worked.
- We tested the _CoreBot_ sample connected to a LUIS app and it worked as expected.

However, as an alternative, we could leave this dependency as it was and update only _azure/identity_, and _jsonwebtoken_ versions.

Please, let us know if you have any questions.

tracyboehrer (approved):

PR #4166, opened 2022-03-18
fix: [#4154] teamsNotifyUser does not accept messages returned from MessageFactory

### LINKED ISSUE #4154: teamsNotifyUser does not accept messages returned from MessageFactory.
## Versions
What package version of the SDK are you using: 1.15.0
What nodejs version are you using: 16.11.0
What browser version are you using: n/a, this is a typescript compiler error
What os are you using: Win 11, 22H2, 22567.200

## Describe the bug
When using teamsNotifyUser in a typescript application [the example provided by the docs](https://docs.microsoft.com/en-us/microsoftteams/platform/bots/how-to/conversations/conversation-messages?tabs=typescript#notifications-to-your-message) on how to notify someone does not work.
Instead, the following error is returned:
```
Argument of type 'Partial<Activity>' is not assignable to parameter of type 'Activity'.
  Property 'type' is optional in type 'Partial<Activity>' but required in type 'Activity'.ts(2345)
```

This is because `MessageFactory.text('Message');` returns a `Partial<Activity>` while `teamsNotifyUser()` requires a straight `Activity`.

## To Reproduce

1. In Teams Toolkit, create a new app. Selecting only Bot as a capability and Typescript as the language.
2. In `bot/teamsBot.ts` replace
```ts
case "welcome": {
    const card = AdaptiveCards.declareWithoutData(rawWelcomeCard).render();
    await context.sendActivity({ attachments: [CardFactory.adaptiveCard(card)] });
    break;
}
```
with
```ts
case "welcome": {
    let message = MessageFactory.text("You'll get a notification, if you've turned them on.");
    teamsNotifyUser(message);
    await context.sendActivity(message);
}
```
3. From `./bot` run `npm run build`
4. See the following build error:
```
teamsBot.ts:39:27 - error TS2345: Argument of type 'Partial<Activity>' is not assignable to parameter of type 'Activity'.
  Property 'type' is optional in type 'Partial<Activity>' but required in type 'Activity'.

39           teamsNotifyUser(message);
``` 

## Expected behavior
To be able to call `teamsNotifyUser()` with the value returned from the MessageFactory.

## Screenshots
image
image

## Additional context
I debated creating a bug against the SDK or the docs but felt it made more sense to create it here.

ceciliaavila:
Fixes #4154

## Description
This PR updates the `teamsNotifyUser` method in teamsActivityHelper class to receive a Partial<Activity> instead of an Activity enabling to call it using a MessageFactory.
This method only uses the Activity's ChannelData which is not a required property.

## Specific Changes
- Updated `teamsNotifyUser` method to receive a Partial<Activity> as parameter.
- Updated the method signature in `botbuilder.api.md` documentation.

## Testing
These images show a Teams bot sending a notification to the user.
image

### DIFF
--- a/libraries/botbuilder/etc/botbuilder.api.md
+++ b/libraries/botbuilder/etc/botbuilder.api.md
@@ -444,7 +444,7 @@ export class TeamsInfo {
 }
 
 // @public
-export function teamsNotifyUser(activity: Activity, alertInMeeting?: boolean, externalResourceUrl?: string): void;
+export function teamsNotifyUser(activity: Partial<Activity>, alertInMeeting?: boolean, externalResourceUrl?: string): void;
 
 // @public
 export class TeamsSSOTokenExchangeMiddleware implements Middleware {
--- a/libraries/botbuilder/src/teamsActivityHelpers.ts
+++ b/libraries/botbuilder/src/teamsActivityHelpers.ts
@@ -12,7 +12,7 @@ function isTeamsChannelData(channelData: unknown): channelData is TeamsChannelDa
     return typeof channelData === 'object';
 }
 
-function validateActivity(activity: Activity): void {
+function validateActivity(activity: Partial<Activity>): void {
     if (!activity) {
         throw new Error('Missing activity parameter');
     }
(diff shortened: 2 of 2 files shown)


gabog:
Hi @EricDahlvang, can you please take a look at this PR? 

Thanks

EricDahlvang (approved):

PR #4167, opened 2022-03-18
fix: [#3192] Missing LICENSE files and invalid package.json - Part 1

sw-joelmut:
Addresses #3192
#minor

## Description
This PR modifies the `update-versions` script, to take the private packages into account when updating all the libraries versions and references and also the `peerDependencies`.

## Specific Changes
- Updated `update-versions` script to gather all the repo packages and update the version for the ones that aren't private.
- Updated `update-versions` script to take the `peerDependencies` into consideration.

### DIFF
--- a/libraries/botbuilder-repo-utils/src/package.ts
+++ b/libraries/botbuilder-repo-utils/src/package.ts
@@ -15,6 +15,7 @@ export interface Package {
 
     dependencies?: Record<string, string>;
     devDependencies?: Record<string, string>;
+    peerDependencies?: Record<string, string>;
 
     scripts?: Record<string, string>;
 }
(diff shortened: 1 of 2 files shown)


gabog (changes requested):
Hi @sw-joelmut, can you revert the updates from 4.1.6 to 4.16? We need to determine what version we want to use for dev packages and we can address that in another PR.

sw-joelmut:
> Hi @sw-joelmut, can you revert the updates from 4.1.6 to 4.16? We need to determine what version we want to use for dev packages and we can address that in another PR.

@gabog, `4.1.6` to `4.16.0` changes reverted, only remains the adjustments for the `update-versions` script. 

Thanks

PR #4180, opened 2022-04-07
fix: Azure Blob transcript key 

### LINKED ISSUE #4176: BlobsTranscriptStore blobkey not persisting '/' for virtual directory
## Versions
What package version of the SDK are you using. `4.15.0`
What nodejs version are you using `14`

## Describe the bug
Trying to use the blobsTranscriptStore to record transcript.  From the code, expecting the blob created for activity in Azure Storage to follow the pattern of `container/{channelId]/{conversationId}/{Timestamp.ticks}-{activity.id}.json`

https://github.com/microsoft/botbuilder-js/blob/64926f42264863ee5d3519522013e238aa295b3c/libraries/botbuilder-azure-blobs/src/blobsTranscriptStore.ts#L61-L69

But instead the blog key ended up with the '/' also URIencoded so the blob name in Azure blob ended up to be `container%2F{channelId}%2F{conversationId}%2F{Timestamp.ticks}-{activid.id}.json`.  Which makes it not compatible with the virtual directory feature of the Azure Blob with the actual character '/' in the key. 

It looks like it is because the way `sanitizeBlobKey` is implemented, which also sanitized the '/' character.

https://github.com/microsoft/botbuilder-js/blob/a7000885acca69f52a0c7f301a44835f3f8f100e/libraries/botbuilder-azure-blobs/src/sanitizeBlobKey.ts#L4-L18

## To Reproduce
Steps to reproduce the behavior in a adaptive dialog bot with bot composer.
1. Configure 'appsettings.json'
   1. set 'runtimeSettings::features::traceTranscript' to 'true'
   2. set 'runtimeSettings::features::blobTranscript' with 'connectionString' and 'containerName'
2. Interact with the bot
3. Check the transcript created in Azure blob container

## Expected behavior
The blob created to have a key with actual '/' in the name rather than sanitized with %2F

## Screenshots
Blob key with %2F generated.  Expect to be real '/' such that it becomes "virtual directory" in storage explorer.

image

ramfattah:
Fixes #4176 

## Description

Expecting Azure blob to creat transcript activity and follow the pattern of `container/{channelId]/{conversationId}/{Timestamp.ticks}-{activity.id}.json`

instead it was creating directory `container%2F{channelId}%2F{conversationId}%2F{Timestamp.ticks}-{activid.id}.json` due to [sanitizeBlobKey ](https://github.com/microsoft/botbuilder-js/blob/c5de4422f37793e6080949974368e3a41846a7a6/libraries/botbuilder-azure-blobs/src/sanitizeBlobKey.ts#L11)method using the [encodeURIComponent ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)function.

## Specific Changes

  - Created optional `decodeTranscriptKey` boolean property in [`BlobsTranscriptStoreOptions`](https://github.com/microsoft/botbuilder-js/blob/ramfattah/blobTranscript/libraries/botbuilder-azure-blobs/src/blobsTranscriptStore.ts#:~:text=export%20interface%20BlobsTranscriptStoreOptions%20%7B) interface to return a new string representing the decoded version of the given encoded blob transcript key. This remains the default behavior to false but can be overridden by setting decodeTranscriptKey to true.

## Steps to reproduce

> Prerequisite: Azure Blob

1) Run `git clone --branch ramfattah/blobTranscript https://github.com/microsoft/botbuilder-js.git`

2) Run `yarn` and `yarn build` in the root folder of SDK

3) Open `libraries/botbuilder-azure-blobs/tests/blobsTranscriptStore.test.js` file
     * Add your blob `connectionString` and `containerName`. Note, it will create the container automatically if the name doesn't already exist. 
     * image
     * Flip `decodeTranscriptKey` flag to true
     * image

4) In terminal, run `mocha <path-to-blobsTranscriptStore.test.js`
     * Ex: `mocha /Users/ram/Documents/bot/botbuilder-js/libraries/botbuilder-azure-blobs/tests/blobsTranscriptStore.test.js`

5) This will create blob transcript key in the container as the following pattern  `container/{channelId]/{conversationId}/{Timestamp.ticks}-{activity.id}.json`

## Before
image

## After

[Image from Gyazo](https://gyazo.com/ebd603d7ed9ab57f8471de489901332e)

### DIFF
--- a/libraries/botbuilder-azure-blobs/etc/botbuilder-azure-blobs.api.md
+++ b/libraries/botbuilder-azure-blobs/etc/botbuilder-azure-blobs.api.md
@@ -31,11 +31,12 @@ export class BlobsTranscriptStore implements TranscriptStore {
     deleteTranscript(channelId: string, conversationId: string): Promise<void>;
     getTranscriptActivities(channelId: string, conversationId: string, continuationToken?: string, startDate?: Date): Promise<PagedResult<Activity>>;
     listTranscripts(channelId: string, continuationToken?: string): Promise<PagedResult<TranscriptInfo>>;
-    logActivity(activity: Activity): Promise<void>;
+    logActivity(activity: Activity, options?: BlobsTranscriptStoreOptions): Promise<void>;
     }
 
 // @public
 export interface BlobsTranscriptStoreOptions {
+    decodeTranscriptKey?: boolean;
     storagePipelineOptions?: StoragePipelineOptions;
 }
 
--- a/libraries/botbuilder-azure-blobs/src/blobsTranscriptStore.ts
+++ b/libraries/botbuilder-azure-blobs/src/blobsTranscriptStore.ts
@@ -33,14 +33,15 @@ function getConversationPrefix(channelId: string, conversationId: string): strin
 }
 
 // Formats an activity as a blob key
-function getBlobKey(activity: Activity): string {
+function getBlobKey(activity: Activity, options?: BlobsTranscriptStoreOptions): string {
     const { timestamp } = z
         .object({ timestamp: z.instanceof(Date) })
         .nonstrict()
         .parse(activity);
 
     return sanitizeBlobKey(
-        [activity.channelId, activity.conversation.id, `${formatTicks(timestamp)}-${activity.id}.json`].join('/')
+        [activity.channelId, activity.conversation.id, `${formatTicks(timestamp)}-${activity.id}.json`].join('/'),
+        options
     );
 }
 
@@ -56,6 +57,12 @@ export interface BlobsTranscriptStoreOptions {
      * storage client
      */
     storagePipelineOptions?: StoragePipelineOptions;
+
+    /**
+     * Optional setting to return a new string representing the decoded version of the given encoded blob transcript key.
+     * This remains the default behavior to false, but can be overridden by setting decodeTranscriptKey to true.
+     */
+    decodeTranscriptKey?: boolean;
 }
 
 /**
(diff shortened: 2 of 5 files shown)


tracyboehrer (approved):

tariromukute:
Hi @tracyboehrer and @ramfattah . I am trying to use the this from the `TranscriptLoggerMiddleware`. The middleware does not call `logActivity` with the options object and does not support setting an options object  see [code here](https://github.com/microsoft/botbuilder-js/blob/5e5d8198f609f577828bf798abfd218d8129aa6d/libraries/botbuilder-core/src/transcriptLogger.ts#L131)

I think the logActivity in the  [blobsTranscriptStore](https://github.com/microsoft/botbuilder-js/blob/5e5d8198f609f577828bf798abfd218d8129aa6d/libraries/botbuilder-azure-blobs/src/blobsTranscriptStore.ts#L283) should also use  `this._isDecodeTranscriptKey` to set options for `getBlobKey`. This way, the middleware won't need to maintain the same attribute.

What are your thoughts, can we create a new issue for this?

tracyboehrer:
@tariromukute The main issue is that what options are being used?  Blobs options?  The transcript middleware doesn't know anything at all about which Storage is being used (Memory, CosmosDB, Blobs, etc...).  It only knows about TranscriptLogger.  So while the BlobsTranscriptStore has an optional options argument, it's not known to the middleware.

One option might be to implement a transcript logger that does have some knowledge about the store being used.  Though perhaps a better option would be to create a BlobsTranscriptStore using a particular BlobsStorageOptions.

PR #4241, opened 2022-05-12
fix: [#4204] Fix remaining eslint warnings - enable ESLint config

### LINKED ISSUE #4204: Fix remaining eslint warnings
## Description
There are still eslint warnings to solve in several libraries in order to treat the violations as errors in CI.
This is the current status:
image

### Proposed solution
Fix all the remaining warnings in the different libraries and update the [.eslintrc.json](https://github.com/microsoft/botbuilder-js/blob/main/.eslintrc.json) file to prevent new issues to be introduced.

sw-joelmut:
Fixes #4204

> **Note:** This PR `lint` script will successfully pass once all warnings are addressed. This is being tackled in separate PRs of the same issue.

## Description
This PR updates the ESLint configuration adding new rules and disabling the ones regarding the `any` type, since it requires deeper analysis to not use the type. Moreover, we force the `lint` script to fail if there is at least one warning, preventing to introduce new ones in the future.

> **Note:** we excluded the @typescript-eslint/no-explicit-any and @typescript-eslint/explicit-module-boundary-types to tackle those issues as a separate task, as they require deeper analysis and testing.

## Specific Changes
**.eslintrc.json**
- Added `node: true` to the `env` property, due to not recognizing multiple NodeJs types.
- Disabled `@typescript-eslint/no-explicit-any`, `@typescript-eslint/explicit-module-boundary-types` and `@typescript-eslint/ban-types`, since there are many part of the source code that uses implicit or explicit `any` types, which requires deeper analysis to decide which type should use instead.
- Added `jsdoc/require-jsdoc` rule to warn the user when JSDocs are not provided in public classes, methods, properties and functions (ignoring private and protected).

## Testing
The following images show the configuration working when there are/aren't errors.
image


tracyboehrer (approved):

tracyboehrer:
@sw-joelmut Can you take a look at this?  Failing during yarn lint.

PR #4404, opened 2023-01-10
fix: refactor Route registration and Application.run() for better Invoke-handling

stevengum:
#minor

## Description
The current pattern in botbuilder-m365's `Application` is that all routes are registered to a single array and processed in order of the registration (via `Application.addRoute()`). This leads to an edge case where Teams' Invoke Activities may not be responded within the required time window.

#### Scenario:
A developer first registers 10 AppRoutes. Their RouteSelectors are for non-Invoke activities, but each require a minimum of 500ms to determine whether or not the RouteHandler should handle the activity.

The developer then registers 3 Message Extension AppRoutes and 3 AdaptiveCards' CardAction AppRoutes.

When an Invoke activity is received by the Application, it iterates over the first 10 AppRoutes, which take at least 5 seconds for all the RouteSelectors to determine that they won't process the activity. Teams' Invokes need to be responded to within 5 seconds, so the end-user in Teams that triggered the Invoke is going to see an error message in the Teams GUI, saying something like "the bot didn't respond" or "the bot is unreachable".

### Proposed Solution
Add another private member array to Application that holds all AppRoutes for selecting and handling Invoke activities (e.g., those registered by `MessageExtensions` and the applicable ones from `AdaptiveCards`). This new member (i.e., `Application._invokeRoutes`) is run through _first_ when an Invoke Activity is received. If none of these `Application._invokeRoutes` handle the Invoke, the remaining AppRoutes (via the existing `Application._routes`) are called in order of registration.

Additionally, we'll need to update the docstrings, documentation provide samples to succintly demonstrate that responding to Invokes quickly is important.

## Specific Changes
  - Add new overload in `Application`: `public addRoute(selector: RouteSelector, handler: RouteHandler<TState>, isInvokeRoute: boolean = false): this`
  - Update `MessageExtensions` and relevant `AdaptiveCards` registration methods to use this overload as appropriate

## Testing
No tests yet, will rebase and update/add tests after @stevenic's next planned commit with sample and tests

### DIFF
--- a/libraries/botbuilder-m365/src/AdaptiveCards.ts
+++ b/libraries/botbuilder-m365/src/AdaptiveCards.ts
@@ -67,7 +67,7 @@ export class AdaptiveCards<TState extends TurnState> {
                     // Queue up invoke response
                     await context.sendActivity({ value: { body: response, status: 200 } as InvokeResponse, type: ActivityTypes.InvokeResponse })
                 }
-            });
+            }, true);
         });
         return this._app;
     }
@@ -124,7 +124,7 @@ export class AdaptiveCards<TState extends TurnState> {
                     // Queue up invoke response
                     await context.sendActivity({ value: { body: response, status: 200 } as InvokeResponse, type: ActivityTypes.InvokeResponse })
                 }
-            });
+            }, true);
         });
         return this._app;
     }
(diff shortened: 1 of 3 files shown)


Stevenic:
Thanks for calling out the edge case... There are actually numerous cases like this. It's by design that order matters. At the same time I don't want developers to have to understand nuances like Invoke timeouts and such. I'm going to take this PR for now but I may end up backing it out at some point.  The alternative would be to document that you should include any long running filters towards the end of your registration list.

Stevenic (approved):

PR #4413, opened 2023-01-20
fix: streaming client should connect under Node.js

### LINKED ISSUE #4412: Using Node.js with ASE did not work
## Versions

What package version of the SDK are you using. 4.14.1
What nodejs version are you using 18
What os are you using N/A

## Describe the bug

Use DirectLineJS ASE or NodeWebSocket.connect() to connect to ASE or any Web Socket server.

### Using DirectLineJS

```js
import { DirectLineStreaming } from 'botframework-directlinejs';

const TOKEN = '...';

async function main() {
  const token = await fetchToken();

  const connection = new DirectLineStreaming({
    domain: 'https://your-bot.azurewebsites.net/.bot/v3/directline',
    token: TOKEN
  });

  connection.connectionStatus$.subscribe(connectionStatus => {
    console.log(`Connection status: ${connectionStatus}`);
  });

  connection.activity$.subscribe(activity => {
    console.log(`Activity: ${activity.text}`);
  });
}

main();
```

### Using `NodeWebSocket.connect()`

```js
import { NodeWebSocket } from 'botframework-streaming';

async function main() {
  const socket = new NodeWebSocket();

  await socket.connect('wss://your-bot.azurewebsites.net/');
}

main();
```

## Expected behavior

It should connect and will not throw any exceptions.

### Using DirectLineJS

It should connect.

```
Failed to connect Error: Unable to connect client to Node transport.
/home/compulim/repos/experiment-dljs-ase-nodejs/node_modules/botframework-streaming/lib/webSocket/nodeWebSocketClient.js:63
                throw new Error(`Unable to connect client to Node transport.`);
                      ^

Error: Unable to connect client to Node transport.
    at WebSocketClient.<anonymous> (/home/compulim/repos/experiment-dljs-ase-nodejs/node_modules/botframework-streaming/lib/webSocket/nodeWebSocketClient.js:63:23)
    at Generator.throw (<anonymous>)
    at rejected (/home/compulim/repos/experiment-dljs-ase-nodejs/node_modules/botframework-streaming/lib/webSocket/nodeWebSocketClient.js:13:65)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

### Using `NodeWebSocket.connect`

It should not use URL as hostname. Because it is using the URL as the hostname, DNS cannot resolve `wss://your-bot.azurewebsites.net/`. Instead, it should resolve `your-bot.azurewebsites.net`.

```
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

Error: getaddrinfo ENOTFOUND wss://your-bot.azurewebsites.net/
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:109:26) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'wss://your-bot.azurewebsites.net/'
}
```

## Additional context

`NodeWebSocket.connect` is not implemented correctly. Currently, it is:

1. Create a Web Socket server (*not client*)
2. Fake handshake
3. Fake UPGRADE request
4. Wait until the Web Socket is closed

It is broken in multiple ways:

- It should not create a server to fake the connection
- It is insufficiently faking the handshake
- It should wait until the Web Socket is connected (`open` event), instead of close

I believe `NodeWebSocket.connect` never work and never tested properly.

compulim:
Fixes #4412.

## Description

`NodeWebSocket.connect()` is not implemented correctly and should never work. There are multiple problems in the original implementation:

- It expects the first argument is hostname
   - Instead, it should be server URL (for example, ws://your-bot.azurewebsites.net/.bot/v3/directline?token=123)
- It is using a server, fake handshake, to get the `WebSocket` object out, which is probably not working
- It is waiting for `close` event for connection to established, which is wrong

We are not sure why the original implementation has done this way.

We are updating the implementation to use `ws` package to connect.

### Removing second argument (`port = 8082`)

We are removing the second argument (`port = 8082`) because:

1. The `serverAddress` will contain the port number (by default, `ws:` is 80, and `wss:` is 443)
2. If we keep the current logic, we will *always* ignore the port number in `serverAddress` because `port` is default to 8082

This does not sound intuitive to developer because the port number in `serverAddress` will *always* get replaced.

If we don't default port number to 8082, we will need to change the API saying there is no default (i.e. `undefined`). At the end of the day, we are still modifying the API and will introduce breaking changes.

Regardless of whether we keep `port` or not, we are still introducing breaking changes. Thus, we are choosing to remove `port` to simplify our code.

If the developer need to change the port to `8082`, they could use the following code snippet:

```js
const url = new URL(serverAddress);

url.port = 8082;

await nodeWebSocket.connect(url.href);
```

## Specific Changes

❗This PR will introduce breaking changes and modified the API. Please bump version appropriately.

- Update `NodeWebSocket.connect()` to connect to a Web Socket server as a client
- Add tests
   - We did not add unhappy path (`connect()` rejections) because it will print errors to console and there is no easy way to suppress
- Cleaned up code related to `watershed`
   - `watershed` was one of the candidate packages for using Web Socket in Node.js
   - However, ultimately, we chose `ws`
   - There are no `binary`, `data`, and `text` event in the `ws` version of Web Socket
   - Thus, we are removing test code related to these events
- Separating mocks
   - Instead of using an uber `FauxSock` which mocks Web Socket and TCP/IP Socket, we are opting for a more specific mock for TCP/IP Socket, namely `FauxSocket`

## Testing

Added tests to ensure `NodeWebSocket` can connect to a real Web Socket server.


ckkashyap (approved):
I looked at NodeWebSocket.ts in detail and it looks fine.

tracyboehrer:
@ckkashyap @compulim What is the expected impact of this being a breaking change?  What would be a shame is to correct one issue and break another set of customers.

compulim:
I am modifying the API to become `connect(serverAddressOrHostName: string, port = 8082)`. So it is conceptually backward-compatible.

1. If `serverAddressOrHostName` is a valid URL (via `new URL`), then use it as-is, without the second argument (ignore `port`)
2. Otherwise, `serverAddressOrHostName` is a host name, then
   - Construct an URL object by using `serverAddressOrHostName` as `hostname` and second argument as `port`
   - The protocol used is `ws:` and not `wss:`
   - This is for backward compatibility as previous version suggests non-secure protocol
   
As we do not have any existing tests that proves the existing code is working, we are doing our best efforts to make it backward compatible.

compulim:
I tried to write a test to show how previous signature would work. However, there are obvious mistakes in the code that made it very unlikely to work:

- `options.headers.upgrade` must set to `'websocket'`
- Second argument of `WebSocket.server.completeUpgrade` should be `{}`, instead of `undefined`

I am keeping the old code as-is, without any tests.

libraries/botframework-streaming/src/webSocket/nodeWebSocket.ts
```diff
@@ -74,23 +76,50 @@ export class NodeWebSocket implements ISocket {
...
+
+        if (url?.hostname) {
+            return new Promise<void>((resolve, reject) => {
+                const ws = (this.wsSocket = new WebSocket(url));
+
+                ws.once('error', ({ message }) => reject(new Error(message)));
+                ws.once('open', () => resolve());
+            });
+        }
+
```
ckkashyap, line 99: A warning here may be useful?

ckkashyap (approved):

johnataylor (approved):

PR #4416, opened 2023-01-24
type structure changes, tested

### LINKED ISSUE #4415: targeted meeting notification interface changes
The existing types and interfaces associated with the `sendMeetingNotification` API call are incorrect

singhk97:
Fixes #4415 

## Description

* Refactored types and interfaces associated with `sendMeetingNotification` api call, so that it can be extended without any breaking changes. And by 'extend`, I mean adding new types, making-optional existing properties and adding properties to existing interfaces.

## Specific Changes

* Made `channelData` an optional property
* Updated the type of `onBehalfOf` property to be a list of `TeamsOnBehalfOf` objects.
* Notification payload type expected the `notification` parameter in `sendMeetingNotification` API function is  `TeamsMeetingNotification`. Currently it is equal to `TargetedMeetingNotification`. However it can made be a union of any arbitrary type. This makes the payload type extensible without any breaking changes.
* The request body mappers were simplified so that any value passed to the `value` property would be serialized without any hard coded constraints. 
* `surfaces` now is a list of `TeamsMeetingSurface` type. Just like `TeamsMeetingNotification` it can be a union of any arbitrary type. Making it possible to have more than one type of surface object in the list of surfaces.

## Testing

* Performed end-to-end test
* Unit tests remain the same

### DIFF
--- a/libraries/botframework-connector/src/teams/models/mappers.ts
+++ b/libraries/botframework-connector/src/teams/models/mappers.ts
@@ -434,40 +434,7 @@ export const TeamsMeetingDetails: msRest.CompositeMapper = {
     },
 };
 
-export const TeamsMeetingNotificationInfo: msRest.CompositeMapper = {
-    serializedName: 'TeamsMeetingNotificationInfo',
-    type: {
-        name: 'Composite',
-        className: 'TeamsMeetingNotificationInfo',
-        modelProperties: {
-            recipients: {
-                serializedName: 'recipients',
-                type: {
-                    name: 'Sequence',
-                    element: {
-                        type: {
-                            name: 'String',
-                        },
-                    },
-                },
-            },
-            surfaces: {
-                serializedName: 'surfaces',
-                type: {
-                    name: 'Sequence',
-                    element: {
-                        type: {
-                            name: 'Composite',
-                            className: 'TeamsMeetingNotificationSurface',
-                        },
-                    },
-                },
-            },
-        },
-    },
-};
-
-export const TeamsMeetingNotification: msRest.CompositeMapper = {
+export const TeamsMeetingNotification: msRest.CompositeMapper =  {
     serializedName: 'TeamsMeetingNotification',
     type: {
         name: 'Composite',
@@ -482,8 +449,7 @@ export const TeamsMeetingNotification: msRest.CompositeMapper = {
             value: {
                 serializedName: 'value',
                 type: {
-                    name: 'Composite',
-                    className: 'TeamsMeetingNotificationInfo',
+                    name: 'any',
                 },
             },
             channelData: {
@@ -497,34 +463,6 @@ export const TeamsMeetingNotification: msRest.CompositeMapper = {
     },
 };
 
-export const TeamsMeetingNotificationSurface: msRest.CompositeMapper = {
-    serializedName: 'TeamsMeetingNotificationSurface',
-    type: {
-        name: 'Composite',
-        className: 'TeamsMeetingNotificationSurface',
-        modelProperties: {
-            surface: {
-                serializedName: 'surface',
-                type: {
-                    name: 'String',
-                },
-            },
-            contentType: {
-                serializedName: 'contentType',
-                type: {
-                    name: 'String',
-                },
-            },
-            content: {
-                serializedName: 'content',
-                type: {
-                    name: 'Composite',
-                    className: 'TaskModuleContinueResponse',
-                },
-            },
-        },
-    },
-};
 
 export const TeamsMeetingNotificationChannelData = {
     serializedName: 'TeamsMeetingNotificationChannelData',
--- a/libraries/botframework-schema/src/teams/index.ts
+++ b/libraries/botframework-schema/src/teams/index.ts
@@ -1820,28 +1820,50 @@ export interface MeetingEndEventDetails extends MeetingEventDetails {
 
 /**
  * @interface
- * Specifies details of a Teams meeting send notification payload.
+ * Specifies base details of a Teams meeting send notification payload.
  */
-export interface TeamsMeetingNotification {
+export interface TeamsMeetingNotificationBase<T> {
     /**
      * @member {string} [type] The type of meeting notification.
      */
-    type: 'targetedMeetingNotification' | string;
+    type: string;
+
     /**
-     * @member {TeamsMeetingNotificationInfo} [type] The specific details of the meeting notification.
+     * @member {T} [value] The specific details of the meeting notification.
      */
-    value: TeamsMeetingNotificationInfo;
+    value: T;
+} 
+
+
+/**
+ * @interface
+ * Specifies details of a targeted Teams meeting send notification payload.
+ */
+export interface TargetedMeetingNotification extends TeamsMeetingNotificationBase<TargetedMeetingNotificationInfo> {
     /**
-     * @member {TeamsMeetingNotificationChannelData} [type] The channel data of the meeting notification.
+     * @member {string} [type] The type of meeting notification.
      */
-    channelData: TeamsMeetingNotificationChannelData;
-}
+    type: 'targetedMeetingNotification';
+
+    /**
+     * @member {TeamsMeetingNotificationChannelData} [channelData] The channel data of the meeting notification.
+     */
+    channelData?: TeamsMeetingNotificationChannelData;
+} 
+
+
+/**
+ * @type
+ * Defines the base Teams meeting notification type
+ */
+export type TeamsMeetingNotification = TargetedMeetingNotification;
+
 
 /**
  * @interface
  * Specifies recipients and surfaces to target in a Teams meeting notification.
  */
-export interface TeamsMeetingNotificationInfo {
+export interface TargetedMeetingNotificationInfo {
     /**
      * @member {string[]} [recipients] The list of recipient meeting MRIs.
      */
@@ -1849,26 +1871,26 @@ export interface TeamsMeetingNotificationInfo {
     /**
      * @member {TeamsMeetingNotificationSurface[]} [surfaces] The List of notification surface configuration.
      */
-    surfaces: TeamsMeetingNotificationSurface[];
+    surfaces: TeamsMeetingSurface[];
 }
 
-/**
- * @interface
- * Specifies the surface and content for a Teams meeting notification.
- */
-export interface TeamsMeetingNotificationSurface {
+
+export type TeamsMeetingSurface = TeamsMeetingStageSurface<any>;
+
+
+export interface TeamsMeetingStageSurface<T> {
     /**
      * @member {string} [surface] The surface type.
-     */
-    surface: 'meetingStage' | string;
+    */
+    surface: 'meetingStage';
     /**
      * @member {string} [contentType] The content type.
-     */
+    */
     contentType: 'task' | string;
     /**
-     * @member {TaskModuleContinueResponse} [content] The content to display in the meeting notification.
+     * @member {T} [content] The content to display in the meeting notification.
      */
-    content: TaskModuleContinueResponse;
+    content: T;
 }
 
 /**
@@ -1879,7 +1901,7 @@ export interface TeamsMeetingNotificationChannelData {
     /**
      * @member {TeamsMeetingOnBehalfOf} [onBehalfOf] The user that the bot is sending the notification on behalfOf, if any.
      */
-    onBehalfOf?: TeamsMeetingOnBehalfOf;
+    onBehalfOf?: TeamsMeetingOnBehalfOf[];
 }
 
 /**
--- a/libraries/botframework-connector/src/teams/models/teamsMappers.ts
+++ b/libraries/botframework-connector/src/teams/models/teamsMappers.ts
@@ -16,8 +16,6 @@ export {
     TeamsMeetingInfo,
     TeamsMeetingParticipant,
     TeamsMeetingNotification,
-    TeamsMeetingNotificationInfo,
-    TeamsMeetingNotificationSurface,
     TeamsMeetingNotificationChannelData,
     TeamsMeetingOnBehalfOf,
     TeamsMeetingNotificationRecipientFailureInfo,


libraries/botframework-schema/src/teams/index.ts
```diff
@@ -1820,73 +1820,96 @@ export interface MeetingEndEventDetails extends MeetingEventDetails {
...
      */
-    value: TeamsMeetingNotificationInfo;
+    type: 'targetedMeetingNotification';
+
     /**
-     * @member {TeamsMeetingNotificationChannelData} [type] The channel data of the meeting notification.
+     * @member {MeetingNotificationChannelData} [channelData] The channel data of the meeting notification.
      */
-    channelData: TeamsMeetingNotificationChannelData;
+    channelData?: MeetingNotificationChannelData;
```
YunnyChung, line 1851: I think channelData needs to be in the base class as any meeting notification payload will have three properties. 

(1) Type
(2) T Value
(3) ChannelData 

For targeted meeting notification payload, the value is going to be TargetedMeetingNotificationValue/Info :)

libraries/botframework-schema/src/teams/index.ts
```diff
@@ -1928,9 +1951,9 @@ export interface TeamsMeetingNotificationRecipientFailureInfo {
  * @interface
  * Specifies the list of recipients for which the Teams meeting notification was not sent.
  */
-export interface TeamsMeetingNotificationRecipientFailureInfos {
+export interface MeetingNotificationRecipientFailureInfos {
```
YunnyChung, line 1954: Is it better to introduce this additional layer over directly defining `MeetingNotificationRecipientFailureInfo[]` as MeetingNotificationResponse?

libraries/botbuilder/etc/botbuilder.api.md
```diff
@@ -458,7 +458,7 @@ export class TeamsInfo {
     static getTeamMember(context: TurnContext, teamId?: string, userId?: string): Promise<TeamsChannelAccount>;
     // @deprecated
     static getTeamMembers(context: TurnContext, teamId?: string): Promise<TeamsChannelAccount[]>;
-    static sendMeetingNotification(context: TurnContext, notification: TeamsMeetingNotification, meetingId?: string): Promise<TeamsMeetingNotificationRecipientFailureInfos | {}>;
+    static sendMeetingNotification(context: TurnContext, notification: MeetingNotification, meetingId?: string): Promise<MeetingNotificationRecipientFailureInfos | {}>;
```
YunnyChung, line 461: Similar to my other comment - can we simply do MeetingNotificationRecipientFailureInfo[]?

YunnyChung: And is there a reason why we use `MeetingNotificationRecipientFalureInfos | {}` over MeetingNotificationResponse? ex: What I am expecting is similar to `Promise<TeamsPagedMembersResult>`

libraries/botframework-connector/src/teams/models/mappers.ts
```diff
@@ -548,11 +485,11 @@ export const TeamsMeetingNotificationChannelData = {
     },
 };
 
-export const TeamsMeetingOnBehalfOf: msRest.CompositeMapper = {
-    serializedName: 'TeamsMeetingOnBehalfOf',
+export const MeetingOnBehalfOf: msRest.CompositeMapper = {
```
YunnyChung, line 488: Is there an existing OnBehalfOf object in JS SDK? Asking as the intention is to use the existing OnBehalfOf object for the new channelData object for Meeting (MeetingChannelData) :) From this perspective, naming this one as `OnBehalfOf` makes more sense to me as the intention is more explicit. Wdyt?

libraries/botframework-schema/etc/botframework-schema.api.md
```diff
@@ -953,11 +953,55 @@ export interface MeetingInfo {
...
+    surface: 'meetingStage';
+}
+
 // @public
 export interface MeetingStartEventDetails extends MeetingEventDetails {
     startTime: Date;
 }
 
+// @public
+export type MeetingSurface = MeetingStageSurface<any>;
```
YunnyChung, line 1003: This makes me slightly confused. So the list of surfaces should be able to contain MeetingStageSurface<T> and any other surface which is not meeting stage in the future (from the extensibility perspective). If MeetingSurface is set to be meeting stage surface with different content type, what will happen when there is a completely different surface type?

YunnyChung (commented):
First PR review: leaving a few comments :)

libraries/botframework-schema/etc/botframework-schema.api.md
singhk97: Let's say in the future, we have a different meeting surface altogether called `MeetingTabSurface`. We can simply extend `MeetingSurface = MeetingStageSurface<any> | MeetingTabSurface`. And so the list of surfaces can look like the following `surfaces: [meetingStageSurface, meetingTabSurface]` where the `meetingStageSurface` is an object that conforms to the `MeetingStageSurface<any>` type, and `meetingTabSurface` conforms to type `MeetingTabSurface`. And that `surfaces` will conform to `MeetingSurface[]` type. The pipe `|` symbol stands for OR.

libraries/botframework-schema/src/teams/index.ts
singhk97: The reason I decided to put this in targeted meeting notification is because currently it's the only concrete type that requires `channelData` property. Now there's two possibilities going forward.
1. For any future concrete types, all of them have `channelData` property either as required or optional.
2. At least one of the future concrete types would not ever require `channelData` property in the payload.

Now if we decide to move `channelData` to the base class and option 2 becomes a reality, then we run into a bit of an issue. With what I have currently, we are safe guarding against this issue. Also keep in mind that if we define any future concrete type we can simply just add `channelData` property to the object without relying on inheritance. This is because the type of the notification payload isn't `MeetingNotificationBase` but instead `MeetingNotification`.

libraries/botframework-schema/src/teams/index.ts
singhk97: Agreed

libraries/botbuilder/etc/botbuilder.api.md
singhk97: Agreed

libraries/botframework-connector/src/teams/models/mappers.ts
singhk97: Update `MeetingOnBehalfOf` to just `OnBehalfOf`.

libraries/botbuilder/etc/botbuilder.api.md
singhk97: Updated

libraries/botframework-connector/src/teams/models/index.ts
```diff
@@ -128,7 +128,7 @@ export type TeamsMeetingInfoResponse = TeamsMeetingInfo & {
 /**
  * Contains response data for the sendMeetingNotification operation.
  */
-export type TeamsSendMeetingNotificationResponse = TeamsMeetingNotificationRecipientFailureInfos & {
+export type TeamsMeetingNotificationResponse = MeetingNotificationResponse & {
```
yingduyingdu, line 131: Is there any reason why we didn't remove `Teams `for `TeamsMeetingNotificationResponse`

libraries/botframework-connector/src/teams/operations/teams.ts
```diff
@@ -250,18 +250,18 @@ export class Teams {
      */
     sendMeetingNotification(
         meetingId: string,
-        notification: TeamsMeetingNotification,
+        notification: MeetingNotification,
         options?: msRest.RequestOptionsBase,
-    ): Promise<Models.TeamsSendMeetingNotificationResponse>
+    ): Promise<Models.TeamsMeetingNotificationResponse>
```
yingduyingdu, line 255: Is there any reason why we didn't remove `Teams `for `TeamsMeetingNotificationResponse`

libraries/botframework-connector/src/teams/models/index.ts
singhk97: So the reason being is that in this file, all the Teams related models are prefixed with `Teams`. Plus, this is simply an implementation detail and not being exposed to the end user, so there is no inconsistency to worry about.

libraries/botframework-connector/src/teams/operations/teams.ts
singhk97: I described the reason in the comment above

libraries/botframework-schema/etc/botframework-schema.api.md
```diff
@@ -953,11 +953,47 @@ export interface MeetingInfo {
...
+
+// @public
+export interface MeetingNotificationResponse {
+    recipientsFailureInfo?: MeetingNotificationRecipientFailureInfo[];
+}
+
+// @public (undocumented)
+export interface MeetingStageSurface<T> {
+    content: T;
+    contentType: 'task' | string;
```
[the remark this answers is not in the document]
singhk97: The way it's defined is already like an enum, but not really. If it was just `contentType: 'task'`, then any object of this type would be required to have the value as "task", otherwise the compiler will complain. Currently `contentType` can be any string and the "task" is just a guide. I could change it to `contentType: "task" | "unkown"` and it will behave in the exact way you are describing. The reason I have it this way is because Yunny wanted to allow for flexibility. Is this what we have for the dotnet sdk? If so then I'll make the change right away.

libraries/botframework-schema/etc/botframework-schema.api.md
```diff
@@ -953,11 +953,47 @@ export interface MeetingInfo {
...
+    recipientMri: string;
+}
+
+// @public
+export interface MeetingNotificationResponse {
+    recipientsFailureInfo?: MeetingNotificationRecipientFailureInfo[];
+}
+
+// @public (undocumented)
+export interface MeetingStageSurface<T> {
```
yingduyingdu, line 983: Does it make sense to add _extends Surface_ for` interface MeetingStageSurface<T>`
_export interface MeetingStageSurface<T> extends Surface {
     ...
}_

libraries/botframework-schema/etc/botframework-schema.api.md
```diff
@@ -953,11 +953,47 @@ export interface MeetingInfo {
...
+    surface: 'meetingStage';
+}
+
 // @public
 export interface MeetingStartEventDetails extends MeetingEventDetails {
     startTime: Date;
 }
 
+// @public
+export type MeetingSurface = MeetingStageSurface<any>;
```
yingduyingdu, line 995: If add `extends Surface` for` interface MeetingStageSurface<T>` (full example in the conversation above), Is there a way to show the inheritance relationship between `MeetingStageSurface `and `Surface`?

Also, does it make sense to rename `MeetingSurface `in  `export type MeetingSurface = MeetingStageSurface<any>;` to `Surface`. or "MeetingStageSurface"

libraries/botframework-schema/etc/botframework-schema.api.md
```diff
@@ -1614,6 +1658,18 @@ export interface TabSuggestedActions {
...
 }
 
+// @public
+export interface TargetedMeetingNotification extends MeetingNotificationBase<TargetedMeetingNotificationInfo> {
+    channelData?: MeetingNotificationChannelData;
+    type: 'targetedMeetingNotification';
+}
+
+// @public
+export interface TargetedMeetingNotificationInfo {
```
yingduyingdu, line 1668: Can we rename the interface to `TargetedMeetingNotificationValue`.
It will align the name in existing repos.

singhk97: Will do in the in the next iteration

libraries/botframework-schema/etc/botframework-schema.api.md
singhk97: `MeetingStageSurface<T> extends Surface`, simply copies the properties/type pair in `Surface` to `MeetingStageSurface<T>`, nothing more. So then there's not much reason to have a base interface since it will have only one child interface. It seems like the dotnet sdk names it `Surface`. I will make the change from `MeetingSurface` to `Surface` in the next iteration.

libraries/botframework-schema/etc/botframework-schema.api.md
singhk97: MeetingStageSurface<T> extends Surface, simply copies the properties/type pair in Surface to MeetingStageSurface<T>, nothing more. So then there's not much reason to have a base interface since it will have only one child interface.

libraries/botframework-schema/etc/botframework-schema.api.md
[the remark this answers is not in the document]
yingduyingdu: Hi @singhk97 , in .net, `contentType `is a enum type with `Unknown `and `Task`

libraries/botframework-schema/etc/botframework-schema.api.md
yingduyingdu: But aren't we going to support more surfaces in the future? Such as TabSurface or ContentBubbleSurface?

libraries/botframework-schema/etc/botframework-schema.api.md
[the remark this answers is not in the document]
singhk97: Ok makes sense, will update the type in the next iteration

libraries/botframework-schema/etc/botframework-schema.api.md
singhk97: Let's say we have a `Surface` interface with `surface` property with type string. In the sub-interface (think subclass) `MeetingStageSurface<T>` we will be overriding it to something like `surface: "meetingStageSurface"`. Now if we assign a variable the type `Surface`, it will not accept an object that has type `MeetingStageSurface<T>` even though it is a sub-interface of `Surface`. That is, polymorphism doesn't work. Now if the sub-interface did not extend the parent interface, and everything else remains the same, then behavior will be the exact same. This means that having `MeetingStageSurface<T>` extend  `Surface` being completely redundant.

Now in the future if we have `TabSurface`. We simply update `type MeetingSurface = MeetingStageSurface<T> | TabSurface` and any object of type `MeetingSurface` can be either of the two assigned types. This is pretty much the polymorphism equivalent in typescript.

libraries/botframework-schema/src/teams/index.ts
YunnyChung: Happy to hop on a call;

The below is the intention that I want the code to capture. 

(1) Meeting Notification payload will have the following three properties. 
- Type (required)
- T Value  (required)
- ChannelData (optional).

(2) TargetedMeetingNotification is a part of MeetingNotification family whose type is 'TargetedMeetingNotificaiton'. 
- The unique part of TargetedMeetingNotification is that the Value always should contain `Recipients` and `Surfaces`. 

(3) Let's say we introduce another member of Meeting Notification payload whose type is 'Test'. 
- `Test` meeting notification payload will have a different structure for its `Value` (ex: it will have int and string)
- But it still has an ability to send user attributes as it is a part of Meeting Notification payload (therefore, it should have channelData as an optional parameter)

If your current implementation does allow this scenario, I am all good :)  

Coming from my OOP oriented programming language perspective, I cannot quite see how the current implementation of model classes satisfies this intention unless each child class from MeetingNotificationBase specifies ChannelData.

YunnyChung (commented):
Follow-up

libraries/botframework-schema/src/teams/index.ts
singhk97: I think what I have is very close to what you are describing. (1), (2) are already satisfied. The new interface `Test` will be part of the `MeetingNotification` family with `Value` being whatever type we want it to be as feature designers. This interface will also have `channelData` property explicitly defined instead of it being inherited from the base interface. This will then satisfy (3).

I think what I have is very close to what you would like. Let's hop on call soon to clarify.

libraries/botframework-schema/etc/botframework-schema.api.md
YunnyChung: @singhk97 , I think @yingduyingdu's point is to check whether the current design aligns with what the model class hierarchy that we have discussed. 

If I am understanding your point correctly, though the implementation details might be different, this is the right way to (a) show the correct intention to developers and (b) support the future extensibility. Am I aligned with you? 

My assumption was for typescript, how it represents the class hierarchy would be very similar to the other existing OOP languages, so the current implementations gave me some surprise :) But my assumption might have been wrong :)

YunnyChung (commented):
A follow up comment :)

libraries/botframework-schema/src/teams/index.ts
YunnyChung: Hey @singhk97 , happy to hop on a call - will set up something :) 
I know the current design CAN satisfy the behaviors. But is it the right way of doing so when we know that any child classes of MeetingNotificationBase need to have ChannelData as an optional property? Does the current design reflect the intention of the API correctly?

libraries/botframework-schema/etc/botframework-schema.api.md
YunnyChung: Hey @singhk97 , I want to also have a conversation on this as well. 
First, we decided to make SDK changes on top of the existing one to show the correct intentions of the API to developers. One of the intentions is to let developers know that they can send any list of surfaces in one API.  

I don't think the current implementation shows that intention of the API. Also,I am not sure how the current implementation will support this sceanrio: when a developer sends both `MeetingStageSurface<T>` and `TabSurface` in one API request (and both surface will have completely different schemas), how would that work when there is no common class/interface between them?

libraries/botframework-schema/src/teams/index.ts
singhk97: Will move the `channelData` to base class in the next iteration based on offline discussion.

libraries/botframework-schema/etc/botframework-schema.api.md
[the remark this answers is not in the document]
singhk97: Hi @yingduyingdu, In an offline discussion with Yunny I found out that passing `unknown` in the payload would return a 4xx response. And we shouldn't promote something that doesn't work (i.e by making it one of the possible types). She also mentioned that the reason it's there in the dotnet repo is because of a C# enum idiosyncrasy. For this reason will simply keep the type as `contentType: 'task'`.

YunnyChung (approved):
Looks good to me! Thank you :)

yingduyingdu (approved):

johnataylor (approved):

PR #4429, opened 2023-02-09
fix: Update channelServiceRoutes to add next() parameter to support restify 10.0.0+ 

anishprasad01:
#minor
Fixes **Internal**: https://portal.microsofticm.com/imp/v3/incidents/details/365652257/home

## Description
restify 10.0.0 and greater enforces the need for a `next()` callback that passes execution to the next function in the handler chain. `channelServiceRoutes.ts` does not currently address this requirement. Since this class is required to interact with skills, this means that root bots that import this class will fail to start.

## Specific Changes

  - Add the `next: Function` parameter to the handlers in `channelServiceRoutes.ts`
  - Add a `return next()` statement to the aforementioned handlers
  - Update the tests in `channelServiceRoutes.test.js` to pass in an empty `Function` as `next` in each test

## Testing

Ran the updated unit tests successfully.

### DIFF
--- a/libraries/botbuilder/src/channelServiceRoutes.ts
+++ b/libraries/botbuilder/src/channelServiceRoutes.ts
@@ -92,6 +92,7 @@ export class ChannelServiceRoutes {
                             res.send(resourceResponse);
                         }
                         res.end();
+                        return next();
                     })
                     .catch((err) => {
                         ChannelServiceRoutes.handleError(err, res);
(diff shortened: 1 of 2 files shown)


johnataylor:
we need to first verify that this change doesn't break existing Node and Restify dependencies

anishprasad01:
> we need to first verify that this change doesn't break existing Node and Restify dependencies

Tested the code as of most recent commit in Sample 81 with `Node 16.19.0 + restify 8.6.1` and `Node 18.14.0 + restify 10.0.0`. Both ran successfully. Unit tests also passing.

johnataylor (approved):

PR #4442, opened 2023-03-15
fix: [#4440][Bot node.js] Compile error for accessing "conversation" and "organizer" fields for get meeting details bot API

### LINKED ISSUE #4440: [Bot node.js] Compile error for accessing "conversation" and "organizer" fields for get meeting details bot API
Reported by Teams via: https://domoreexp.visualstudio.com/MSTeams/_workitems/edit/2923118

image

Apparently this works in the DotNet version.

Added in:  https://github.com/microsoft/botbuilder-js/pull/3738

ceciliaavila:
Fixes #4440

## Description
This PR fixes the compile error in TS when trying to access the properties returned by the _getMeetingInfo_ function.
It replaces the **_TeamsMeetingInfo_** return type with the correct **_MeetingInfo_** type, also matching with the .NET implementation.

_**Note**: Although this would be a breaking change, as the method's signature was updated, the data returned by the method remained the same. Anyone using the old interface wouldn't be able to access its properties._

## Specific Changes
- Replaced **_TeamsMeetingInfo_** with **_MeetingInfo_** as return type in _getMeetingInfo_ function.
- Updated **_TeamsMeetingInfoResponse_** type to use the correct _MeetingInfo_ interface.
- Updated method's signature in **_botbuilder.api.md_**.

## Testing
These images show the compile error solved and the meeting info properly returned by the method.
image


tracyboehrer (approved):

tracyboehrer:
@ceciliaavila This does change the signatures so the test-compat fails

ceciliaavila:
> @ceciliaavila This does change the signatures so the test-compat fails

Yes, I'm fixing the check now. 

Note: Although this would be a breaking change, as the method's signature was updated, the data returned by the method remained the same. Anyone using the old interface wouldn't be able to access its properties.

PR #4461, opened 2023-04-22
fix: [#4452][#4456][#4460][botframework-streaming] Should reject pending requests on disconnection

### LINKED ISSUE #4452: [bf-streaming] Web Socket disconnected should rejects all ongoing requests
## Versions
4.19.3

18

No

What os are you using
Ubuntu

## Describe the bug

`RequestManager` keep tracks of all outgoing requests. More-or-less the call manager of RPC.

When a Web Socket disconnected, `botframework-streaming` does not signal `RequestManager` to reject all pending requests. Thus, all pending requests are kept forever and never returned.

One failing case, after Web Socket disconnected, C2 can still send an activity. For that send request, despite it is disconnected and should fail, were never rejected and just keep on forever.

## To Reproduce

1. Instantiate `DirectLineStreaming` via `botframework-directlinejs` package
2. Call `postActivity()`
3. Before the message is put on the Web Socket connection, kill the connection

## Expected behavior

`postActivity()` should fail.

However, it is not failing because `NodeWebSocketClient` and `BrowserWebSocketClient` did not tell `RequestManager` about the disconnection.

While we can say `NodeWebSocketClient`/`BrowserWebSocketClient` should signal disconnection, however, `RequestManager` has no public members to reject all pending requests.

Even worse, the requests at `RequestManager._pendingRequests` are *not rejectable*. There are no way to call `pendingRequest.reject()` function.

That means, no requests can be rejected regardless of whatever situation. This means, either a request resolved, or keep on forever.

### LINKED ISSUE #4456: [botframework-streaming] Test cases with `.to.not.throw` are inconclusive
## Versions
`main`

18

Unrelated

What os are you using
Unrelated

## Describe the bug

Many of the tests write like this:

image

However, when calling `expect(fn).to.not.throw`. This does *not* call `fn()` to evaluate the result. It simply *do nothing*.

It should be `expect(fn).to.not.throw()`.

## To Reproduce

In Mocha, try this test.

```js
it.only('should not throw', () => {
    const fn = () => {
        throw new Error('actually throw');
    };

    expect(fn).to.not.throw;
});
```

## Expected behavior

This test should fail.

However, because `.to.not.throw` is not called, the test is not calling `fn()` and it is not thrown.

After changing to `.to.not.throw()`, the test will fail and this is correct behavior.

## Additional context

There are 62 tests that are inconclusive because of this issue.

### LINKED ISSUE #4460: [botframework-streaming] Race condition: requests missed when server respond quickly
## Versions
`main`

`19`

Unrelated

What os are you using
Unrelated

## Describe the bug

When sending a request, if the processor respond quickly, the request may not be resolved. This is because they are queued up a bit too late.

## To Reproduce

1. Create a `WebSocketServer` and `NodeWebSocketClient` and connect them together
2. On client side, call `await nodeWebSocketClient.send()`

## Expected behavior

The `send()` call should resolve.

Instead, the `send()` call is never resolved/rejected.

compulim:
Fixes #4452. Fixes #4456. Fixes #4460.

# APPROVERS: This fix may need to port to C#. Please check C# source code and port accordingly.

## Description

When Web Socket disconnected, all pending requests should be rejected.

## Specific Changes

- Added `RequestManager.rejectAllResponses()` to reject all pending responses
- Updated `browserWebSocketClient` and `nodeWebSocketClient` to call `RequestManager.rejectAllResponses()` on disconnection
- Fixed inconclusive tests related to `.to.throw`
- Fixed a racing condition in `ProtocolAdapter`
   - If server respond quickly, some responses may be dropped
   - This is because their respective requests are registered in `RequestManager` only after the request is send, the response may already get through

## Testing

Added tests to make sure `NodeWebSocketClient` and `WebSocketServer` can communicate with each other by using real Web Socket. This should make our test more stable.

Note: because there is no infrastructure for testing `BrowserWebSocketClient`, there are no tests added for it.

## Additional context

This fix is related to an issue in `botframework-directlinejs`. When `postActivity()` is called after disconnection, it is not failing.

`botframework-directlinejs` did not fail the call because `botframework-streaming` does not reject any pending requests.

### DIFF
--- a/libraries/botframework-streaming/src/webSocket/webSocketTransport.ts
+++ b/libraries/botframework-streaming/src/webSocket/webSocketTransport.ts
@@ -59,7 +59,7 @@ export class WebSocketTransport implements ITransportSender, ITransportReceiver 
      * @returns `true` if the the transport is connected and ready to send data, `false` otherwise.
      */
     get isConnected(): boolean {
-        return this.ws?.isConnected;
+        return !!this.ws?.isConnected;
     }
 
     /**
--- a/libraries/botframework-streaming/tests/NodeWebSocket.test.js
+++ b/libraries/botframework-streaming/tests/NodeWebSocket.test.js
@@ -10,7 +10,7 @@ describe('NodeWebSocket', function () {
     it('creates a new NodeWebSocket', function () {
         const socket = new NodeWebSocket(new FauxSock());
         expect(socket).to.be.instanceOf(NodeWebSocket);
-        expect(socket.close()).to.not.throw;
+        expect(() => socket.close()).to.not.throw();
     });
 
     it('starts out connected', function () {
@@ -21,14 +21,17 @@ describe('NodeWebSocket', function () {
     it('writes to the socket', function () {
         const socket = new NodeWebSocket(new FauxSock());
         const buff = Buffer.from('hello');
-        expect(socket.write(buff)).to.not.throw;
+        expect(() => socket.write(buff)).to.not.throw();
     });
 
     it('attempts to open a connection', function () {
         const socket = new NodeWebSocket(new FauxSock());
         expect(
             socket.connect().catch((error) => {
-                expect(error.message).to.equal('connect ECONNREFUSED 127.0.0.1:8082');
+                expect(
+                    error.message === 'connect ECONNREFUSED 127.0.0.1:8082' ||
+                        error.message === 'connect ECONNREFUSED ::1:8082'
+                ).to.be.true;
             })
         );
     });
@@ -37,7 +40,7 @@ describe('NodeWebSocket', function () {
         const sock = new FauxSock();
         const socket = new NodeWebSocket(sock);
         expect(sock.messageHandler).to.be.undefined;
-        expect(socket.setOnMessageHandler(() => {})).to.not.throw;
+        expect(() => socket.setOnMessageHandler(() => {})).to.not.throw();
         expect(sock.messageHandler).to.not.be.undefined;
     });
 
@@ -45,7 +48,7 @@ describe('NodeWebSocket', function () {
         const sock = new FauxSock();
         const socket = new NodeWebSocket(sock);
         expect(sock.errorHandler).to.be.undefined;
-        expect(socket.setOnErrorHandler(() => {})).to.not.throw;
+        expect(() => socket.setOnErrorHandler(() => {})).to.not.throw();
         expect(sock.errorHandler).to.not.be.undefined;
     });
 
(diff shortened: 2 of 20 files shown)


libraries/botframework-streaming/tests/WebSocket.test.js
```diff
@@ -139,91 +139,15 @@ describe('Streaming Extensions WebSocket Library Tests', function () {
             const transport = new WebSocketTransport(sock);
             expect(transport.isConnected).to.be.true;
             const buff = Buffer.from('hello', 'utf8');
-            expect(transport.onReceive(buff)).to.not.throw;
-        });
-    });
-
-    describe('WebSocket Client Tests', function () {
```
compulim, line 146: These tests are moved to new `WebSocketClientServer.test.js`.

libraries/botframework-streaming/src/webSocket/webSocketTransport.ts
```diff
@@ -59,7 +59,7 @@ export class WebSocketTransport implements ITransportSender, ITransportReceiver
      * @returns `true` if the the transport is connected and ready to send data, `false` otherwise.
      */
     get isConnected(): boolean {
-        return this.ws?.isConnected;
+        return !!this.ws?.isConnected;
```
compulim, line 62: If `ws` is `null/undefined`, `isConnected` should return `false`.

libraries/botframework-streaming/tests/NodeWebSocket.test.js
```diff
@@ -21,14 +21,17 @@ describe('NodeWebSocket', function () {
...
+        expect(() => socket.write(buff)).to.not.throw();
     });
 
     it('attempts to open a connection', function () {
         const socket = new NodeWebSocket(new FauxSock());
         expect(
             socket.connect().catch((error) => {
-                expect(error.message).to.equal('connect ECONNREFUSED 127.0.0.1:8082');
+                expect(
+                    error.message === 'connect ECONNREFUSED 127.0.0.1:8082' ||
```
compulim, line 32: The test should work under IPv6.

libraries/botframework-streaming/tests/PayloadSender.test.js
```diff
@@ -135,13 +135,11 @@ describe('PayloadTransport', function () {
...
         });
 
-        it('gracefully fails when trying to write before connecting.', function (done) {
+        it('gracefully fails when trying to write before connecting.', function () {
+            // When not connected, PayloadSender should not throw.
+            // It should drop the packet silently.
             const ps = new PayloadSender();
-            ps.disconnected = () => done();
             expect(ps.isConnected).to.equal(false);
-            ps.connect(new FauxSock());
```
compulim, line 142: Removed bunches of line.

The test is for "**write before connecting**". It should not call `ps.connect()`.

compulim:
All required checks are completed. Some *optional* checks are failing because:

- tests.yml / tests.yml ubuntu (node 16.x) (pull_request)
   - Named pipe doesn't work under Ubuntu
- E2E-BF-Streaming-DL-ASE-Test 
   - It seems the pipeline is down since 2023 March 23

tracyboehrer (approved):

PR #4494, opened 2023-06-21
fix: [botframework-streaming] Some named pipe tests are not passing after fixing their inconclusiveness

### LINKED ISSUE #4462: [botframework-streaming] Some named pipe tests are not passing after fixing their inconclusiveness
## Versions
`main`

`18`

Unrelated

What os are you using
Windows

## Describe the bug

Some tests are not conclusive because some of their expectations are not running. After fixing them, some issues are discovered with the library. Tests that are failing and some investigations:

- Streaming Extensions NamedPipe Library Tests > NamedPipe Transport Tests > throws when reading from a dead socket
   - *This is an issue on production code.*
   - Reading from a dead socket will never resolve/reject. 
   - The only way to reject reading is: read from a *live* socket, close/error out the socket
- Streaming Extensions NamedPipe Library Tests > NamedPipe Client Tests > sends without throwing
   - *This is an issue on test code.*
   - Without connecting the named pipe, the `send()` call will *never* resolve
   - The test code should connect to the server and respond, then the `send()` call will resolve and completing the scenario
- Streaming Extensions NamedPipe Library Tests > NamedPipe Server Tests > sends without throwing
   - *This is an issue on test code.*
   - Without connecting the named pipe, the `send()` call will *never* resolve
   - The test code should connect to the client and respond, then the `send()` call will resolve and completing the scenario

## To Reproduce

> Currently, those offending tests are disabled by `it.skip()`.

1. Modify offending tests by changing it from `it.skip()` to `it()`
2. Run `npm test` under `/libraries/botframework-streaming`

## Expected behavior

All tests should pass.

## Screenshots

image

image

image

erquirogasw:
Fixes #4462

## Description
This PR fixes the NamedPipe unit tests that weren't passing.
It adds code to the _receive_ method of the _NamedPipeTransport_ class to handle null and destroyed sockets.
Additionally, it updates the transport unit test to be compatible with the new changes and the client-server tests to make the connection correctly.

## Specific Changes
  - New condition in the _receive_ method of the _NamedPipeTransport_ class to handle null and destroyed sockets.
  - Updated '_throws then reading from a dead socket_' unit test to be compatible with the changes of the _receive_ method.
  - Updated client and server '_sends without throwing_' unit tests to make the connection correctly.

## Testing
image

### DIFF
--- a/libraries/botframework-streaming/src/namedPipe/namedPipeTransport.ts
+++ b/libraries/botframework-streaming/src/namedPipe/namedPipeTransport.ts
@@ -85,7 +85,11 @@ export class NamedPipeTransport implements ITransportSender, ITransportReceiver 
      * @returns The buffer containing the data from the transport.
      */
     receive(count: number): Promise<INodeBuffer> {
-        if (this._activeReceiveResolve) {
+        if (!this.socket) {
+            throw new Error('Socket unavailable to make the connection.');
+        } else if (this.socket.destroyed) {
+            throw new Error('Socket was destroyed.');
+        } else if (this._activeReceiveResolve) {
             throw new Error('Cannot call receive more than once before it has returned.');
         }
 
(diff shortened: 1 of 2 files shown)


tracyboehrer:
These tests are failing due to missing Azure resources.  Solution pending.

Issue #4454, opened 2023-04-21
Use HashSet instead of string array for endorsement


jineshjin:
### Use this [query](https://github.com/Microsoft/botbuilder-js/issues?q=is%3Aissue+is%3Aopen++label%3Afeature-request+) to search for the most popular feature requests.

**Is your feature request related to a problem? Please describe.**
Yes as part of complexity

**Describe the solution you'd like**
we would like to have O(1) complexity instead of O(N)

**Describe alternatives you've considered**
Have to rely on O(N)

**Additional context**
As mentioned here: https://github.com/microsoft/botbuilder-js/blob/main/libraries/botframework-connector/src/auth/endorsementsValidator.ts#L50 when are we planning to use HashSet for endorsement.

PR #4526, opened 2023-08-21
fix: [botframework-connector] Use HashSet instead of string array for endorsement

crdev13:
Fixes [#4454](https://github.com/microsoft/botbuilder-js/issues/4454)
#minor

## Description
This small PR just changes an argument type in the static _validate_ method of the _EndorsementsValidator_ class.

## Specific Changes

  - Before this change the _validate_ method was receiving an array for endorsements, now it now receives a HashSet and reduces the time comlexity when it tries to verify if a key(channelId) exists. 

## Testing
All unit tests passed
endorsements_validator

### DIFF
--- a/libraries/botframework-connector/src/auth/endorsementsValidator.ts
+++ b/libraries/botframework-connector/src/auth/endorsementsValidator.ts
@@ -25,14 +25,14 @@ export class EndorsementsValidator {
      * some specific channels. That list is the endorsement list, and is validated here against the channelId.
      * @returns {boolean} True is the channelId is found in the Endorsement set. False if the channelId is not found.
      */
-    static validate(channelId: string, endorsements: Set<string>): boolean {
+    static validate(channelId: string, endorsements: string[]): boolean {
         // If the Activity came in and doesn't have a Channel ID then it's making no
         // assertions as to who endorses it. This means it should pass.
         if (channelId === null || channelId.trim() === '') {
             return true;
         }
 
-        if (!endorsements) {
+        if (endorsements === null) {
             throw new AuthenticationError('endorsements required', StatusCodes.UNAUTHORIZED);
         }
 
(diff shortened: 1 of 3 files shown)


crdev13:
@microsoft-github-policy-service agree

libraries/botframework-connector/src/auth/endorsementsValidator.ts
```diff
@@ -25,14 +25,14 @@ export class EndorsementsValidator {
      * some specific channels. That list is the endorsement list, and is validated here against the channelId.
      * @returns {boolean} True is the channelId is found in the Endorsement set. False if the channelId is not found.
      */
-    static validate(channelId: string, endorsements: string[]): boolean {
+    static validate(channelId: string, endorsements: Set<string>): boolean {
```
ceciliaavila, line 28: Changing the function's signature would break the backward compat. Instead, could you create a new function that receives the Set argument and call it inside the old validate function?

ceciliaavila (changes requested):

tracyboehrer (approved):

libraries/botframework-connector/src/auth/endorsementsValidator.ts
tracyboehrer: I'm not opposed to breaking this if having another method is clutter.  Is there harm in having two?

ceciliaavila: I suggest a private function to be called inside this old one (passing the converted strings[] as set<string>) so those who are using this with an array won't experience a break.

ceciliaavila (approved):

crdev13:
@ceciliaavila When you get a chance can you please run `pr-style.yml / pr-style (pull_request)` again?

PR #4560, opened 2023-11-08
fix: Disable failing expired token test in CloudAdapter

sw-joelmut:
#minor

## Description
This PR disables the `throws exception on expired token` test due to invalid SigningKey value in the token. This will be skipped for now until a fix is provided, so it doesn't block new PR merges.


tracyboehrer (approved):

sw-joelmut:
Hi @tracyboehrer 

There is a unit test failing in main, preventing other PRs to pass. We created this PR to disable it temporarily until we provide a fix.

PR #4585, opened 2023-12-14
fix: add `/.default` to oAuthScope default value for appCredentials

### LINKED ISSUE #4582: UserAssignedIdentity(WorkloadIdentity) auth fails with 'scope https://api.botframework.com is not valid'
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `botframework-connector@4.21.3` for the project I'm working on.

I am using the botbuilder with the msteams connector. My configuration uses a UserAssignedMSI and the botframeworkauthentication is configured like this:

```
const botFrameworkAuthentication = new ConfigurationBotFrameworkAuthentication(
    {},
    new ConfigurationServiceClientCredentialFactory({
      MicrosoftAppId: config.botId,
      MicrosoftAppType: "UserAssignedMSI", // Bot test framework can only handle "MultiTenant",
      MicrosoftAppTenantId: config.botTenantId,
    })
);
```
I use it in conjunction with Azure Workload Identity and my bot is running inside a Pod deployed in AKS.

I get an error when a response should be returned to the bot service: (while fetching the token)
```
azure:identity:warning IdentityClient: authentication error. HTTP status: 400, AADSTS70011: The provided request must include a 'scope' input parameter. The provided value for the input parameter 'scope' is not valid. The scope https://api.botframework.com is not valid. Trace ID: f136be63-1f12-4d80-94be-804dabe01600 Correlation ID: bcc381f3-2e02-455d-a46c-fd3705f24eb4 Timestamp: 2023-12-11 13:59:55Z
```

The correct scope would be `https://api.botframework.com/.default`

I spend some debugging and found a diff comparing the UserAssignedMSI vs the SingleTenant code branch.
Single Tenant version:
https://github.com/microsoft/botbuilder-js/blob/f3db3e98bb139c7aecc921483ea188574de7aada/libraries/botbuilder-core/src/configurationServiceClientCredentialFactory.ts#L97-L130

If you drill further down it is clear that the `audience` is taken as input and then used as `scope` in the oauth flows.
This only works, if `/.default` is appended to the scope. 
For the single tenant version this is done in the `msalAppCredentials`class: https://github.com/microsoft/botbuilder-js/blob/f3db3e98bb139c7aecc921483ea188574de7aada/libraries/botframework-connector/src/auth/msalAppCredentials.ts#L108-L112

In the UserAssignedMSI version the scope is taken without any further modification: https://github.com/microsoft/botbuilder-js/blob/f3db3e98bb139c7aecc921483ea188574de7aada/libraries/botframework-connector/src/auth/managedIdentityAppCredentials.ts#L37

In order to use the same logic like in the SingleTenant version here is the diff that solved my problem:

```diff
diff --git a/node_modules/botframework-connector/src/auth/managedIdentityAppCredentials.ts b/node_modules/botframework-connector/src/auth/managedIdentityAppCredentials.ts
index cc19c1b..fb4f414 100644
--- a/node_modules/botframework-connector/src/auth/managedIdentityAppCredentials.ts
+++ b/node_modules/botframework-connector/src/auth/managedIdentityAppCredentials.ts
@@ -34,7 +34,13 @@ export class ManagedIdentityAppCredentials extends AppCredentials {
 
         this.tokenProviderFactory = tokenProviderFactory;
         super.appId = appId;
-        this.authenticator = new ManagedIdentityAuthenticator(this.appId, this.oAuthScope, this.tokenProviderFactory);
+
+        const scopePostfix = '/.default';
+        let scope = this.oAuthScope;
+        if (!scope.endsWith(scopePostfix)) {
+            scope = `${scope}${scopePostfix}`;
+        }
+        this.authenticator = new ManagedIdentityAuthenticator(this.appId, scope, this.tokenProviderFactory);
     }
 
     /**
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

What do you think? If you agree I can prepare an pull request for the change

patst:
Fixes #4582

## Description
The default constant has the issuer url. The oAuthScope requires to end with `/.default`.

## Specific Changes

  - added  `/.default` to the default oAuthScope value
  
## Testing
we are using this change in our application (patched via `patch-package` library)

### DIFF
--- a/libraries/botframework-connector/src/auth/appCredentials.ts
+++ b/libraries/botframework-connector/src/auth/appCredentials.ts
@@ -48,7 +48,7 @@ export abstract class AppCredentials implements ServiceClientCredentials {
     constructor(
         appId: string,
         channelAuthTenant?: string,
-        oAuthScope: string = AuthenticationConstants.ToBotFromChannelTokenIssuer
+        oAuthScope: string = AuthenticationConstants.ToBotFromChannelTokenIssuer + "/.default"
     ) {
         this.appId = appId;
         this.tenant = channelAuthTenant;
--- a/libraries/botframework-connector/src/auth/appCredentials.ts
+++ b/libraries/botframework-connector/src/auth/appCredentials.ts
@@ -48,7 +48,7 @@ export abstract class AppCredentials implements ServiceClientCredentials {
     constructor(
         appId: string,
         channelAuthTenant?: string,
-        oAuthScope: string = AuthenticationConstants.ToBotFromChannelTokenIssuer + "/.default"
+        oAuthScope: string = AuthenticationConstants.ToBotFromChannelTokenIssuer + '/.default'
     ) {
         this.appId = appId;
         this.tenant = channelAuthTenant;


patst:
> @ceciliaavila Can your team review this in light of how this is handled in the other repos? There is some divergence here and very well could need consideration SDK-wide. Python does something similar. DotNet does not appear to. Though the MsalAppCredentials does during refreshToken.

thanks for taking a look at this. 

Looking forward to get it resolved this way or another, since the current behaviour is just broken.

I fixed the eslint finding (had trouble getting it to run locally on an arm macbook, `yarn install` does not seems to work for all the dependencies in the repository )

ceciliaavila:
Closing in favor of PR #4607

PR #4608, opened 2024-01-23
fix: Remove old @microsoft/recognizers-text-number version with postinstall scripts

JhontSouth:
#minor

## Description
This PR removes the old version or **_@microsoft/recognizers-text-number_** from the other recognizers packages that require it.

## Specific Changes
  - Added postinstall scripts in botbuilder-dialogs and botbuilder-dialogs-adaptive package.json files to remove the old version of  **_@microsoft/recognizers-text-number_**.

## Testing
These images show the dependencies tree before and after running the postinstall scripts.
image

### DIFF
--- a/libraries/botbuilder-dialogs-adaptive/package.json
+++ b/libraries/botbuilder-dialogs-adaptive/package.json
@@ -47,8 +47,7 @@
     "node-fetch": "^2.6.7"
   },
   "devDependencies": {
-    "@types/node-fetch": "^2.5.3",
-    "rimraf": "^3.0.2"
+    "@types/node-fetch": "^2.5.3"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


tracyboehrer (approved):

andrewconnell:
By using Yarn in the post-install script, you're FORCING a dependency on anyone using your package must have Yarn installed.

There's zero reason for this... why not use NPM?

It's fine if you use Yarn in your dev & builds, but to force it on _ALL_ consumers of the package? 👎👎

For example, if a user doesn't have Yarn installed, this breaks all builds of new Teams projects created with TTK (see my referenced bug above

tracyboehrer:
@andrewconnell Apologies.  We'll take care of it.

PR #4656, opened 2024-04-30
fix: Implement Error Handling with Try-Catch for processActivity in a Promise-returning Context

Amrit10737726:
#minor
https://github.com/microsoft/BotBuilder-Samples/issues/2014

Should we consider:   https://github.com/microsoft/BotBuilder-Samples/issues/3148


sw-joelmut:
Hi @tracyboehrer,
 
Since the issue https://github.com/microsoft/BotBuilder-Samples/issues/2014 is an old one (from 2019, pre CloudAdapter existence), we were able to reproduce it by using BotFrameworkAdapter (v4.6.2).
But we couldn’t reproduce this issue with the latest CloudAdapter implementation, as the bot awaits the "adapter.process" method properly, and the error is captured by the OnTurnError handler.

image
image

PR #4775, opened 2024-10-22
fix: Sort Activity Types and add Delay activity

lord-maestro-encora:
## Description

- Activity Types enum are not sorted. Harder to look what exists or not. Also documentation was not sorted either.

- Delay activity was not set into the Activity Types enum. I had to hardcode the string in my code or set as a constant at project level

## Specific Changes

- Activity Types enum is now sorted. And is clearer to look what exists or not. Documentation for the struct is also sorted trying to keep ~ 80 cols per line.

- Delay activity was added. We will not need to write string or set a constant at project level

### DIFF
--- a/libraries/botframework-schema/src/index.ts
+++ b/libraries/botframework-schema/src/index.ts
@@ -2089,32 +2089,35 @@ export enum ActivityEventNames {
 
 /**
  * Defines values for ActivityTypes.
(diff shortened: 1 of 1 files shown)


lord-maestro-encora:
@microsoft-github-policy-service agree

ceciliaavila (changes requested):
Please check the lint error and the `yarn test:compat` that is failing.

Issue #4684, opened 2024-06-19
Update dependencies to latest version


sw-joelmut:
### Description
Update repo dependencies to latest version taking into consideration vulnerability alerts, breaking changes, node support, and any other limitations that could change how botbuilder is consumed and use.

- [ ] Update vulnerable dependencies
- [ ] Update all dependencies

PR #4784, opened 2024-11-01
bump: [#4684] Upgrade d3-format using import-sync for esm-only package

ceciliaavila:
Addresses #4684
#minor

## Description
This PR updates the `d3-format` package to its latest version 3.1.0. Starting on version 3.x, this package is ESM-only, so we used _import-sync_ to be able to use this version.

## Specific Changes
  - Updated `d3-format` from 2.0.0 to 3.1.0 in _adaptive-expressions_ library.
  - Added `import-sync` as a dependency in _adaptive-expressions_.
  - Added _d3-format_ to the _depcheck_ ignore list as it threw a false positive on unused dependency.
  - Updated the `formatNumbers` and `string` classes to import the package _d3-format_ using the _importSync_ function.

## Testing
The following images show the installed version and the unit tests passing after the changes.
image


ceciliaavila:
We need to replace import-sync as it won't work on Node 22.

ceciliaavila:
Replaced with #4842

PR #4792, opened 2024-11-12
bump: [#4684] Upgrade chai using import-sync for esm-only package

sw-joelmut:
Addresses #4684
#minor

## Description
This PR updates the `chai` package to its latest version 5.1.2. Starting on version 5.x, this package is ESM-only, so we used _import-sync_ to be able to use this version.

## Specific Changes
  - Updated `chai` from 4.5.0 and 5.1.2 in _botbuilder_ and _botframework-streaming_ libraries.
  - Added `import-sync` as a dependency in _botbuilder_ and _botframework-streaming_.
  - Added _chai_ to the _depcheck_ ignore list as it threw a false positive on unused dependency.

## Testing
The following images show the installed version.
image


ceciliaavila:
We need to replace import-sync as it won't work on Node 22.

sw-joelmut:
Replaced by #4844

PR #4796, opened 2024-11-19
feat: Update versions for generators

ceciliaavila:
#minor

## Description
This PR includes the generators' templates to be updated by the **_update-versions_** script by including them in the _workspaces_ list.
It also updates the versions in the _package.json_ file to match the one used in the libraries (4.1.6) and upgrades some dependencies in the TS bots to fix errors found when building the generated bots.

## Specific Changes
- Changed botbuilder dependencies' version to 4.1.6 in all bots' templates.
- Updated TS bots by adding the _types/node_ package and upgrading _typescript_ dependency to version 4.9.3. Also, updated _tsconfig_ files updating the _module_ and adding _moduleResolution_ and _esModuleInterop_ properties.
- Added the _'generators/generator-botbuilder/generators/app/templates/'_ path to workspaces in the _package.json_ file.
- Updated _updateVersions_, _workspace_, and _package_ files to include the new projects in the script.

## Testing
These images show the updated version in the bots and the bots working with the changes.
image
image

### DIFF
--- a/generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.js
+++ b/generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.js
@@ -1,6 +1,6 @@
 {
     "name": "<%= botname %>",
-    "version": "1.0.0",
+    "version": "4.1.6",
     "description": "<%= botDescription %>",
     "author": "Generated using Microsoft Bot Builder Yeoman generator v<%= version %>",
     "license": "MIT",
@@ -17,10 +17,10 @@
     },
     "dependencies": {
         "@microsoft/recognizers-text-data-types-timex-expression": "1.1.4",
-        "botbuilder": "^4.23.1",
-        "botbuilder-ai": "^4.23.1",
-        "botbuilder-dialogs": "^4.23.1",
-        "botbuilder-testing": "^4.23.1",
+        "botbuilder": "4.1.6",
+        "botbuilder-ai": "4.1.6",
+        "botbuilder-dialogs": "4.1.6",
+        "botbuilder-testing": "4.1.6",
         "dotenv": "^8.2.0",
         "restify": "^11.1.0"
     },
--- a/generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.ts
+++ b/generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.ts
@@ -1,6 +1,6 @@
 {
     "name": "<%= botname %>",
-    "version": "1.0.0",
+    "version": "4.1.6",
     "description": "<%= botDescription %>",
     "author": "Generated using Microsoft Bot Builder Yeoman generator v<%= version %>",
     "license": "MIT",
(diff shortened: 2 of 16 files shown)


generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.js
```diff
@@ -17,10 +17,10 @@
     },
     "dependencies": {
         "@microsoft/recognizers-text-data-types-timex-expression": "1.1.4",
-        "botbuilder": "^4.23.1",
-        "botbuilder-ai": "^4.23.1",
-        "botbuilder-dialogs": "^4.23.1",
-        "botbuilder-testing": "^4.23.1",
+        "botbuilder": "4.1.6",
```
rido-min, line 20: any reason to do not use the latest?

generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.ts
```diff
@@ -38,22 +38,23 @@
...
       "@types/mocha": "^7.0.2",
+      "@types/node": "^18.19.47",
       "@types/restify": "8.4.2",
       "mocha": "^7.1.2",
       "nodemon": "^2.0.4",
       "nyc": "^15.0.1",
       "ts-node": "^8.10.1",
       "tslint": "^6.1.2",
-      "typescript": "^4.0.7"
+      "typescript": "^4.9.3"
```
rido-min, line 58: can we use the latest TS?

generators/generator-botbuilder/generators/app/templates/core/tsconfig.json
```diff
@@ -3,13 +3,15 @@
     "composite": true,
     "declaration": true,
     "target": "es2017",
-    "module": "commonjs",
+    "module": "NodeNext",
```
rido-min, line 6: what's the motivation to diverge from commonjs? afaik we have not finalized support to latest node

rido-min (changes requested):
clarify which versions we want to target from the templates
discuss node versions support

generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.js
ceciliaavila: We wanted to keep the same version as the libraries, so it's clearer that the script will replace it.

generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.ts
ceciliaavila: We can, we just kept the same version as the samples.

generators/generator-botbuilder/generators/app/templates/core/tsconfig.json
ceciliaavila: We had to include the `moduleResolution` property to solve a TS issue. It was recommended to match the value for the `module` property with the one for `moduleResolution`, however, it works if we keep the _commonjs_ setting.
image

generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.ts
rido-min: okay, maybe a different PR to update TS?

generators/generator-botbuilder/generators/app/templates/core/tsconfig.json
rido-min: I'd recommend to keep using commonjs

tracyboehrer:
@rido-min  This purpose was to support creating and versioning the generators package during build.  This is to allow the generators to be built and deployed whenever we do a release of the JS SDK.  This does not encompass template specific issues, though we could also make changes there if needed.

rido-min (approved):

generators/generator-botbuilder/generators/app/templates/core/tsconfig.json
ceciliaavila: Sorry, I was confused, the value for `moduleResolution` must match the one for `module`. it does not work with commonjs, but _NodeNext_ supports both CommonJS and ECMAScript modules ([documentation](https://www.typescriptlang.org/docs/handbook/modules/reference.html#the-module-compiler-option)).

PR #4819, opened 2024-12-19
fix: fixes bot state not being saved when it should be

hugop95:
## Description

`botState.saveChanges` documentation states the following

https://github.com/microsoft/botbuilder-js/blob/3b8fcab21a0a5434706da8eace17117722ffd78b/libraries/botbuilder-core/src/botState.ts#L96-L110

> if no object has been cached, an empty object will be created and then saved

However:
https://github.com/microsoft/botbuilder-js/blob/3b8fcab21a0a5434706da8eace17117722ffd78b/libraries/botbuilder-core/src/botState.ts#L111-L114

If `cached` does not exist, we will never go through the `if` statement.

## Specific Changes
  - Fixes the bug 
  - Adds a unit test

## Testing

- Unit test added.

### DIFF
--- a/libraries/botbuilder-core/src/botState.ts
+++ b/libraries/botbuilder-core/src/botState.ts
@@ -110,7 +110,7 @@ export class BotState implements PropertyManager {
      */
     saveChanges(context: TurnContext, force = false): Promise<void> {
         let cached: CachedBotState = context.turnState.get(this.stateKey);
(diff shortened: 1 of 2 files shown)


hugop95:
Nevermind! Just realized that this is only supposed to happen if `force` is true in the end 😅.

PR #4875, opened 2025-04-09
fix: [#4840] The use of the package browserify-sign could violate Microsoft crypto policy

### LINKED ISSUE #4840: The use of the package browserify-sign could violate Microsoft crypto policy
The library `crypto-browserify` which is a dependency in the newest version of `botframework-connector` uses `browserify-sign` which uses another package `elliptic` which, in turn uses `hash.js`. The use of the package `hash.js` violates Microsoft policy for the [use approved crypto libraries](https://liquid.microsoft.com/Web/Object/Read/MS.Security/Requirements/Microsoft.Security.Cryptography.10010). 

We have received a security alert for having this module in our package-lock, and after investigating, it turns out that we get it from `botframework-connector`. We kindly ask that this module is either removed or that an investigation is opened to verify that no code flows into any of the functionality of `hash.js`.

ceciliaavila:
Fixes #4840

## Description
This pull request substitutes the `crypto-browserify` package in both tsup and webpack browser builds with the native Web Crypto API module. With these changes, the `hash.js` dependency causing policy violations is no longer used.

## Specific Changes
- Removed the `crypto-browserify` dependency from both the _root package.json_ and the _browser-echo-bot_.
- Updated `tsup` configuration to use `window.crypto` as the alias for the `crypto` module.
- Updated `webpack` configuration to use `window.crypto` as the alias for the `crypto` module.

## Testing
These images show the **_browser-echo-bot_** working after the changes and the _**hash.js**_ package no longer installed.
image

### DIFF
--- a/libraries/botframework-connector/package.json
+++ b/libraries/botframework-connector/package.json
@@ -59,7 +59,7 @@
     "build:browser:run": "tsup --config ../../tsup/browser.config.ts",
     "build:downlevel-dts": "downlevel-dts lib _ts3.4/lib --checksum",
     "clean": "rimraf _ts3.4 lib tsconfig.tsbuildinfo",
-    "depcheck": "depcheck --config ../../.depcheckrc --ignores azure,sinon,browserify-fs,buffer,crypto-browserify,https-browserify,stream-browserify,stream-http",
+    "depcheck": "depcheck --config ../../.depcheckrc --ignores azure,sinon,browserify-fs,buffer,https-browserify,stream-browserify,stream-http",
     "lint": "eslint .",
     "postbuild": "npm-run-all -p build:browser build:downlevel-dts",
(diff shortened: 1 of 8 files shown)


Majd-Zayyad-MSFT:
Hi @ceciliaavila, when can we expect this PR to be merged? it's been open for more than a month and we need the fix that's included here (:

ceciliaavila:
> Hi @ceciliaavila, when can we expect this PR to be merged? it's been open for more than a month and we need the fix that's included here (:

@tracyboehrer, do you have an estimated release date?

tracyboehrer:
@ceciliaavila One more.  Sorry.  This is my bad for letting it sit so long.

ceciliaavila:
@tracyboehrer, conflicts are fixed. Thanks!

PR #4876, opened 2025-04-11
fix: [#4860] Adaptive expression context access with non-standard objects

### LINKED ISSUE #4860: Adaptive expression context access with non-standard objects
## Versions
What package version of the SDK are you using. 4.23.1
What nodejs version are you using 22
What browser version are you using -
What os are you using macOS

## Describe the bug
When providing a non-standard object to `$root` in `AC.expand()` when rendering adaptive cards, the context cannot access properties that are in non-standard objects. 

## To Reproduce
`AC.expand({ $root: new DataObjectOfYourChoice({ propertyToAccess: "hasAValue"  }) })`

AC template: `"text": "${propertyToAccess}"`

## Expected behavior
Text should show "hasAValue", but the property is returned as undefined

## Additional context
I tried opening a fix PR for this but i'm unsure of the contribution rules in this repo of if it is even possible. Pushing to a feature branch is not allowed for my git user.

The fix would be in `libraries/adaptive-expressions/src/functionUtils.internal.ts` lines 221-232:

```
 const prop: string = Object.keys(instance).find(
      (k: string): boolean => k.toLowerCase() === property.toLowerCase(),
  );
  if (prop !== undefined) {
      value = instance[prop];
  } else if (instance[property] !== undefined || instance[property.toLowerCase()] !== undefined) {
      value = instance[property] ?? instance[property.toLowerCase()];
  }
```

My change is the `else` block.

ceciliaavila:
Fixes #4860

## Description
This PR fixes an issue in the _**accessProperty**_ function using _toLowerCase_ to search for property names within an instance. This code was likely ported from C#, but in JS, property names are case-sensitive, which can lead to unexpected results.

## Specific Changes
- Removed the _toLowerCase_ search in **_accessProperty_** function.
- Fix casing errors in unit tests and added a test case for this scenario.

## Testing
This image shows the unit tests passing after the changes and the new test case added.
image


tracyboehrer:
@ceciliaavila It's fine to abandon this.

